PR #518, opened 2015-01-14
[JBIDE-19035] Fix up ConnectionModel to support saving connection info i...

jcantrill:
...n v3


adietish:
this went outdated/obsolete since we're separating v3 from v2 in https://issues.jboss.org/browse/JBIDE-19048

PR #525, opened 2015-02-02
Fixed v2 tests and added tests for common plugins

yin:
Also updated copyright headers to include year 2015


adietish:
Thanks! merged your tests into openshift.tests, avoiding to add another plugin for now.

adietish:
Thanks! merged your tests into openshift.tests, avoiding to add another plugin for now.

PR #528, opened 2015-02-05
[JBIDE-19176] Copy eclipse loggers into openshift.client so eclipse doesnt throw errors

jcantrill:
@adietish copied in logging appender


adietish:
thanks! merging

PR #529, opened 2015-02-06
JBIDE-19000: Basic Browse Feature

jcantrill:
@adietish Initial implementation of browsing resources.  I added additional jira issues to flush out the other details like property pages, styled text, deferred loading.


plugins/org.jboss.tools.openshift.client/META-INF/MANIFEST.MF
```diff
@@ -23,4 +23,13 @@ Export-Package: com.openshift.client,
...
  com.openshift.internal.client.response,
  com.openshift.internal.client.ssh,
- com.openshift.internal.client.utils
+ com.openshift.internal.client.utils,
+ com.openshift3.client,
+ com.openshift3.client.model,
+ com.openshift3.internal.client
+Import-Package: org.eclipse.debug.internal.core,
+ org.eclipse.osgi.service.debug;version="1.2.0",
+ org.eclipse.pde.internal.core.bundle,
```
adietish: wondering about these 2 dependencies: pde and debug? do we need those or is this an accidental addition?

adietish: after verifying: the debug bundle is needed by the Trace while the pde is not required.

PR #530, opened 2015-02-06
Fix up manifest for [JBIDE-19000]

jcantrill:


adietish:
thanks for the quick response!

PR #536, opened 2015-02-11
[JBIDE-19162] Add feature projects to support new openshift v3 plugins

jcantrill:
@adietish I am unable to get the 'test' features to build as they cant find the actual test projects. Not certain what im missing locally


adietish:
great, thx a lot!
merging to upstream/v3

PR #537, opened 2015-02-12
JBIDE-19162 add openshift v3 feature + test feature to category.xml

maxandersen:
PR #536 did not add features to `category.xml` thus nothing in the actual generated site.

Note: this does not make it available in main jboss tools updatesite yet so that step is still needed.


maxandersen:
just push --forced update since I forgot to include sources in the first one.

PR #539, opened 2015-02-12
V3 rebased on master

jcantrill:
[JBIDE-19252]
@maxandersen @adietish Rebased v3 onto master. Updated plugin dependency issues to fix tests and other issues.  Builds successfully with all passing tests


adietish:
great, thx!

PR #550, opened 2015-02-17
Jbide 19280 display service properties

jcantrill:
@adietish service display properties


adietish:
i rebased and pushed this to upstream/master

PR #551, opened 2015-02-18
[JBIDE-19285] Display replication controller specific properties

jcantrill:
@adietish RC display properties


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/BuildConfigPropertySource.java
```diff
@@ -23,25 +23,23 @@
...
 	public BuildConfigPropertySource(IBuildConfig resource) {
 		super(resource);
-		this.config = resource;
 	}
 
 	@Override
 	public IPropertyDescriptor[] getResourcePropertyDescriptors() {
 		List<IPropertyDescriptor> all = new ArrayList<IPropertyDescriptor>();
 		all.addAll(getBuildTriggerPropertyDescriptors());
-		switch(config.getBuildStrategy().getType()){
```
adietish: any chance config can and/or config.getBuildStrategy are null?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/BuildConfigPropertySource.java
```diff
@@ -53,7 +51,7 @@ public BuildConfigPropertySource(IBuildConfig resource) {
 			break;
 		default:
 		}
-		switch(config.getBuildSource().getType()){
+		switch(getResource().getBuildSource().getType()){
```
adietish: dito: any need for null checks to avoid NPEs?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/BuildConfigPropertySource.java
jcantrill: I think there is always a chance, but that would mean only a few things:

1.  There is an error on the server-side since I believe there is validation.  BuildConfigs always need a build strategy otherwise they arent useful
2.  We don't recognize this type of strategy which would be a version issue and I think gets back to the ticket we logged to re-evaluate the use of enums

----- Original Message -----

PR #557, opened 2015-02-23
[JBIDE-19096] implementing connections wizard

adietish:


plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
```diff
@@ -42,95 +46,114 @@
...
+		if (updateModel) {
+			if (connection != null) {
+				updateModel(connection, connectionFactory, Status.OK_STATUS, Status.OK_STATUS);
+			} else {
+				updateModel(connection, connectionFactory, Status.CANCEL_STATUS, Status.OK_STATUS);
+			}
 		}
-		setConnectionCreated(connectionCreationStatus);
-		updateFrom(connection);
+		addConnectionListener(connection, this.connection);
```
jcantrill: @adietish is there any concern here that if 'updateModel' = false that the connection might be null?  and if we are updating is there a need to add a connection listener?

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
```diff
@@ -176,60 +205,99 @@ public void createConnection() {
...
+			}
+		} catch (IOException e) {
+			updateModel(null, 
+					AutomaticConnectionFactoryMarker.getInstance(),
+					ValidationStatus.error(NLS.bind("Could not connect to host at {0}.", host), e), 
+					Status.OK_STATUS);
+		}
+	}
+
+	private boolean equalsTypeAndHost(IConnection thisConnection, IConnection thatConnection) {
```
jcantrill: @adietish this and that....anyway to distinguish them more or is this just basically an 'equals' method?

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
```diff
@@ -176,60 +205,99 @@ public void createConnection() {
...
-
-	private void setConnectionCreated(IStatus status) {
-		firePropertyChange(PROPERTY_CONNECTION_CREATED, this.connectionCreated, this.connectionCreated = status);
+	private void setConnectionCreationError(IStatus status) {
+		firePropertyChange(PROPERTY_CONNECTION_CREATION_ERROR, this.connectionCreationError, this.connectionCreationError = status);
 	}
 
-	public IStatus getConnectionCreated() {
-		return connectionCreated;
+	public IStatus getConnectionCreationError() {
```
jcantrill: @adietish nameing?  is this only useful for error?  Maybe mention that it specifically is some Status

jcantrill:
@adietish LGTM...merge away.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
adietish: yeah this only stores an error, only meaningful if the connection in this null (could not create it) so that we have a hint for the reason of the failure

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
adietish: +1, this basically is an equals() which should get moved to some common super-class

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
adietish: this is taken care of in ObservablePojoUtils.removePropertyChangeListener

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
adietish: So the basic problem is that (v3) Connection does not extend AbstractConnection and thus wont do the same equals operations as ExpressConnection. I'll have to do deep refactoring for this which I'll do in a new jira.
Merging this one for now

PR #559, opened 2015-02-25
[JBIDE-18999] Add support for v3 openshift connections

jcantrill:
@adietish WIP.  Please review and lets discuss.  This allows us to create v3 connections with token authorization.

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/utils/URIUtils.java
+++ b/plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/utils/URIUtils.java
@@ -1,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Red Hat, Inc. Distributed under license by Red Hat, Inc.
+ * All rights reserved. This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution, and is
+ * available at http://www.eclipse.org/legal/epl-v10.html
+ * 
+ * Contributors: Red Hat, Inc.
+ ******************************************************************************/
+package org.jboss.tools.openshift.common.core.utils;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * Helper methods for manipulating URIs
+ */
+public class URIUtils {
+	
+	private URIUtils(){
+	}
+	
+	public static  Map<String, String> splitFragment(String location){
+		URI uri = null;
+		try {
+			uri = new URI(location);
+		} catch (URISyntaxException e) {
+			throw new RuntimeException(e);
+		}
+		return splitFragment(uri);
+	}
+	
+	public static Map<String, String> splitFragment(URI uri){
+		HashMap<String, String> fragments = new HashMap<String, String>();
+		String fragment = uri.getFragment();
+		if(fragment != null){
+			String [] entries = fragment.split("&");
+			for (String entry : entries) {
+				String[] pair = entry.split("=");
+				fragments.put(pair[0], pair[1]);
+			}
+		}
+		return fragments;
+	}
+}
--- a/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
+++ b/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
@@ -303,8 +303,10 @@ class ConnectionWizardPageModel extends ObservableUIPojo {
 	public IStatus connect() {
 		IStatus status = Status.OK_STATUS;
 		try {
-			if(connection != null) {
-				connection.connect();
+			if(connection != null && !connection.connect()) {
+				String message = String.format("Unable to connect to %s", connection.getHost());
+				OpenShiftCommonUIActivator.log(message, null);
+				status = StatusFactory.errorStatus(OpenShiftCommonUIActivator.PLUGIN_ID, message);
 			}
 //		} catch (NotFoundOpenShiftException e) {
 //			// connectionCreationError user without domain
--- a/plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
+++ b/plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
@@ -5,13 +5,17 @@ Bundle-SymbolicName: org.jboss.tools.openshift.core
 Bundle-Version: 3.0.0.qualifier
 Bundle-Activator: org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator
 Bundle-Vendor: %Bundle-Vendor
-Require-Bundle: org.jboss.tools.openshift.common.core;bundle-version="[3.0.0,4.0.0)",
- org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
+Require-Bundle: org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
  org.jboss.tools.common.core;bundle-version="3.6.0",
- org.eclipse.core.runtime
+ org.eclipse.core.runtime,
+ org.jboss.tools.openshift.common.core;bundle-version="3.0.0",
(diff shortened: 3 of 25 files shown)


plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/utils/URIUtils.java
```diff
@@ -0,0 +1,46 @@
...
+		URI uri = null;
+		try {
+			uri = new URI(location);
+		} catch (URISyntaxException e) {
+			throw new RuntimeException(e);
+		}
+		return splitFragment(uri);
+	}
+	
+	public static Map<String, String> splitFragment(URI uri){
```
adietish: can you please explain this? in code it looks much more as if you would split argument (key/value-) pairs than fragments: http://en.wikipedia.org/wiki/Fragment_identifier

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
```diff
@@ -303,8 +303,10 @@ public IStatus getConnectionCreationError() {
 	public IStatus connect() {
 		IStatus status = Status.OK_STATUS;
 		try {
-			if(connection != null) {
-				connection.connect();
+			if(connection != null && !connection.connect()) {
```
adietish: when connecting many errors can occurr. You can experience network outage, the server can go internal server etc. and last but not least authorization can fail. It it thus imho not sufficient to have a boolean to inidicate whether we could connect or not. We need all details about what went wrong. I thus definitely prefer the prior design where #connect throws, carrying the failure details in the exception.

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
```diff
@@ -5,13 +5,17 @@ Bundle-SymbolicName: org.jboss.tools.openshift.core
 Bundle-Version: 3.0.0.qualifier
 Bundle-Activator: org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator
 Bundle-Vendor: %Bundle-Vendor
-Require-Bundle: org.jboss.tools.openshift.common.core;bundle-version="[3.0.0,4.0.0)",
- org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
+Require-Bundle: org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
  org.jboss.tools.common.core;bundle-version="3.6.0",
- org.eclipse.core.runtime
+ org.eclipse.core.runtime,
+ org.jboss.tools.openshift.common.core;bundle-version="3.0.0",
```
adietish: you removed 'org.jboss.tools.openshift.common.core;bundle-version="[3.0.0,4.0.0)' and replaced it with the opened ended dependency range 'org.jboss.tools.openshift.common.core;bundle-version="3.0.0"'? I guess that's by mistake, right?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
```diff
@@ -11,41 +11,71 @@
...
-//		initializeCapabilities();
-		return true;
+		if(authorize()){
+//			initializeCapabilities();
+			return true;
+		} 
+		return false;
+	}
+
+	private boolean authorize() {
```
adietish: This looks like a method that should throw (ex. what happens if server goes 500 internal error, or we run into a timeout?)

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/auth/AuthorizationClient.java
```diff
@@ -0,0 +1,203 @@
...
+import org.apache.http.impl.client.HttpClients;
+import org.jboss.tools.openshift.core.auth.IAuthorizationClient;
+import org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator;
+
+import com.openshift.client.NoopSSLCertificateCallback;
+import com.openshift.client.IHttpClient.ISSLCertificateCallback;
+import com.openshift.client.OpenShiftException;
+
+@SuppressWarnings("restriction")
+public class AuthorizationClient  implements IAuthorizationClient{
```
adietish: Shouldnt this go into the openshift client library?

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/utils/URIUtils.java
jcantrill: @adietish A sample url that is returned by the OAuth service is shown in the associated test which is a number of key/value pairs that are only accessible (as far as i can tell) by making a call to 'getFragement'.  This is similiar to what the web console does too.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
jcantrill: @adietish This change was made to make sure to get the error and NOT add the connection to the explorer, but I also added some cases which will produce an exception that is caught by a subsequent throw.  The boolean return may not be needed here but probably requires us to discuss further.  We may even wish to remove the boolean return from the method signature or maybe identify it as something that indicates maybe that we are already connected?

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
jcantrill: @adietish +1 Probably can be revereted...I'm sure this was not intentional

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
jcantrill: @adietish See previous comment.  I think we should consider changing the method signature and maybe identifying the exceptions that it might throw.  These might be those already identified by v2

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/auth/AuthorizationClient.java
jcantrill: @adietish probably...My only opposition to that is the client library doesnt have httpclient available to it...but I'm not opposed to adding it.  We probably should also introduce some factory classes to support.  Probably some place also to add this into the auth strategy, but I'm not sure exactly where or how

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/utils/URIUtils.java
adietish: but those are no fragments aren't they? by the code it looks as if you'd parse url parameters?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/auth/AuthorizationClient.java
adietish: If those authorizers are not present in osjc, how would a 3rd pary user of osjc connect and authorize? To me this very much looks like we need this in OSJC

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/utils/URIUtils.java
jcantrill: @adietish they are similiar to URL parameters but there is no '?' to indicate the query values so they aren't returned from the URL when you make that call, but they are returned as part of the fragment

adietish:
merged with minor corrections in https://github.com/jbosstools/jbosstools-openshift/pull/566
@jcantrill can you please create a jira to move authorization to osjc (while verifying that httpclient would be able to pick Eclipse preferences/proxy settings)?

jcantrill:
@adietish Added https://issues.jboss.org/browse/OSJC-143

PR #561, opened 2015-02-26
[JBIDE-19374] Load and save v3 connections

jcantrill:
@adietish added load and save connections

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.ui/META-INF/MANIFEST.MF
+++ b/plugins/org.jboss.tools.openshift.common.ui/META-INF/MANIFEST.MF
@@ -5,7 +5,7 @@ Bundle-SymbolicName: org.jboss.tools.openshift.common.ui;singleton:=true
 Bundle-Version: 3.0.0.qualifier
 Bundle-Localization: plugin
 Bundle-Activator: org.jboss.tools.openshift.internal.common.ui.OpenShiftCommonUIActivator
-Require-Bundle: org.jboss.tools.openshift.common.core;bundle-version="1.0.0",
+Require-Bundle: org.jboss.tools.openshift.common.core;bundle-version="3.0.0",
  org.jboss.tools.openshift.egit.ui;bundle-version="2.6.100",
  org.jboss.tools.common.ui;bundle-version="3.6.0",
  org.jboss.tools.common.core;bundle-version="3.7.0",
--- a/plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
+++ b/plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
@@ -5,13 +5,17 @@ Bundle-SymbolicName: org.jboss.tools.openshift.core
 Bundle-Version: 3.0.0.qualifier
 Bundle-Activator: org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator
 Bundle-Vendor: %Bundle-Vendor
-Require-Bundle: org.jboss.tools.openshift.common.core;bundle-version="[3.0.0,4.0.0)",
- org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
+Require-Bundle: org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
  org.jboss.tools.common.core;bundle-version="3.6.0",
- org.eclipse.core.runtime
+ org.eclipse.core.runtime,
+ org.jboss.tools.openshift.common.core;bundle-version="3.0.0",
+ org.apache.httpcomponents.httpclient;bundle-version="4.3.6",
+ org.apache.httpcomponents.httpcore;bundle-version="4.3.3",
+ org.apache.commons.codec;bundle-version="1.6.0"
 Bundle-RequiredExecutionEnvironment: JavaSE-1.6
 Bundle-ActivationPolicy: lazy
 Export-Package: org.jboss.tools.openshift.core.connection;uses:="com.openshift.client,org.jboss.tools.openshift.common.core.connection,com.openshift3.internal.client",
  org.jboss.tools.openshift.core.preferences,
- org.jboss.tools.openshift.internal.core;x-friends:="org.jboss.tools.openshift.ui"
+ org.jboss.tools.openshift.internal.core;x-friends:="org.jboss.tools.openshift.ui",
+ org.jboss.tools.openshift.internal.core.auth
 Service-Component: META-INF/connectionFactory.xml
--- a/plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
+++ b/plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
@@ -11,7 +11,9 @@ Require-Bundle: org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
  org.jboss.tools.openshift.common.core;bundle-version="3.0.0",
  org.apache.httpcomponents.httpclient;bundle-version="4.3.6",
  org.apache.httpcomponents.httpcore;bundle-version="4.3.3",
- org.apache.commons.codec;bundle-version="1.6.0"
+ org.apache.commons.codec;bundle-version="1.6.0",
+ org.jboss.ide.eclipse.as.dmr;bundle-version="3.1.0",
+ org.apache.commons.lang;bundle-version="2.6.0"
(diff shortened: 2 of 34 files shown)


plugins/org.jboss.tools.openshift.common.ui/META-INF/MANIFEST.MF
```diff
@@ -5,7 +5,7 @@ Bundle-SymbolicName: org.jboss.tools.openshift.common.ui;singleton:=true
 Bundle-Version: 3.0.0.qualifier
 Bundle-Localization: plugin
 Bundle-Activator: org.jboss.tools.openshift.internal.common.ui.OpenShiftCommonUIActivator
-Require-Bundle: org.jboss.tools.openshift.common.core;bundle-version="1.0.0",
+Require-Bundle: org.jboss.tools.openshift.common.core;bundle-version="3.0.0",
```
adietish: ideally this range should be close ended "[3.0.0, 4.0.0)"

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
```diff
@@ -5,13 +5,19 @@ Bundle-SymbolicName: org.jboss.tools.openshift.core
 Bundle-Version: 3.0.0.qualifier
 Bundle-Activator: org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator
 Bundle-Vendor: %Bundle-Vendor
-Require-Bundle: org.jboss.tools.openshift.common.core;bundle-version="[3.0.0,4.0.0)",
- org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
+Require-Bundle: org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
  org.jboss.tools.common.core;bundle-version="3.6.0",
- org.eclipse.core.runtime
+ org.eclipse.core.runtime,
+ org.jboss.tools.openshift.common.core;bundle-version="3.0.0",
```
adietish: you removed the closed dependency range "[3.0.0,4.0.0)" and replaced it with an open ended one.

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
```diff
@@ -5,13 +5,19 @@ Bundle-SymbolicName: org.jboss.tools.openshift.core
...
+ org.apache.httpcomponents.httpclient;bundle-version="4.3.6",
+ org.apache.httpcomponents.httpcore;bundle-version="4.3.3",
+ org.apache.commons.codec;bundle-version="1.6.0",
+ org.jboss.ide.eclipse.as.dmr;bundle-version="3.1.0",
```
adietish: where do depend on dmr? ideally we shouldnt.

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
```diff
@@ -5,13 +5,19 @@ Bundle-SymbolicName: org.jboss.tools.openshift.core
...
+ org.apache.commons.lang;bundle-version="2.6.0"
 Bundle-RequiredExecutionEnvironment: JavaSE-1.6
 Bundle-ActivationPolicy: lazy
 Export-Package: org.jboss.tools.openshift.core.connection;uses:="com.openshift.client,org.jboss.tools.openshift.common.core.connection,com.openshift3.internal.client",
  org.jboss.tools.openshift.core.preferences,
- org.jboss.tools.openshift.internal.core;x-friends:="org.jboss.tools.openshift.ui"
+ org.jboss.tools.openshift.internal.core;x-friends:="org.jboss.tools.openshift.ui",
```
adietish: i bet we miss a x-friend to org.jboss.tools.openshift.test here

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/OpenShiftCoreActivator.java
```diff
@@ -43,12 +48,39 @@ public static BundleContext getBundleContext() {
 	    return myContext;
 	}
 
+	@Override
     public void start(BundleContext context) throws Exception {
         super.start(context);
         myContext = context;
+        String[] connections = OpenShiftPreferences.getInstance().getConnections();
```
adietish: i would definitely move the code that's responsible to add serialized connections to the registry to its own method.

tests/org.jboss.tools.openshift.test/META-INF/MANIFEST.MF
```diff
@@ -18,7 +18,10 @@ Require-Bundle: org.jboss.tools.openshift.client;bundle-version="[3.0.0,4.0.0)",
...
  org.jboss.tools.locus.mockito;bundle-version="1.9.5",
  org.jboss.tools.openshift.common.ui,
- org.apache.commons.lang;bundle-version="2.6.0"
+ org.apache.commons.lang;bundle-version="2.6.0",
+ org.apache.commons.logging;bundle-version="1.1.1",
+ org.jboss.tools.common.core;bundle-version="3.7.0"
 Bundle-RequiredExecutionEnvironment: JavaSE-1.6
 Import-Package: com.openshift.client,
- org.jboss.tools.openshift.internal.common.ui
+ org.jboss.tools.openshift.internal.common.ui,
```
adietish: I'd not mix import packages and require bundles.

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
adietish: got it, the connection serializer needs it. We now have 2 different serializers: ConnectionURL and your Serializer. Ideally we should unify this at some point

plugins/org.jboss.tools.openshift.common.ui/META-INF/MANIFEST.MF
jcantrill: @adietish agree

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
jcantrill: @adietish agree...don't we have the impediment that we need to be concerned with people who have existing v2 connections.  I prefer the json approach I took because its more explicit, but moving v2 to that format would require some sort of migration right? In regards to DMR dependency, I didnt see an issue since we already implicitly depend on it in OSJC

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
jcantrill: @adietish agree. needs to be replaced with a range, but range dependency doesnt seem consistent in the manifest  in general.

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
adietish: @jcantrill yes if we want to be consistent we'd have to migrate existing v2 connections since existing users have connections stored in the other format.
About the dependency: I was not picky in terms of dependency per se but more on the fact that I would consider dmr an implementation detail on how osjc marshals/unmarshals the json. But I now groked that dmr was used to persist connections, so that an all different story.

adietish: @jcantrill I now spotted another incosistency: ConnectionRegistry is using ConnectionURL for all it's connections (v3 and v2). While we currently use ConnectionURL to serialize v2 connections we use json for v3. IMHO we should either use ConnectionURL for all connections when serializing or move ConnectionURL out of the registry.

jcantrill: @adietish my preference would be to drop connectionurl as its point is really not obvious too me and is just kind of confusing.  I feel like anything that this class can provide could be integrated into a connection class

adietish:
I rebased, added the above changes and added a ConnectionPersistency class with tests: https://github.com/jbosstools/jbosstools-openshift/pull/567

plugins/org.jboss.tools.openshift.core/META-INF/MANIFEST.MF
adietish: Dropping is not an Option since we at least need it to load existing connctions. We can hide it in the ExpressConnetion internals and choose a different storage in ConnectionRegistry.

PR #562, opened 2015-02-26
[JBIDE-19381] Add SSL Cert callback for v3 connections

jcantrill:
@adietish adding in ssl cert call back for v3...had to dup some classes because of the way dependencies work and common plugins not depending upon the client plugins.


adietish:
rebased, corrected deps and merged: https://github.com/jbosstools/jbosstools-openshift/pull/569

PR #563, opened 2015-02-27
[JBIDE-19382] Prompt for credentials when unauthorized

jcantrill:
@adietish prompt for creds when unauthorized, but im not convinced we want to prompt using the new connection wizard


adietish:
rebased and merged in PR https://github.com/jbosstools/jbosstools-openshift/pull/570

PR #571, opened 2015-03-10
removed duplicate logging and added exception msg

adietish:


adietish:
@jcantrill can you please review this?
I removed the duplicate logging and added an exception msg

jcantrill:
@adietish LGTM

PR #579, opened 2015-03-19
JBIDE-19132 since no commits in Openshift for JBT 4.2.3, didn't need to bump parent pom. But just in case, might as well use Final here >> jbosstools-4.2.x

nickboldt:


adietish:
Thanks!
rebased and pushed to upstream/jbosstools-4.2.x

PR #581, opened 2015-03-23
[JBDIE-19495] removed erroneously added v3 plugin stubs

adietish:


xcoulon:
+1. PR seems safe and does not break the build.

PR #596, opened 2015-04-10
[JBIDE-19575] Update to support deprecation of com.openshift.client

jcantrill:
@adietish please merge


adietish:
+1
Thanks! 
also removed all access restrictions for the client plugin (internal markers with x-friends) so that 3rd party Eclipse plugins can use the client plugin.

PR #603, opened 2015-04-16
[JBIDE-19629] Fix unable to connect to v3 when installing only v3 from n...

jcantrill:
...ightly build.  Fix username and password validations


adietish:
thanks a lot!
+1

PR #607, opened 2015-04-18
JBIDE-19627 - OpenShift features disable workspace selection

snjeza:
https://issues.jboss.org/browse/JBIDE-19627
OpenShift features disable workspace selection

PR #608, opened 2015-04-20
 JBIDE-19627 openshift disable workspace selection

maxandersen:
Marked osgi services as immediate=false to avoid loading too early.


maxandersen:
never mind - missed #607 was already there.

PR #613, opened 2015-04-28
[JBIDE-19712] Add details to template pane when creating application

jcantrill:
@adietish added details pane for templates which displays description, provider, and other image info of template.  please review


adietish:

PR #614, opened 2015-04-28
[JBIDE-19478] Add template filtering when creating an application

jcantrill:
@adietish please review and merge


adietish:

PR #615, opened 2015-04-29
[JBIDE-19763] Get templates from openshift namespace and group into namespaces

jcantrill:
@adietish added more grouping by  namespace.  review and merge please


adietish:

PR #623, opened 2015-05-18
[JBIDE-19815] Upload template from local disk to create app

jcantrill:
@adietish upload template from disk

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/NewApplicationHandler.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/NewApplicationHandler.java
@@ -15,8 +15,12 @@ import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.ui.handlers.HandlerUtil;
+import org.jboss.tools.openshift.core.connection.Connection;
+import org.jboss.tools.openshift.core.connection.ConnectionNotFoundException;
+import org.jboss.tools.openshift.core.connection.ConnectionsRegistryUtil;
 import org.jboss.tools.openshift.internal.common.ui.utils.UIUtils;
 import org.jboss.tools.openshift.internal.common.ui.utils.WizardUtils;
+import org.jboss.tools.openshift.internal.ui.OpenShiftUIActivator;
 import org.jboss.tools.openshift.internal.ui.wizard.application.NewApplicationWizard;
 import org.jboss.tools.openshift.internal.ui.wizard.application.NewApplicationWizardModel;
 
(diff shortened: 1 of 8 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/NewApplicationHandler.java
```diff
@@ -31,10 +34,15 @@
...
 	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
-		IProject project = UIUtils.getFirstElement(HandlerUtil.getCurrentSelection(event), IProject.class);
-		NewApplicationWizardModel model = new NewApplicationWizardModel(project);
-		NewApplicationWizard wizard = new NewApplicationWizard(model);
-		WizardUtils.openWizard(wizard, HandlerUtil.getActiveShell(event));
+		try {
+			IProject project = UIUtils.getFirstElement(HandlerUtil.getCurrentSelection(event), IProject.class);
+			Connection connection = ConnectionsRegistryUtil.getConnectionFor(project);
+			NewApplicationWizardModel model = new NewApplicationWizardModel(project, connection.getResourceFactory());
```
adietish: what if connection is null?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/NewApplicationHandler.java
```diff
@@ -31,10 +34,15 @@
...
+			NewApplicationWizard wizard = new NewApplicationWizard(model);
+			WizardUtils.openWizard(wizard, HandlerUtil.getActiveShell(event));
+		}catch(NullPointerException e) {
```
adietish: where is an NPE thrown? Shouldnt we add guards for NPE instead of catching?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateListPage.java
```diff
@@ -69,11 +79,101 @@ public TemplateListPage(IWizard wizard, ITemplateListPageModel model) {
...
+			public void widgetSelected(SelectionEvent e) {
+				FileDialog dialog = new FileDialog(getShell(), SWT.OPEN);
+				if(StringUtils.isNotBlank(model.getTemplateFileName())) {
+					File file = new File(model.getTemplateFileName());
+					dialog.setFilterPath(file.getParentFile().getAbsolutePath());
+				}
+				String file = "";
+				while(file != null) {
+					file = dialog.open();
+					if(file != null) {
```
adietish: what could cause file(s) to be null? Why 2 nested file null-checks? why use a while-loop? Can the user pick several files?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/NewApplicationHandler.java
jcantrill: Look below

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/NewApplicationHandler.java
jcantrill: @adietish We should never have a null pointer in this situation because connection should never be null; it could only be null if we stubbed it and didnt retrieve it from a connection.  Maybe we should consider making ConnectionsRegistryUtil.getConnectionFor(project) throw an error or take an argument for it to throw an error which would make it more explicit as to whats going on.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateListPage.java
jcantrill: @adietish maybe this should be a do/while since we want at least 1 pass through.  dialog.open only returns null when a user does not choose a file.  This logic was added to handle the case where a user picks a malformed template file, we notify, and allow them to choose another.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/NewApplicationHandler.java
adietish: hmm, but in general terms throwing indicates an error whereas we simply didnt find a connection for a given project which isnt an error? why not simply "if not null then"

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateListPage.java
adietish: @jcantrill ok gotcha

jcantrill:
@adietish updated ...per our conversation. Added catch for ClassCastException.  I don't think we can check for this in the ResourceFactory because its a generic call (because of type erasure?)

adietish:
+1, thanks!
merged and pushed

PR #625, opened 2015-05-20
[JBIDE-19832] Show all templates when there is no template filter

jcantrill:
@adietish merge please


adietish:
+1, look good code-wise

adietish:
tested it, works for me, thanks!

PR #628, opened 2015-05-29
[JBIDE-19829] Add oauth support to connection wizard

jcantrill:
@adietish @maxandersen Here is the initial preview for the oauth support.  It has some quirks but I believe I've gotten over the major issues I was having.


jcantrill:
@adietish please merge per our conversation with @maxandersen

adietish:
rebased and merged to upstream/master

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
```diff
@@ -0,0 +1,364 @@
...
+			browser.setText(NLS.bind("<html><body><img style=\"display:block; margin-top: 50%; margin-left:auto; margin-right:auto\" alt=\"Embedded Image\" width=\"64\" height=\"64\" src=\"{0}\" /></body></html>",image));
+			setURL(url);
```
adietish: what is this binary data with the html page?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
```diff
@@ -0,0 +1,364 @@
...
+
+		@Override
+		public void awake(IJobChangeEvent event) {
+		}
+
+
+		@Override
+		public void done(final IJobChangeEvent event) {
+			if(event.getJob() instanceof AuthDetailsJob) {
+				Display.getDefault().asyncExec(new Runnable() {
```
adietish: You should get the display that you already have in the browser:
browser.getShell().getDisplay()

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
```diff
@@ -0,0 +1,364 @@
...
+					@Override
+					public void run() {
+						AuthDetailsJob job = (AuthDetailsJob)event.getJob();
+						final IAuthorizationDetails details = job.getDetails();
+						if(details != null) {
+							setURL(details.getRequestTokenLink());
+						}else {
+							IStatus result = job.getResult();
+							showErrorDialog(getShell(),result.getException());
```
adietish: Isnt it sufficient to have display#syncExec around #showErrorDialog or maybe also #setUrl() (is any other UI thread operation involved in the above code?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
```diff
@@ -0,0 +1,364 @@
...
+
+		@Override
+		protected boolean isResizable() {
+			return true;
+		}
+
+
+		@Override
+		protected Point getInitialSize() {
+			return new Point(500, 700);
```
adietish: Do we really need that much height? A user with 1024*768 would have a large amount of his screen covered vertically? (I'll ask max about the majority of our users screensize - we gather user stats with screen sizes. I guess it's 1280*xxx which then would make it ok to have a dialog with 700 px height I guess)

adietish: > [11:00:49 PM] Max Rydahl Andersen: 1280/1400 is the norm, but most presentation systems is max 1024x

So i guess that if there's a way to use less vertically size it would be preferrable but if not possible, nevermind.

PR #634, opened 2015-06-01
[JBIDE-19895] Add explicit browser timeout to Oauth browser

jcantrill:
cc @adietish


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
```diff
@@ -306,6 +323,44 @@ protected Control createDialogArea(Composite parent) {
...
+					synchronized (this) {
+						location = event.location;
+					}
+				}
+			});
+			browser.addProgressListener(new ProgressAdapter() {
+				
+				@Override
+				public void completed(ProgressEvent event) {
+					synchronized (this) {
```
adietish: isnt "this" here the ProgressAdapter which nobody else is accessing?

adietish:
+1

PR #639, opened 2015-06-02
[JBIDE-19898] Resolve 'forbidden' exception on initial connection load for new eclipse session

jcantrill:


jcantrill:
@adietish  fixes the inital load issue. not convinced i like the filtering of the connection changed event...

adietish:
as commented in Jira, I see other effects caused by this: now the connection-wizard pops up for me on a basic-auth connection.

adietish:
rebased, merged and pushed to upstream master (after discussing issue is caused by JBIDE-19905 and the behaviour of this patch is expected)

PR #647, opened 2015-06-04
[JBIDE-19786] Replace browser dialog impl with external browser

jcantrill:
@adietish


adietish:
added further changes on top so that progress bar is showing up in wizard while we retrieve the authentication details.

PR #648, opened 2015-06-04
[JBIDE-19937] Display reason why new app wizard fails for unsupported capability

jcantrill:
@adietish


adietish:
+1, works for me

PR #649, opened 2015-06-04
[JBIDE-19786] fixing oauth token retrieval from stg

adietish:


adietish:

PR #650, opened 2015-06-04
[JBIDE-19942] Null pointer expanding expanding openshift view on inital session

jcantrill:
@adietish  this fixes an issue but is dependent upon a commit I pushed to the rest client


adietish:
rebased, merged and pushed.

PR #651, opened 2015-06-05
[JBIDE-19955] Change the Openshift explorer project lable to show display name or default to name

jcantrill:


jcantrill:
@adietish merged as this is minor...not worth reviewing

PR #652, opened 2015-06-05
[JBIDE-19956] Filter templates during new app by name and tags

jcantrill:
@adietish

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateViewerFilter.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateViewerFilter.java
@@ -19,16 +19,16 @@ import com.openshift.restclient.capability.resources.ITags;
 import com.openshift.restclient.model.IResource;
 
 /**
- * Filter a viewer of resources by tag if the
- * resource is annotated with tags
+ * Filter a list of templates by name or if
+ * they have a tag annotation
  * 
  * @author jeff.cantrill
  */
-public class AnnotationTagViewerFilter extends ViewerFilter {
+public class TemplateViewerFilter extends ViewerFilter {
 	
 	private ITextControl filterText;
 	
-	public AnnotationTagViewerFilter(ITextControl txtFilter) {
+	public TemplateViewerFilter(ITextControl txtFilter) {
 		this.filterText = txtFilter;
 	}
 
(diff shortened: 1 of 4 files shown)


adietish:
code looks good to me. Nevertheless I doubt Max would let this in for 4.3.0.Beta1 since it's nothing critical even though more than handy.

jcantrill:
Added with the expectation it won't get in until beta2

adietish:
ok, good. I set the jira fixVersion accordingly

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateViewerFilter.java
```diff
@@ -38,15 +38,19 @@ public boolean select(Viewer viewer, Object parentElement, Object element) {
 		if(!(element instanceof IResource)) {
 			return false;
 		}
+		final String text = filterText.getText();
+		if(StringUtils.isBlank(text)) {
+			return true;
+		}
 		IResource resource = (IResource) element;
-		if(StringUtils.isBlank(filterText.getText())) {
+		if(resource.getName().contains(text)) {
```
adietish: can't #getName return null?

jcantrill: A resource needs a name for it to exist in the system so it can never be 
null.  Also, our client impl using DMR doesn't allow for it though this 
is not an excuse either.  Technically, we should probably invert the 
check since we already check the filterText for null.

On 06/09/2015 08:20 AM, André Dietisheim wrote:
> In
> plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateViewerFilter.java
> <https://github.com/jbosstools/jbosstools-openshift/pull/652#discussion_r32006550>:
>
>>  		IResource resource = (IResource) element;
>> -		if(StringUtils.isBlank(filterText.getText())) {
>> +		if(resource.getName().contains(text)) {
>
> can't #getName return null?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jbosstools/jbosstools-openshift/pull/652/files#r32006550>.
>

-- 
Jeff Cantrill
Senior Software Engineer, Red Hat Engineering
Red Hat
Office: 703-748-4420 | 866-546-8970 ext. 8162420
jcantril@redhat.com
http://www.redhat.com

adietish: I'd agree on the logical requirement where resources would need to have a resource but I could still imagine an error in unmarshalling where the name doesnt get extracted from json. Then dmr catches in, which is fine for our current situation but imagine we'd change the unmarshalling technology and all of a sudden one would unexpectly get an NPE here. I'd personnally add a simple null-check here jsut to make it more robust even if not strictly required.

jcantrill: @adietish just realized that I'm not sure what the correct fix is here.  If the name returns null I would expect errors elsewhere in the code (i.e. label provider) and we we use the name as something displayed to a user.  Also, this is a filter, what should be the return value if the name is null?  Show the template?  We don't have a name to display?  Don't show the template?  It won't appear anyway.  Suggestions?

adietish: I would think that when there's no name then the filter would not match the given criteria in regard of the name? That's at least how I handled matching in the list of quickstarts/cartridges in v2 app wizard.

adietish:
+1 good to go.

adietish:
tested, works for me

adietish:

PR #654, opened 2015-06-09
[JBIDE-20000] Add validator to restrict adding readonly labels when using templates

jcantrill:
@adietish fixed

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/LabelKeyValidator.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/LabelKeyValidator.java
@@ -8,6 +8,8 @@
  ******************************************************************************/
 package org.jboss.tools.openshift.internal.ui.validator;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.regex.Pattern;
 
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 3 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/LabelKeyValidator.java
```diff
@@ -24,12 +26,14 @@
 	
 	public static final int SUBDOMAIN_MAXLENGTH = 253;
 	private static final Pattern SUBDOMAIN_REGEXP = Pattern.compile("^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$");
+	private Collection<String> readonlykeys;
```
adietish: shouldnt this be a set?

adietish:
@jcantrill could we please avoid those merging commits like in this PR? I noticed this in a few PRs so far and if we can avoid them we'd have a much nicer git history?

jcantrill:
@adietish I would love to avoid merging them but I don't know how. I thought this was a product of using Github and something it does.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/LabelKeyValidator.java
jcantrill: @adietish This probably should be a set but for now there is only one value.  I dont anticipate haveing many others.

adietish:
@jcantrill --rebase wont work? or git fetch upstream/master && git reset --hard FETCH_HEAD && ... ?
Having a "clean" history is pretty valuable imho, much easier to read and track. Worth the effort to have this sorted out I'd think.

adietish:
ok good to go and would be great if you could rebase and remove the merge commit for improved history readability

adietish:

adietish:
@jcantrill rebasing made the merge commit go away

PR #660, opened 2015-06-10
[JBIDE-20015] Fix retrieving webhook for source build strategy

jcantrill:
@adietish fix to retrieve build config hook


adietish:
+1 
can you merge it? btw. I would try to avoid merge commits if possible, they only make history harder to read. Better to maybe rebase and avoid them?

jcantrill:
@adietish agreed..i hate those

PR #664, opened 2015-06-12
[JBIDE-19146] load legacy connections

adietish:


alexeykazakov:
Looks good to me.

plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/connection/ExpressConnectionPersistency.java
```diff
@@ -25,7 +29,10 @@
 
 	@Override
 	protected String[] loadPersisted() {
-		return ExpressCorePreferences.INSTANCE.loadConnections();
+		List<String> connections = new ArrayList<String>();
+		connections.addAll(Arrays.asList(ExpressCorePreferences.INSTANCE.loadConnections()));
+		connections.addAll(Arrays.asList(ExpressCorePreferences.INSTANCE.loadLegacyConnections()));
+		return connections.toArray(new String[connections.size()]);
```
fbricon: that's a convoluted way to concatenate arrays. Rather use commons-lang ArrayUtils.addAll(first, second);

adietish: ok, good catch, thx

fbricon:
+1 to apply once arrays concatenation is fixed

maxandersen:
why is there such thing as legacy connections ? didn't we keep the old format even for the new ones ?

adietish:
This was where in some old version connections were stored. The current impl was not reading thos, something I now added.

maxandersen:
so its fixing and additional issue ?

PR #665, opened 2015-06-12
[JBIDE-20027] Add better error messaging to user for unsupported versions

jcantrill:
@adietish To help when a version is not supported when uploading from local disk

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateListPage.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateListPage.java
@@ -55,8 +55,10 @@ import org.jboss.tools.openshift.internal.ui.dialog.ResourceSummaryDialog;
 
 import com.openshift.restclient.ResourceFactoryException;
 import com.openshift.restclient.ResourceKind;
+import com.openshift.restclient.UnsupportedVersionException;
 import com.openshift.restclient.capability.CapabilityVisitor;
 import com.openshift.restclient.capability.ICapability;
+import com.openshift.restclient.model.IProject;
 import com.openshift.restclient.model.IResource;
 import com.openshift.restclient.model.template.IParameter;
 import com.openshift.restclient.model.template.ITemplate;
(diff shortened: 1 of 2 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/application/TemplateListPage.java
```diff
@@ -151,9 +154,15 @@ public void widgetSelected(SelectionEvent e) {
 						try {
 							model.setTemplateFileName(file);
 							return;
-						}catch(ResourceFactoryException | ClassCastException err) {
+						}catch(ClassCastException err) {
```
maxandersen: why is the exception not passed to the error dialog so it will be shown in Details... button ?

jcantrill: @maxandersen only thing I can think of is to shield the user from a long trace they don't understand which is the reason it gets logged.  In hind site, I guess that's the reason for the details button?  To present the user with a simple message and an opportunity to see more extensive information?

maxandersen: Problem is that this dialog is showing up on 1 to 2 layers away from the error log and the message is just basically saying "whoops - something went wrong and i can't be bothered to give you a hint of what actually went wrong so please now go close all dialogs and lookup the error view which might not even be installed to look for what happened" :)

It is a common good practice in eclipse that:

A) error messages give at least some context
B) have a Details button where underlying errors can be accessed without use of error log view

It thus should be something like:

`ErrorDialog.openError(shell,
                                                MSG_UPLOAD_TEMPLATE,
                                                NLS.bind("Unable to read and/or parse the file \"{0}\" as a template. Erorr: '{1}', err.getMessage(), ValidationStatus.error(e.getMessage(),e));`

I believe that is what is used most places anyway.

adietish: +1, good catch, I replaced the MessageDialog with an ErrorDialog

adietish:
rebased, added a commit for ErrorDialog and merged, pushed to upstream/master

PR #666, opened 2015-06-15
JBIDE-20066: Beta1 branch must reference Beta1 parent

mickaelistria:


maxandersen:
-1 on doing this. rolling back the version is going to cause more problems than solutions.

adietish:
ok

PR #672, opened 2015-06-18
JBIDE-20039 - Application wizard: templates tree should look disabled…

snjeza:
https://issues.jboss.org/browse/JBIDE-20039
Application wizard: templates tree should look disabled when I switch to use templates from local disk (Linux GTK3 only)


adietish:
+1, it solves it, great. Thanks!

adietish:

PR #678, opened 2015-06-27
[JBIDE-19850] Add port forwarding to a pod

jcantrill:
@adietish @maxandersen Port forwarding code for your initial review.  I still need to do some testing but it works in a similiar fashion to the v2 code.


adietish:
In order for this to compile and run we need to first merge in the openshift-restclient-java: https://github.com/openshift/openshift-restclient-java/pull/40

plugins/org.jboss.tools.openshift.client/.project
```diff
@@ -20,6 +20,16 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
```
adietish: I guess that this change made it here by mistake?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/preferences/IOpenShiftCoreConstants.java
```diff
@@ -0,0 +1,20 @@
...
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.openshift.core.preferences;
+
+/**
+ * @author jeff.cantrill
+ */
+public interface IOpenShiftCoreConstants {
+	
+	static final String OPENSHIFT_CLI_LOC = "openshift.cli.location";
```
adietish: Shouldnt we have this implemented in OpenShiftCorePreferences? Isnt this a constant for the v3 preferences?

plugins/org.jboss.tools.openshift.express.ui/src/org/jboss/tools/openshift/express/internal/ui/console/ConsoleUtils.java
```diff
@@ -48,26 +39,12 @@
...
-	/**
 	 * Registers the given listener as a console listener.
 	 * 
 	 * @param consoleListener
 	 */
 	public static void registerConsoleListener(IConsoleListener consoleListener) {
-		ConsolePlugin plugin = ConsolePlugin.getDefault();
-		IConsoleManager consoleManager = plugin.getConsoleManager();
-		consoleManager.addConsoleListener(consoleListener);
+		org.jboss.tools.openshift.internal.common.ui.console.ConsoleUtils.registerConsoleListener(consoleListener);
```
adietish: Is there a name clash for ConsoleUtils for you to use the full package name?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizard.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @author jeff.cantrill
+ */
+public class PortForwardingWizard extends Wizard  {
+
+	private PortForwardingWizardModel model;
+	public PortForwardingWizard(PortForwardingWizardModel model) {
+		setWindowTitle("Application Port Forwarding");
+		this.model = model;
```
adietish: dont we need a progress bar in this wizard? ex. while the remoting call is executed?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
```diff
@@ -0,0 +1,190 @@
...
+		return pod.getName();
+	}
+
+	
+	public boolean getPortForwarding() {
+		return pod.accept(new CapabilityVisitor<IPortForwardable, Boolean>() {
+
+			@Override
+			public Boolean visit(IPortForwardable capability) {
+				return capability.isForwarding();
```
adietish: I bet that this is doing remoting, I thus think that we should have this done in a job and show progress?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
```diff
@@ -0,0 +1,190 @@
...
+				if(cap.isForwarding()) return null;
+				final MessageConsole console = ConsoleUtils.findMessageConsole(getMessageConsoleName());
+				consoleListener.setCapability(cap);
+				ConsoleUtils.registerConsoleListener(consoleListener);
+				MessageConsoleStream stream = console.newMessageStream();
+				stream.println("Starting port-forwarding...");
+				for (IPortForwardable.PortPair port : ports) {
+					stream.println(NLS.bind("{0} {1} -> {2}", new Object [] {port.getName(), port.getLocalPort(), port.getRemotePort()}));
+				}
+				cap.forwardPorts(ports.toArray(new IPortForwardable.PortPair []{}));
```
adietish: same here, I strongly suspect that we should do this in a background job fearing that we might freeze the UI while this is happening

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
```diff
@@ -0,0 +1,190 @@
...
+		public void setCapability(IPortForwardable cap) {
+			this.cap = cap;
+		}
+		@Override
+		public void consolesRemoved(IConsole[] consoles) {
+			final String messageConsoleName = getMessageConsoleName();
+			for (IConsole console : consoles) {
+				if(console.getName().equals(messageConsoleName)) {
+					try {
+						cap.stop();
```
adietish: what happens if #stop throws an exception?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
```diff
@@ -0,0 +1,190 @@
...
+	private class ConsoleListener implements IConsoleListener{
+		private IPortForwardable cap;
+		
+		public void setCapability(IPortForwardable cap) {
+			this.cap = cap;
+		}
+		@Override
+		public void consolesRemoved(IConsole[] consoles) {
+			final String messageConsoleName = getMessageConsoleName();
+			for (IConsole console : consoles) {
```
adietish: This loop looks like an ideal candidate for a utils method in ConsoleUtils? ex. ConsoleUtils#containsNamedConsole(consoleName, IConsole consoles[])

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
```diff
@@ -0,0 +1,190 @@
...
+	private String getMessageConsoleName() {
+		return NLS.bind("Port forwarding to pod {0} ({1})", pod.getName(), pod.getNamespace());
+	}
+
+	public void stopPortForwarding() {
+		pod.accept(new CapabilityVisitor<IPortForwardable, Object>() {
+
+			@Override
+			public Object visit(IPortForwardable cap) {
+				if(!cap.isForwarding()) return null;
```
adietish: Shouldnt this be executed in a job?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
```diff
@@ -0,0 +1,190 @@
...
+				try {
+					final MessageConsole console = ConsoleUtils.findMessageConsole(getMessageConsoleName());
+					MessageConsoleStream stream = console.newMessageStream();
+					stream.println("Stopping port-forwarding...");
+					cap.stop();
```
adietish: same here?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
```diff
@@ -0,0 +1,190 @@
...
+
+	public Boolean getUseFreePorts() {
+		return useFreePorts;
+	}
+
+	public void setUseFreePorts(Boolean useFreePorts) {
+		// do not change the current bindings if port forwarding is already started.
+		if (!getPortForwarding()) { 
+			updateLocalPortBindings(useFreePorts);
+		}
```
adietish: shouldnt this be run in a job?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
```diff
@@ -0,0 +1,190 @@
...
+		final List<String> bindings = new ArrayList<String>();
+		// update local bindings while avoiding duplicates
+		for (IPortForwardable.PortPair port : ports) {
+			if (useFreePorts) {
+				// find free port for every port
+				port.setLocalPort(SocketUtil.findFreePort());
+			} else {
+				// find duplicates and if match we find free port for those
+				// until stops.
+				port.setLocalPort(port.getRemotePort());
```
adietish: @maxandersen wasnt there a special approach required for MacOS so that port forwarding works on MacOS? What was this? I vaguely remember that one could not bind to 127.0.0.1 but had to use the interface IP? Not sure if this was specific to ssh either?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardPage.java
```diff
@@ -0,0 +1,273 @@
...
+
+	}
+
+
+	private SelectionListener onStartPortForwarding() {
+		return new SelectionAdapter() {
+			@Override
+			public void widgetSelected(SelectionEvent e) {
+				try {
+					WizardUtils.runInWizard(new Job("Starting port forwarding...") {
```
adietish: ah, here's the job. My comments further up might be bogus then. Right?
So if you tell the wizard to require a progress bar (wizard#setNeedsProgressMonitor(true) we should get a progress bar while this job is running.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/OpenShiftPreferencePage.java
```diff
@@ -0,0 +1,149 @@
...
+public class OpenShiftPreferencePage extends FieldEditorPreferencePage implements IWorkbenchPreferencePage {
+
+	private static final int SUCCESS = 0;
+
+	private FileFieldEditor cliLocationEditor;
+	
+	private static final String OC_BINARY_NAME;
+	private static final String [] EXTENSIONS;
+	static {
+		if(SystemUtils.IS_OS_WINDOWS) {
```
adietish: idea to have it ~cleaner and be able to drop statics: external factory class ex. PortForwardingBinary#getInstance() - singleton so that i can cache results. portforwardingBinary#getName() postForwardingBinary#getExtension()

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/OpenShiftPreferencePage.java
```diff
@@ -0,0 +1,149 @@
...
+	
+	
+	
+	@Override
+	protected void performDefaults() {
+		super.performDefaults();
+
+	}
+
+	private String findOCLocation() {
```
adietish: this could then be put into the PortForwardingBinary#getLocation()

plugins/org.jboss.tools.openshift.express.ui/src/org/jboss/tools/openshift/express/internal/ui/console/ConsoleUtils.java
jcantrill: I used the full package to distinguish the fact that the common methods were moved out of express.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizard.java
jcantrill: 'Start forward' give you a busy indicator.  I have tested against staging and its fairly quick so I'm not sure a progress bar is needed.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
jcantrill: Not sure what you mean by 'remoting'.  It is making a call to the server which is the point of the functionality.  I can create a job if you think it provides value but I would test against staging to see how that looks to you.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
jcantrill: Seems reasonable this should be a modal operation.  We dont really have any other way to indicate to the user something is happening.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
jcantrill: Do we really care? It might be worth logging ...

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
jcantrill: We are wrapping the binary and only require the pod name and dont specify the IP.  As long as the binary works, this will work for us too.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardPage.java
jcantrill: Is the busy icon good enough.  I feed progress bar an odd choice when there is no specific progress per se.

jcantrill:
@adietish I have tested on Windows and Linux and am satisfied with the code.  Please review, comment.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizard.java
adietish: I pretend that you can never be sure of how quick a remoting call is. Furthermore having a progress bar would be consistent with the rest. But well, it's not a must, just a nice-to-have, but still.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
adietish: I would run this in a job. Remoting is unreliable in terms of "it works" or even speed I think. We should thus make sure that no UI-freeze can happen and run every remoting call from a job.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
adietish: yeah I would at least log it.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardModel.java
adietish: ok

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/portforwading/PortForwardingWizardPage.java
adietish: busy indicator is imho not as good as progress bar, busy indicator cant tell you what's happening. And then progress bar is what we always do in all wizards. Also in terms of code if you already do WizardUtils.runInWizard the progress bar is a free gift when you initialize the wizard to use progress bar while busy indicator only gets shown if you explicitly turn it to be busy and then non-busy again. I dont see any benefit in using busy indicator compared to progress.

jcantrill:
Merged to master and pushed.

PR #680, opened 2015-06-29
JBIDE-20092 - Connection wizard: hitting retrivial link with mouse does not do anything on OSX

snjeza:
https://issues.jboss.org/browse/JBIDE-20092
 Connection wizard: hitting retrivial link with mouse does not do anything on OSX

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
@@ -212,6 +212,11 @@ public class OAuthDetailView extends BaseDetailsView implements IConnectionEdito
 		}
 
 		@Override
+		public void widgetDefaultSelected(SelectionEvent e) {
+			widgetDefaultSelected(e);
+		}
+
+		@Override
 		public void widgetSelected(SelectionEvent e) {
 			if(!StringUtils.isBlank(link)) {
 				return;


adietish:
sorry, has no effect for me.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
```diff
@@ -212,6 +212,11 @@ public String toString() {
 		}
 
 		@Override
+		public void widgetDefaultSelected(SelectionEvent e) {
+			widgetDefaultSelected(e);
+		}
```
nickboldt: Did you mean to call widgetDefaultSelected(e) from inside widgetDefaultSelected(e)? 

Or was this supposed to call widgetSelected instead?

adietish: you're right in noticing that the above patch is actually an endless loop and obviously cant be anything of a fix. I changed the above to a few variants like super.widgetDefaut..(), widgetSelected() etc.
But all this has 0 effect. #widgetDefaultSelected never gets called

nickboldt: Actually, Rob noticed it but I tend to be more verbose about actually putting comments in JIRAs and PRs. :D

adietish: imho all the above explanations are irrelevant since the patch has 0 effects.

adietish: the imho more relevant other tests I did (low-level SWT listeners etc.) are reported in my jira: https://issues.jboss.org/browse/JBIDE-20092?focusedCommentId=13084745&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13084745 (if you're interested)

PR #684, opened 2015-06-30
[JBIDE-20092] replaced link with StyledText and mouse listener

adietish:
@jcantrill  please review, especially the parts highlighted in the code comments


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
```diff
@@ -125,12 +117,15 @@ public Composite createControls(Composite parent, Object context, DataBindingCon
...
+		StyledTextUtils.setTransparent(tokenRequestLink);
+		StyledTextUtils.setLinkText(MSG_TOKEN, tokenRequestLink);
 		GridDataFactory.fillDefaults()
 			.align(SWT.LEFT, SWT.CENTER).span(3, 1).applyTo(tokenRequestLink);
-		tokenRequestLink.setText(MSG_TOKEN);
-		tokenRequestLink.addSelectionListener(new AuthLinkHandler(parent.getShell(), authDetails));
-		
+		if(authDetails != null) {
+			authDetails.getRequestTokenLink();
+		}
```
adietish: this was taken from the contructor in the SelectionAdapter. Is this correct? What is this retrieval all about? It seems to be "null" anyhow?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
```diff
@@ -200,59 +195,49 @@ public String toString() {
 		return IAuthorizationContext.AUTHSCHEME_OAUTH;
 	}
 	
-	private class AuthLinkHandler extends SelectionAdapter{
-		private String link = "";
-		private Shell shell;
-
-		AuthLinkHandler(Shell shell, IAuthorizationDetails details){
-			this.shell = shell;
-			if(details != null) {
```
adietish: this was moved to #createControls, i dont understand what this is all about, it appears to be "null" in all my tests

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
```diff
@@ -200,59 +195,49 @@ public String toString() {
...
-				details.getRequestTokenLink();
-			}
-		}
+	private Listener onRetrieveLinkClicked(final Shell shell) {
+		return new Listener() {
 
-		@Override
-		public void widgetSelected(SelectionEvent e) {
-			if(!StringUtils.isBlank(link)) {
```
adietish: I removed this, I dont see how the link would become non-null anyhow?

jcantrill: Trying to recall, but I believe this is the chicken/egg problem (or was).  Ideally we want to present the link to the user with the actual URL, but we don't know what it is until we try to authenticate and error.  The error returns the link to where to go and get the token.  I'm not sure why we don't do anything with the return value of 'details.getRequestTokenLink();' in the previous block other then to trigger the auth cycle to actually GET the link we need; it otherwise is not useful and could be an artifact that never was removed.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
jcantrill: See my comment below

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/OAuthDetailView.java
jcantrill: See my comments below

jcantrill:
@adietish  LGTM

PR #685, opened 2015-06-30
[JBIDE-20166] Add nullpointer check to return gracefully

jcantrill:


jcantrill:
@adietish merged as was simple change

PR #686, opened 2015-07-02
[JBIDE-20016] Manage projects

jcantrill:
@adietish please review when you have an opportunity


plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
```diff
@@ -179,13 +179,37 @@ protected IStatus run(IProgressMonitor monitor) {
 		job.schedule();
 		return stub;
 	}
-
-	protected void refreshViewer(final Object object) {
+	
+	private void asyncViewerRefresh(final Runnable runner) {
```
adietish: it would make it easier to understand if this method was called something like #asyncUpdateViewer

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -56,16 +57,37 @@ protected void handleConnectionChanged(IConnection connection, String property,
...
 		}
 	}
 	
+	@SuppressWarnings("unchecked")
+	private void handleProjectChanges(Object oldValue, Object newValue) {
+		List<IProject> newProjects = (List<IProject>) newValue;
+		List<IProject> oldProjects = (List<IProject>) oldValue;
+		//theory this should only be one.
+		Object [] diffs = CollectionUtils.disjunction(newProjects, oldProjects).toArray();
+		if(newProjects.size() > oldProjects.size()) {
```
adietish: is the case handled where new projects were added but a project was removed too (which is actually a case where children have to be removed AND children added)?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -56,16 +57,37 @@ protected void handleConnectionChanged(IConnection connection, String property,
...
+			Connection connection = ConnectionsRegistryUtil.getConnectionFor(newProjects.get(0));
+			addChildrenToViewer(connection, diffs);
+		}else if(newProjects.size() < oldProjects.size()){
+			Connection connection = ConnectionsRegistryUtil.getConnectionFor(oldProjects.get(0));
+			removeChildrentFromViewer(connection, diffs);
+			//is this safe across diffs?
+			for (Object project : diffs) {
+				groupMap.remove(project);
+			}
+		}
```
adietish: What about the case where the collection of children stays the same but 1 child was replaced?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -56,16 +57,37 @@ protected void handleConnectionChanged(IConnection connection, String property,
...
+	private void handleProjectChanges(Object oldValue, Object newValue) {
+		List<IProject> newProjects = (List<IProject>) newValue;
+		List<IProject> oldProjects = (List<IProject>) oldValue;
+		//theory this should only be one.
+		Object [] diffs = CollectionUtils.disjunction(newProjects, oldProjects).toArray();
+		if(newProjects.size() > oldProjects.size()) {
+			Connection connection = ConnectionsRegistryUtil.getConnectionFor(newProjects.get(0));
+			addChildrenToViewer(connection, diffs);
+		}else if(newProjects.size() < oldProjects.size()){
+			Connection connection = ConnectionsRegistryUtil.getConnectionFor(oldProjects.get(0));
```
adietish: Same here I think: even though collections differ in size items could have been replaced requiring us to add and remove children

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -100,7 +122,7 @@ private void refreshGrouping(List<ResourceGrouping> groupings, String kind) {
 			ConnectionsRegistry registry = (ConnectionsRegistry) parentElement;
 			return registry.getAll(Connection.class).toArray();
 		} else if (parentElement instanceof Connection) {
-			return ((Connection) parentElement).get(ResourceKind.PROJECT).toArray();
+			return ((Connection) parentElement).getResources(ResourceKind.PROJECT).toArray();
```
adietish: this is a remoting call, are we sure that this wont be execute in the UI thread causing it to freeze?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -114,7 +136,7 @@ private void refreshGrouping(List<ResourceGrouping> groupings, String kind) {
 		try{
 			if (parentElement instanceof Connection) {
 				Connection connection = (Connection) parentElement;
-				return connection.get(ResourceKind.PROJECT).toArray();
+				return connection.getResources(ResourceKind.PROJECT).toArray();
```
adietish: remoting call?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPage.java
```diff
@@ -0,0 +1,291 @@
...
+					new AbstractDelegatingMonitorJob("Loading OpenShift projects...") {
+
+						@Override
+						protected IStatus doRun(IProgressMonitor monitor) {
+							pageModel.loadProjects();
+							return Status.OK_STATUS;
+						}
+					}, new DelegatingProgressMonitor(), getContainer(), dbc);
+		} catch (InvocationTargetException e) {
+			LOG.logError(NLS.bind("Could not load projects for connection {0}", pageModel.getConnection().toString()), e);
```
adietish: this looks like a case for the multi-catch close in jdk7 causing this block to get simpler:
} catch (InvocationTargetException | InterruptedException e)  {
      LOG.logError(NLS.bind("Could not load projects for connection {0}", pageModel.getConnection().toString()), e);
}

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPage.java
```diff
@@ -0,0 +1,291 @@
...
+						}
+						return status;
+					}
+					
+				};
+				try {
+					org.jboss.tools.common.ui.WizardUtils.runInWizard(
+							job, job.getDelegatingProgressMonitor(), getContainer(), dbc);
+				} catch (InvocationTargetException ex) {
+					OpenShiftUIActivator.getDefault().getLogger().logError(NLS.bind("Could not delete OpenShift project {0}", project.getName()), ex);
```
adietish: multi-catch?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPage.java
```diff
@@ -0,0 +1,291 @@
...
+					new AbstractDelegatingMonitorJob("Refreshing Projects...") {
+
+						@Override
+						protected IStatus doRun(IProgressMonitor monitor) {
+							pageModel.refresh();
+							return Status.OK_STATUS;
+						}
+					}
+					, new DelegatingProgressMonitor(), getContainer(), dbc);
+		} catch (InvocationTargetException e) {
```
adietish: multi-catch (since exact same error msg)?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPage.java
```diff
@@ -0,0 +1,291 @@
...
+				.cellToolTipProvider(cellToolTipProvider)
+				.name("Display Name").align(SWT.LEFT).weight(2).minWidth(100).buildColumn()
+				.buildViewer();
+
+		return viewer;
+	}
+
+	private void refreshModel(final DataBindingContext dbc) {
+		try {
+			org.jboss.tools.common.ui.WizardUtils.runInWizard(
```
adietish: looks like the full-package name isnt required here any more since there's no WizardUtils in openshift any more?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPageModel.java
```diff
@@ -0,0 +1,117 @@
...
+			}
+		};
+	}
+
+	public void loadProjects() {
+		if (connection == null) {
+			setProjects(Collections.<IProject>emptyList());
+		} else {
+			List<IProject> projects = connection.getResources(ResourceKind.PROJECT);
+			setProjects(projects);
```
adietish: remoting call: are we sure that this is called from a background job?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/LabelValueValidator.java
```diff
@@ -26,24 +27,34 @@
...
+	}
+
 	@Override
 	public IStatus validate(Object paramObject) {
 		if(!(paramObject instanceof String))
-			return FAILED;
+			return ValidationStatus.error("Value is not an instance of a string");
 		String value= (String) paramObject;
 		if(StringUtils.isEmpty(value))
 			return FAILED;
```
adietish: so far in the tooling we treat the case where we have a missing required parameter as IStatus.CANCEL causing the "*" to appear instead of the error marker:
ValidationStatus.cancel(FAILED)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPageModel.java
```diff
@@ -0,0 +1,117 @@
...
+
+	public IProject getSelectedProject() {
+		return selectedProject;
+	}
+
+	public Connection getConnection() {
+		return connection;
+	}
+	
+	public void dispose() {
```
adietish: @Override?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/NewProjectWizard.java
```diff
@@ -0,0 +1,67 @@
...
+					return new Status(Status.ERROR, 
+							OpenShiftUIActivator.PLUGIN_ID, 
+							NLS.bind("Could not create project \"{0}\": {1}", getModel().getProjectName(), e.getMessage()), e);
+				}
+			}
+		};
+		
+		try {
+			WizardUtils.runInWizard(job, getContainer());
+		} catch (Exception e) {
```
adietish: I think that catch (InvocationTargetException | InterruptedException e) would be better since we swallow everything here (which then might be keep maintainers wondering where the heck the exception was swallowed)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/NewProjectWizardModel.java
```diff
@@ -0,0 +1,90 @@
...
+		firePropertyChange(PROPERTY_DESCRIPTION, this.description, this.description = description);
+	}
+
+	public String getDisplayName() {
+		return displayName;
+	}
+
+	public void setDisplayName(String displayName) {
+		firePropertyChange(PROPERTY_DISPLAY_NAME, this.displayName, this.displayName = displayName);
+		this.displayName = displayName;
```
adietish: not required, since the fireProperyChange already does this (..., this.displayName = displayName)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/NewProjectWizardPage.java
```diff
@@ -0,0 +1,127 @@
...
+		ValueBindingBuilder
+			.bind(WidgetProperties.text(SWT.Modify).observe(txtDispalayName))
+			.validatingAfterConvert(new IValidator() {
+
+				@Override
+				public IStatus validate(Object value) {
+					String param = (String) value;
+					if(StringUtils.isNotEmpty(param)) {
+						if(param.contains("\t") || param.contains("\n")) {
+							return ValidationStatus.error("Display name may not contain tabs or new lines");
```
adietish: this should be ValidationStatus.cancel(...)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/LabelValueValidator.java
adietish: see https://issues.jboss.org/browse/JBIDE-19975

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/NewProjectWizardPage.java
```diff
@@ -0,0 +1,127 @@
...
+		GridDataFactory.fillDefaults()
+			.align(SWT.FILL, SWT.CENTER).grab(true, false).applyTo(txtDispalayName);
+		ValueBindingBuilder
+			.bind(WidgetProperties.text(SWT.Modify).observe(txtDispalayName))
+			.validatingAfterConvert(new IValidator() {
+
+				@Override
+				public IStatus validate(Object value) {
+					String param = (String) value;
+					if(StringUtils.isNotEmpty(param)) {
```
adietish: is displayName not required? we're returning ok() if it's empty, is this correct?

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
jcantrill: I purposely chose to avoid the word 'update' as it has a connotation with modifying the children of the tree (e.g. update, add, delete, refresh).  I don't really like the name as is but couldnt think of anything better

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
jcantrill: refactored to use Diffs

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
jcantrill: Updated to use Diffs but one question I dont see is answered is the 'update' case.  There is no way of knowing when an update occurs here unless it shows up as an add/remove.  There is a corner case where the viewer will not be updated for changes to an existing project.  These cases may be small so its probably not a  huge issue.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
jcantrill: I think the Diffs changes will resolve this concern

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPage.java
jcantrill: Updated

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPage.java
jcantrill: Updated

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPage.java
jcantrill: Updated

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPage.java
jcantrill: Fixed

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPageModel.java
jcantrill: How do we insure that? Both calls are made from the Page which are wrapped in an AbstractDelegatingMonitorJob.  Should we move the use of the Job to the model?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPageModel.java
jcantrill: Nothing currently to override

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/NewProjectWizard.java
jcantrill: Fixed

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/NewProjectWizardModel.java
jcantrill: Fixed

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/NewProjectWizardPage.java
jcantrill: DisplayName is not required since we can fallback to the actual name.  Display name is currently kept as an annotation on project so it must not be identified as super important.  Also, I think projects are just k8e namespaces with a little sugar so I'm sure this is part of the reason too.

jcantrill:
@adietish updated in response to your comments

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
adietish: As far as I know in databinding if an element of the list is changed while the list by itself wont (no elements removed/added) then no listChange event happens. Afaik a change in an element is - in databinding - catched by the element listener, not the list change listener. So depending on your change mgmt strategy Diff wont inform you sufficiently if you catch element changes in the list change event?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPageModel.java
adietish: ok so sounds fully sufficient, it was just not clear to me when only reading code (and not live testing).

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPageModel.java
adietish: ok, wasnt sure about #dispose

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/NewProjectWizardPage.java
adietish: ok

adietish:
I guess we still need to handle the element change case discussed above?

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
adietish: Yes I understand but I see the method simply running Runnables which can contain anything you want. When reading "refresh" I would immediately associate it to refreshing the viewer and then discovering in code that's its running general purpose runnables in a asnync manner makes me go "oh, unexpected".
What about #runAsync?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizardPageModel.java
jcantrill: That is not the line I checked.  Adding it in now

jcantrill:
@adietish  Updated and pushed with changes per our IRC discussion.

jcantrill:
Merged into master and pushed

PR #688, opened 2015-07-03
[JBIDE-20181] Add m2e nature to openshift.client and openshift.client

dgolovin:


adietish:
why would you want m2e involved in Eclipse? I was leaving this out on purpose.
IMHO this brings no benefit to the table but only drawbacks:
slower builds and frequent error markers caused by the project getting out of sync and requiring maven/update. This is even backed by the fact that we have resource copying in this pom which m2e excludes from the build-lifecycle requiring you to build on the cmd-line anyhow

dgolovin:
Mine purpose is to unify import in eclipse across all jbosstools modules. Current approach is manual and when you want to contribute something you have to understand a lots of things related to importing/building/configuring/fixing problems in eclipse. What we are trying to do is using oomph installer to make it easier, so it would work like:
1. run oomph
2. select dev ide product
3. drop in jbosstools setup
4. select module you interested in contribution
5. run the installation
6. start to work

and that would be the same across all jbosstools modules. No manual steps and no extra questions :)

adietish:
ok +1

adietish:

maxandersen:
@dgolovin technically you wouldn't need to add m2e to the build to do this since oomph would import the pom.xml via m2e anyway would they not ?

@adietish if you are running with m2e and tycho-plugin installed then you should *not* need to build on cmd-line anymore. That was the case 2-3 years ago, but fixed now. Please try and let me know if still an issue.

PR #689, opened 2015-07-09
[JBIDE-20189] handle forbidden exceptions

jcantrill:
@adietish Please review.  I'm not sure if we should make a change in the content provider


adietish:
+1 GTLGM

jcantrill:
Merged to master and pushed

PR #691, opened 2015-07-10
[JBIDE-19759] added "new wizard" capability to v3 app wizard

adietish:


plugins/org.jboss.tools.openshift.client/.classpath
```diff
@@ -1,16 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
-	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry exported="true" kind="lib" path="lib/openshift-restclient-java-3.0.0-SNAPSHOT.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/commons-compress-1.8.1.jar"/>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry exported="true" kind="lib" path="lib/openshift-restclient-java-3.0.0-SNAPSHOT.jar" sourcepath="/openshift-restclient-java"/>
```
jcantrill: @adietish I like this reference being their permanently, but did you intend to check it in?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
```diff
@@ -0,0 +1,59 @@
...
+					if (tag.contains(filterText)) {
+						return true;
+					}
+				}
+				return false;
+			}
+		}, Boolean.FALSE);
+	}
+
+	public static IProject getProject(IResource resource) {
```
jcantrill: At a minimum, this should be pushed into the rest client.  Resources that are not projects already can get their project.  Looking at the code for Project, it currently returns null but probably should return itself

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
```diff
@@ -63,58 +68,108 @@ public Image getImage(Object element) {
...
-			if(prefix == null) {
-				prefix = "<unknown user>";
-			}
-			return style(prefix, conn.toString());
+			return getStyledText((Connection) element);
 		}
 		return super.getStyledText(element);
 	}
 
+	private StyledString getStyledText(IService service) {
```
jcantrill: Should we move these to some utility method to allow for testing and reduce the size of the class?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
```diff
@@ -63,58 +68,108 @@ public Image getImage(Object element) {
...
 			case ResourceKind.REPLICATION_CONTROLLER:
-				IReplicationController rc = (IReplicationController) resource;
-				return (style(resource.getName(), String.format("selector: %s", StringUtils.serialize(rc.getReplicaSelector()))));
+				return getStyledText((IReplicationController) resource);
 			case ResourceKind.SERVICE:
 				IService service = (IService) resource;
-				String serviceQualifiedText = String.format("selector: %s", StringUtils.serialize(service.getSelector()));
-				return style(service.getName(), serviceQualifiedText);
+				return getStyledText(service);
+			case ResourceKind.TEMPLATE:
```
jcantrill: For these to appear in the Explorer an additional ResourceGroup is necessary in the ContentProvider

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
```diff
@@ -44,6 +44,32 @@ public ImportApplicationWizard() {
 	}
 
 	@Override
+	public void init(IWorkbench workbench, IStructuredSelection selection) {
+		Connection connection = UIUtils.getFirstElement(selection, Connection.class);
+		if (connection != null) {
+			model.setConnection(connection);
+		} else {
+			IResource resource = UIUtils.getFirstElement(selection, IResource.class);
```
jcantrill: This block is basically all doing the same thing.  Maybe a private method that tries to get the first item that is not null based on order IResource, Project, Grouping and then you have a single place where the connection is retrieved and selected item is set.  Also, it should be you will NEVER get IProject without getting IResource since a project is a resource.  You might not need that block.

plugins/org.jboss.tools.openshift.client/.classpath
adietish: @jcantrill not rellay, a checking by mistake.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
adietish: +1 this sounds like a good idea.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
adietish: i agree on this but the client API has a gotcha that prevents uniformous treatement of resources and projects:
* on a IResource that is NOT a project you get the project via .getProject()
* on a IProject which IS a project you get null from .getProject(), you need to take the resource itself

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
adietish: I didnt intend to have these showing up, i only wanted to be able to use the same label provider in the application wizard to add visual consistency

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
jcantrill: @adietish hmm...if  IProject#getProject returns null, this is a bug in the restclient IMO and we should fix it there

adietish: +1

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
adietish: +1, done here: https://issues.jboss.org/browse/OSJC-194

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
adietish: @jcantrill nice! as I see the latest osrcj build has this fix, I removed the util method on my side

PR #697, opened 2015-07-16
[JBIDE-20290] dont refresh resource groupings for project (resources)

adietish:


adietish:
@jcantrill plz review

adietish:
simple fix, merging. Can still correct according to comments later.

PR #698, opened 2015-07-16
[JBIDE-20293] only display status msg when app creation fails (was: j…

adietish:
…son)


adietish:
simple fixed, merged to upstream/master without review

PR #699, opened 2015-07-16
[JBIDE-20266] updating list of templates when returning from "manage …

adietish:
…projects"


adietish:
rebased, merged and pushed to upstream/master and jbosstools-4.3.0.Beta2

PR #705, opened 2015-07-21
[JBIDE-20297] reloading projects after new project was created

adietish:


adietish:
@fbricon tested this, merging

PR #709, opened 2015-07-21
[JBIDE-20333] corrected method calls to BrowserUtility

adietish:


adietish:
the change in jbosstools-base that caused this was reverted: https://github.com/jbosstools/jbosstools-base/commit/2643f8c33295ccc739950def05b090b0ed1fd8bf
Closing without merging.

PR #713, opened 2015-07-26
[JBIDE-20377] Port forwarding disappears on refresh

jcantrill:


jcantrill:
@adietish please review..is this acceptable

adietish:
+1 LGTM
@jcantrill not sure if logging to console makes any big sense, only the sake of being the same as in v2 seems to justify it I guess. "oc" wont output anything that we could display, right?

jcantrill:
It logs to console because v2 logs to console.  We could log traffic from the port forward process as we have access to the input stream, but otherwise I find the console we create unnecessary..  Merging

adietish:
+1 since what I see so far is nothing but repeating what the code does, no added info, quite useless imho

PR #715, opened 2015-07-28
[JBIDE-20370] Display the OpenShift Explorer when the 'New Application' wizard completes

fbricon:

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizard.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizard.java
@@ -22,6 +22,8 @@ import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchWizard;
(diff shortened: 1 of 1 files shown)


jcantrill:
@fbricon LGTM.  Merge away....or is that something @adietish or I need to do?

fbricon:
I have god powers over all JBT repos, I can do it :-)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizard.java
```diff
@@ -43,7 +45,9 @@
  * @author Andre Dietisheim
  */
 public class NewApplicationWizard extends Wizard implements IWorkbenchWizard, IConnectionAwareWizard<Connection> {
-	
+
+	private static final String OPENSHIFT_EXPLORER_VIEW_ID = "org.jboss.tools.openshift.express.ui.explorer.expressConsoleView";
+
```
adietish: This very much looks like a constant that we should allow others to use. and thus should extract to some public class

fbricon: Indeed, but I had no other use case and didn't where else to put it. In the spirit of YAGNI, I left it here :-)

PR #718, opened 2015-07-30
[JBIDE-20404] Fix unable to show port forwarding wizard after previously stopping forwarding

jcantrill:
@fbricon please review


fbricon:
Seems legit AFAICT. +1

PR #723, opened 2015-08-06
JBIDE-20167 Openshift Tooling Usage Tracking.

alexeykazakov:
Replaced countEvent() by trackEvent(). So we will send all events right away instead of counting them for 24 hours.


adietish:
@alexeykazakov why should we? aren't we creating a lot of additional traffic which gets us even more over limit at Analytics?

PR #724, opened 2015-08-06
[JBIDE-20032] Use GIT_REF to check out imported application

fbricon:

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
@@ -86,8 +86,9 @@ public class ImportApplicationWizard extends Wizard implements IWorkbenchWizard,
 	private boolean importProject() {
 		try {
 			final DelegatingProgressMonitor delegatingMonitor = new DelegatingProgressMonitor();
-			IStatus jobResult = WizardUtils.runInWizard(
-					new ImportJob(model.getGitUrl(), model.getCloneDestination(), delegatingMonitor), delegatingMonitor, getContainer());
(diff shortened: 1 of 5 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
```diff
@@ -86,8 +86,9 @@ public boolean performFinish() {
 	private boolean importProject() {
 		try {
 			final DelegatingProgressMonitor delegatingMonitor = new DelegatingProgressMonitor();
-			IStatus jobResult = WizardUtils.runInWizard(
-					new ImportJob(model.getGitUrl(), model.getCloneDestination(), delegatingMonitor), delegatingMonitor, getContainer());
+			ImportJob importJob = new ImportJob(model.getGitUrl(), model.getCloneDestination(), delegatingMonitor)
+					.setGitRef(model.getGitRef());
```
adietish: I definitely would add gitRef to the constructor of the importJob

fbricon: I'm not a big fan of having constructors with huge lists of parameters, the builder pattern is much better suited for that case.

I added it as a setter because, another parameter needs to be injected next (an import filter a.k.a. contextdir). Injecting that as a constructor arg, would gives us a 5 args constructor. Meh.

adietish: ok, +1 then. I agree, 5 params is too much, should have a builder, but not worth the effort here.

fbricon: Cool. it also makes sense to keep the setters because gitref and contextdir are clearly optional parameters

fbricon:
Merged in master

PR #729, opened 2015-08-10
JBIDE-20372 - Connection wizard: selecting Openshift v3 now clears th…

rafabene:
…e server url


fbricon:
I can't vouch whether it's the proper solution or not, but it certainly works as expected.
LGTM but I'd prefer if @adietish or @jcantrill could +1 it. I can apply the change then.

adietish:
tested it, works for me. Thanks!

PR #730, opened 2015-08-10
JBIDE-20364 Connection wizard: all text (not just "retrieve") is clic…

dazarov:
…kable


fbricon:
It does the job, but that really looks like an ugly workaround.
-0 on applying it.
I'll leave it to @adietish or @jcantrill to decide.

adietish:
@dazarov thanks for the PR
Because of broken link widgets in MacOS dialogs we have to use styled text instead which then causes the issue you tried to tackle. Given that this issue exists in all places where we had to use styled text instead I'd strongly prefer a generic solution. If we use your current workaround we'd have to apply this approach to all occurrences and then, once the link widget gets fixed for MacOS we'd have a lot of fun getting back.
A generic solution could look like StyledTextUtils#isLinkClicked(label) which then checks if the click happened in the "<a>...</a>" (underline style-range) denoted area. The listener could then simply call the above and react upon.

PR #732, opened 2015-08-11
[JBIDE-20095] impl'ed validation for parameters

adietish:
…y trackEvent(). So we will send all events right away instead of counting them for 24 hours.


adietish:
@fbricon plz review

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/TemplateParametersPage.java
```diff
@@ -110,39 +125,116 @@ public void doubleClick(DoubleClickEvent event) {
 				.align(SWT.FILL, SWT.FILL).applyTo(resetButton);
 		resetButton.setText("Reset");
 		resetButton.addSelectionListener(onReset());
+
+		Label requiredExplanationLabel = new Label(templateParametersGroup, SWT.None);
+		requiredExplanationLabel.setText("* = value required, edit or double click to change them.");
```
fbricon: click the 'Edit...' button or double-click on a value to edit it.

fbricon:
LGTM, if you can change the label as commented above

PR #738, opened 2015-08-12
JBIDE-20355 and JBIDE-20371

rafabene:
JBIDE-20355 is ok through having the same validation code performed on FieldEditorPreferencePage#propertyChange method.

JBIDE-20371 is just initial since the Link content and behaviour needs to be implemented by someone with more knowledge about SWT and Eclipse API.


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/OpenShiftPreferencePage.java
```diff
@@ -63,8 +66,17 @@ public void createFieldEditors() {
 		this.cliLocationEditor.setFilterPath(SystemUtils.getUserHome());
 		this.cliLocationEditor.setFileExtensions(EXTENSIONS);
 		addField(cliLocationEditor);
-		
-	}
+		//TODO Fix the Layout
+        Link link = new Link(getFieldEditorParent(), SWT.WRAP );
+        link.setText("Openshift plugin needs the openshift client binary (CLI) to provide some features like port forwarding.<br>"
```
fbricon: The OpenShift Client binary (oc) is required for features such as Port Forwarding or Log Streaming.

PR #739, opened 2015-08-13
[JBIDE-20455] open oc file chooser at location set in "oc location"

adietish:


adietish:
@fbricon pls review

fbricon:
Browse location is properly initialized

One weird thing though: if you select a file, then change the value in the textbox, to something non-existant, validation doesn't occur when losing focus. when re-entering focus then, validation kicks in

adietish:
I noticed the weirdness but didnt find a solution in a reasonable fashion

adietish:
merged and pushed to upstream/master

PR #742, opened 2015-08-13
Remove .project and .classpath from git repo

rafabene:


jcantrill:
I would be on board with this but will defer to @adietish and @fbricon  judgement.  @adietish  has mentioned its necessary because of the difficulties in spinning up a new dev to do development

jcantrill:
@adietish Can you comment on this or close if its appropriate

fbricon:
We're gonna keep the Eclipse settings files, since we're only developing using Eclipse it won't hurt the sensibility of other IDE users. It also allows people allergic to m2e to be able to import the plugins in Eclipse.

Thus rejecting this PR.

PR #744, opened 2015-08-14
[JBIDE-20463] reset token value when new connection is selected

fbricon:

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/ConnectionEditor.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/ConnectionEditor.java
@@ -25,7 +25,6 @@ import org.eclipse.jface.layout.GridLayoutFactory;
 import org.eclipse.jface.viewers.ArrayContentProvider;
 import org.eclipse.jface.viewers.ColumnLabelProvider;
 import org.eclipse.jface.viewers.ComboViewer;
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/ConnectionEditor.java
```diff
@@ -223,8 +222,7 @@ private void disposeBindings() {
 	@Override
 	protected void onSelectedConnectionChanged(IObservableValue selectedConnection) {
 		IConnection conn = (IConnection) selectedConnection.getValue();
-		if(!(conn instanceof Connection)) return;
-		detailViewModel.setSelectedConnection((Connection)conn);
+		detailViewModel.setSelectedConnection(conn);
```
jcantrill: Is there a reason to remove the instance check?

jcantrill:
@fbricon assuming it resolves the issue, it LGTM

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/ConnectionEditor.java
fbricon: yes, setSelectedConnection takes an IConnection argument, and the instance check prevents NewConnectionMarker instances to be set, which leads to the UI not being reset.

fbricon:
Applied in master

PR #745, opened 2015-08-14
[JBIDE-20464] opening oauth login in ext. browser (not dialog) in MacOS

adietish:


adietish:
@fbricon pls review

fbricon:
Couldn't we just get the response of the token url into a string, then parse it to extract the token (response.substring(indexOf("--token="), indexOf(<following space>)), put it directly into the textbox?

adietish:
@fbricon nope since the user still has to do the oauth login/auth in the browser (RH SSO in our stg server)

fbricon:
Applied in master

PR #747, opened 2015-08-17
[JBIDE-19584] Hide 'New Connection' menu from OpenShiftException nodes in OS Explorer

fbricon:
I don't know how we can get the parent Connection node from the error node, from the ISelection, 
so I didn't add the Refresh menu

At least hiding the new Connection menu from error nodes should be enough for now


jcantrill:
LGTM

fbricon:
Added Refresh menu on OpenShiftException node

PR #748, opened 2015-08-17
[JBIDE-20053] reset all views when using a new connection

fbricon:


adietish:
+1, LGTM

PR #751, opened 2015-08-19
JBIDE-20324 - command handler should only be taking action if selecti…

robstryker:
…on is adaptable

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/command/RemoveConnectionHandler.java
+++ b/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/command/RemoveConnectionHandler.java
@@ -31,6 +31,9 @@ public class RemoveConnectionHandler extends AbstractHandler {
 	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
 		IConnection[] connections = UIUtils.getElements(HandlerUtil.getCurrentSelection(event), IConnection.class);
+		if( connections.length == 0 )
+			return null;
+
 		if (MessageDialog.openConfirm(HandlerUtil.getActiveShell(event)
 				, "Remove connection"
 				, NLS.bind("You are about to remove the connection(s):\n{0}\n\n"
@@ -41,14 +44,16 @@ public class RemoveConnectionHandler extends AbstractHandler {
 
 									@Override
 									public String toString(IConnection connection) {
-										return "\n" + connection.toString();
+										return connection == null ? "" : ("\n" + connection.toString());
 									}
 								})
 						)
 				)) {
 			for(IConnection connection : connections){
-				ConnectionsRegistrySingleton.getInstance().remove(connection);
+				if( connection != null )
+					ConnectionsRegistrySingleton.getInstance().remove(connection);
 			}
+
 		}
 		return null;
 	}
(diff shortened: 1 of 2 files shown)


plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/command/RemoveConnectionHandler.java
```diff
@@ -41,7 +44,7 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
 
 									@Override
 									public String toString(IConnection connection) {
-										return "\n" + connection.toString();
+										return connection == null ? "\nnull" : ("\n" + connection.toString());
```
fbricon: there's no point in removing a null connection.
The confirmation dialog should be opened only if the list of connections contains valid values.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/command/RemoveConnectionHandler.java
```diff
@@ -31,6 +31,9 @@
 	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
 		IConnection[] connections = UIUtils.getElements(HandlerUtil.getCurrentSelection(event), IConnection.class);
+		if( connections.length == 0 )
+			return null;
```
fbricon: null connections should be filtered out.

If resulting list is empty, then return

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/command/RemoveConnectionHandler.java
```diff
@@ -31,6 +31,9 @@
 	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
 		IConnection[] connections = UIUtils.getElements(HandlerUtil.getCurrentSelection(event), IConnection.class);
+		if( connections.length == 0 )
```
fbricon: if (connections == null || connections.length == 0)

robstryker: Not needed, since UIUtils is guaranteed to return a non-null array.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/command/RemoveConnectionHandler.java
robstryker: Null connections are now filtered out by UIUtils, but it's still possible (somehow) that something's slipped through, so I just add that as a safety. But it's extremely unlikely it'd ever be hit.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/command/RemoveConnectionHandler.java
robstryker: It is now assumed that the connection list is valid (based on the changes in UIUtils), but I simply add the 'null' as a safety to prevent NPE.

fbricon: OK, if documented in UIUtils API

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/command/RemoveConnectionHandler.java
fbricon: Not OK. In the unlikely event that happens you shouldn't even open the dialog. filter filter filter

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/command/RemoveConnectionHandler.java
fbricon: OK, if documented in UIUtils API

fbricon:
Merged as f1f5ceb9f2f06f1f0b2348862c87f4350acb2027, after some tweaking ;-)

PR #762, opened 2015-08-26
[JBIDE-19949] Automatically import sources once an application is created

fbricon:

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
@@ -8,7 +8,9 @@
  ******************************************************************************/
 package org.jboss.tools.openshift.internal.ui.wizard.importapp;
 
+import java.util.Collection;
 import java.util.Collections;
+import java.util.Map;
 
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.jface.dialogs.ErrorDialog;
@@ -28,6 +30,7 @@ import org.jboss.tools.openshift.internal.common.ui.wizard.IConnectionAwareWizar
 import org.jboss.tools.openshift.internal.ui.OpenShiftUIActivator;
 import org.jboss.tools.openshift.internal.ui.explorer.ResourceGrouping;
 
+import com.openshift.restclient.model.IBuildConfig;
 import com.openshift.restclient.model.IProject;
 import com.openshift.restclient.model.IResource;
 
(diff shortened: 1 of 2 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
```diff
@@ -47,8 +50,28 @@ public ImportApplicationWizard() {
 		this.model = new ImportApplicationWizardModel();
 	}
 
+	public ImportApplicationWizard(Map<IProject, Collection<IBuildConfig>> projectsAndBuildConfigs) {
+		this();
+		if (projectsAndBuildConfigs != null && projectsAndBuildConfigs.size() == 1) {
+			Map.Entry<IProject, Collection<IBuildConfig>> entry = projectsAndBuildConfigs.entrySet().iterator().next();
+			IProject project = entry.getKey();
+			Connection connection = ConnectionsRegistryUtil.safeGetConnectionFor(project);
```
jcantrill: Presumably this should not be null but my only concern would be the case where the connection is not found; should we use the alternate getter?  Do we need to explicitly handle that case or display something to a user?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
```diff
@@ -47,8 +50,28 @@ public ImportApplicationWizard() {
...
+			setModelConnection(connection);
+			Collection<IBuildConfig> buildConfigs = entry.getValue();
+			if (buildConfigs.size() == 1) {
+				model.setSelectedItem(buildConfigs.iterator().next());
+			} else {
+				model.setSelectedItem(project);
```
jcantrill: What are we doing here?  Either we set to a buildconfig or project?  How is that manifested to a user? The buildconfig or the project is selected?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizard.java
```diff
@@ -148,6 +168,25 @@ public void run() {
 		return success;
 	}
 
+	protected Map<IProject, Collection<IBuildConfig>> getBuildConfigs(Collection<IResource> resources) {
+		Map<IProject, Collection<IBuildConfig>> projects = new LinkedHashMap<>();
+		for (IResource resource : resources) {
+			if (resource instanceof IBuildConfig) {
```
jcantrill: You can actually ask the kind of a resource via resource#getKind()  Not sure it really matters here though

jcantrill:
My comments are just some items to note, not necessarily change requests.  LGTM

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
fbricon: if 1 buildconfig is found, then we skip the buildconfig selection page and go directly to the git clone page.
If more buildconfigs are found, we select the parent project and display the buildconfig selection page (see addPages()). The user will have to choose what to import

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizard.java
fbricon: Yeah I don't think that makes any difference here

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/importapp/ImportApplicationWizard.java
fbricon: If that blows here, then I don't know how to gracefully handle that. It's not any different that what is done in init().
I'd rather keep it as is, for the sake of simplicity.

PR #765, opened 2015-08-27
[JBIDE-20502] Add 'Show In > Web Console' command

fbricon:


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/OpenInWebConsoleHandler.java
```diff
@@ -0,0 +1,76 @@
...
+		}
+		if (connection != null) {
+			String url = getWebConsoleUrl(connection, resource);
+			openInBrowser(url);
+		}
+		return null;
+	}
+
+	private String getWebConsoleUrl(Connection connection, IResource resource) {
+		StringBuilder url = new StringBuilder(connection.getHost()).append("/console");
```
jcantrill: This concerns me a little.  I didnt think about it for the case that @adietish initially did, but what if the path moves.  The tooling will be broken

liggitt: yeah, `/console` is configurable... the console may not even be hosted on the same server

maxandersen:
any API's available to locate the right url or do we need to hardcode assumptions ?

liggitt:
the APIs don't know anything about the web console. there can be 0, 1, or many web consoles surfacing the API content

jcantrill:
@fbricon i think the best we could do is provide some advanced connection configuration where you could modify the url to the console away from the default of 'console'. enhancement?

fbricon:
@jcantrill yes that was my idea too. Question is, should merge the current PR now or wait to get a more sophisticated mechanism?

jcantrill:
@fbricon I would add an enhancement for the work and merge.  There are many things that could happen but this works and supports getting to the console now.

fbricon:
Applied in master

PR #776, opened 2015-09-14
[JBIDE-20599] Allow the handler to set the connection instead of recent

jcantrill:
@fbricon review?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImagePage.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImagePage.java
@@ -10,8 +10,6 @@
  ******************************************************************************/
 package org.jboss.tools.openshift.internal.ui.wizard.deployimage;
 
-
-
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.core.databinding.Binding;
 import org.eclipse.core.databinding.DataBindingContext;
@@ -28,6 +26,8 @@ import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.jface.layout.GridLayoutFactory;
 import org.eclipse.jface.viewers.ComboViewer;
 import org.eclipse.jface.viewers.StructuredViewer;
+import org.eclipse.jface.viewers.Viewer;
+import org.eclipse.jface.viewers.ViewerSorter;
 import org.eclipse.jface.wizard.IWizard;
 import org.eclipse.linuxtools.docker.core.IDockerConnection;
 import org.eclipse.osgi.util.NLS;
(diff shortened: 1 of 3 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImagePage.java
```diff
@@ -231,11 +231,18 @@ private void createProjectControl(Composite parent, DataBindingContext dbc) {
...
+		final OpenShiftExplorerLabelProvider labelProvider = new OpenShiftExplorerLabelProvider();
 		cmboProject.setContentProvider(new ObservableListContentProvider());
-		cmboProject.setLabelProvider(new OpenShiftExplorerLabelProvider());
+		cmboProject.setLabelProvider(labelProvider);
 		cmboProject.setInput(
 				BeanProperties.list(IDeployImagePageModel.PROPERTY_PROJECTS).observe(model));
-		
+		cmboProject.setSorter(new ViewerSorter() {
+			@Override
+			public int compare(Viewer viewer, Object e1, Object e2) {
```
fbricon: shouldn't `default` always be on top?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
```diff
@@ -81,7 +81,6 @@
 	private boolean originatedFromDockerExplorer;
 	
 	public DeployImageWizardModel() {
-		setConnection(ConnectionsRegistrySingleton.getInstance().getRecentConnection(Connection.class));
```
fbricon: why is this removed?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImagePage.java
jcantrill: It should, but maybe we should consider removing it all together.  I'm not sure how the current ording works in the explorer window.  I assume it returns in order of what was returned by the call.  I dont know whay it wouldnt do the same here.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
jcantrill: This is what was causing the problem.  It returns the one recently accessed but we want the one you selected.  So I defered it to being set in the handler. Hmm... maybe we need to explicitly set 'recent' when calling from the docker explorer side?

fbricon:
Applied in master and 4.3.x

PR #777, opened 2015-09-14
bump client plugin to use openshift-rest-client 3.1.0-SNAPSHOT

jcantrill:
@fbricon not sure if you want to wait and bump this with the plugin versions?


fbricon:
@nickboldt when are we supposed to bump the eclipse plugins version in master?

nickboldt:
Whenever you have to mark master different from 4.3.x.

fbricon:
Merged in master and 4.3.x

PR #785, opened 2015-09-29
JBIDE-20834 - Fix issues with styled label providers

snjeza:
https://issues.jboss.org/browse/JBIDE-20834
Fix issues with styled label providers


fbricon:
@adietish I'll let you decide if it's worth adding to 4.3.x too. I don't think it's necessary, but if it works for Mars.1, then it's probably ok

adietish:
i was waiting for this since there was no urgency. But I suspect this to fix JBIDE-21141, verifying now that I have a possible reason

adietish:
According to https://bugs.eclipse.org/bugs/show_bug.cgi?id=466499 this is only fixed in 4.5.1, we can thus only merge this to our master, not to 4.3.x

adietish:
merged and pushed to jbosstools-4.3.x

PR #787, opened 2015-10-02
[JBIDE-20905] resetting validation status (& deco) upon template change

adietish:


fbricon:
+1 to apply in master only. Generified API depends on Neon

PR #793, opened 2015-10-20
[JBIDE-20531] Add ports during deploy image work flow

jcantrill:
PR provides:
* Expose additional and/or edit ports
* Format fixes

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
@@ -24,7 +24,6 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.IJobChangeEvent;
-import org.eclipse.core.runtime.jobs.IJobChangeListener;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.runtime.jobs.JobChangeAdapter;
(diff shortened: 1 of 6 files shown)


jcantrill:
@adietish @fbricon  please review at your leisure

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
```diff
@@ -383,6 +382,25 @@ public void addEnvironmentVariable(String key, String value) {
 		this.environmentVariables.add(new EnvironmentVariable(key, value, true));
 		firePropertyChange(PROPERTY_ENVIRONMENT_VARIABLES, old, Collections.unmodifiableList(environmentVariables));
 	}
+	
+	
+
+	@Override
+	public void addServicePort(IServicePort port) {
+		List<IServicePort> old = new ArrayList<>(this.servicePorts);
```
fbricon: so this allows potential duplicate IServicePort. should we worry?

fbricon:
Other than the question above, I have no other remarks. LGTM

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
jcantrill: Technically yes even though the calling code has protections.  We could modify it to depend on Collections and utilize a Set?

fbricon: since it's a public method, better guard against whatever is thrown at it here

PR #800, opened 2015-10-27
[JBIDE-20764] implemented server adapter for openshift v3

adietish:


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerExtendedProperties.java
```diff
@@ -0,0 +1,32 @@
...
+	}
+	
+	public boolean hasWelcomePage() {
+		return true;
+	}
+	
+	public String getWelcomePageUrl() {
+		return null;
+	}
+}
```
fbricon: add the following so that the XML Configuration and Filesets noded don't show up:

	
	@Override
	public boolean allowConvenienceEnhancements() {
		return false;
	}

See https://issues.jboss.org/browse/JBIDE-21116

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerExtendedProperties.java
```diff
@@ -0,0 +1,37 @@
...
+	@Override
+	public boolean allowConvenienceEnhancements() {
+		return false;
+	}
+
+	public boolean hasWelcomePage() {
+		return true;
+	}
+
+	public String getWelcomePageUrl() {
```
fbricon: This seems to work:

```
	public boolean hasWelcomePage() {
		return getWelcomePageUrl() != null;
	}

	public String getWelcomePageUrl() {
		IService service = OpenShiftServerUtils.getService(server);
		if (service != null) {
			IProject project = service.getProject();
			if (project != null) {
				List<IRoute> routes = project.getResources(ResourceKind.ROUTE);
				if (routes != null && !routes.isEmpty()) {
					//if more than 1 route, well, we're not in the UI layer here, 
					//so can't let the user select the url to open, tough luck. 
					//So we pick the 1st one we find
					return routes.get(0).getURL();
				}
			}
		}
		return null;
	}
```

adietish: great, thanks!

PR #809, opened 2015-11-06
[JBIDE-21061] notify upon chaning server-/local template-tab

adietish:


adietish:
rebased, merged and pushed to jbosstools-4.3.x

PR #810, opened 2015-11-11
WIP - DO NOT MERGE

jcantrill:
@fbricon @adietish  Please take an initial look.  Things that need to be addressed:
* Had to keep reference in the manager in order to be able to fire connection changed events
** Seems this could get out of sync with the 'cache' that is kept by OpenShiftExplorerContentProvider
* Does it need functionality to enable/disable watches or start/stop? If so, where is this?
* Do we need an overlay to distinguish manual refresh vs an existing watch?
* The refresh is choppy when you add or delete.  We have this same problem with refresh, but it would be nice if the node didnt collapse and expand
* I think I saw cases where the server indicated a completed build but I didnt see that refresh in eclipse.  Something to think about


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -0,0 +1,121 @@
...
+public class WatchManager {
+
+	private static final Collection<String> WATCH_KINDS = Arrays.asList(ResourceKind.BUILD, ResourceKind.POD);
+	private static final String DELIMITER = "|";
+	private static WatchManager instance;
+	private static final Logger LOGGER = LoggerFactory.getLogger(WatchManager.class);
+	
+	private Map<String, IWatcher> watches = Collections.synchronizedMap(new HashMap<>());
+
+	public static WatchManager getInstance() {
```
fbricon: either use `synchronized` or better, the Initialization On Demand Holder pattern : http://blog.crazybob.org/2007/01/lazy-loading-singletons.html

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -0,0 +1,121 @@
...
+		public void received(IResource resource, ChangeType change) {
+			LOGGER.debug("Watch received change '{}' for {}", change, resource.toJson());
+			IResource newItem = null;
+			IResource oldItem = null;
+			switch(change) {
+			case ADDED:
+				resources.add(resource);
+				newItem = resource;
+				break;
+			case DELETED:
```
fbricon: better check resources contains the resource in the 1st place. Unlikely to happen but if it does then, bam, indexoutofboundexception in your face

fbricon:
So, I *think*, from a UX perspective, having the resources being watched by default would be expected. I'd be great if the user didn't have to worry about having to trigger watching or not.
That means that whenever the connection is live, resources would be automatically watched.

Now this might prove difficult (I don't know) to implement/setup. Anyway, there may be cases when watching can not be done automatically, thus a user'd have to trigger it manually. In that case, I think an overlay (like an eye) on watched resources could be a good idea. If a resource is being watched, then add a menu to stop watching, else to start watching

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -0,0 +1,121 @@
...
+		}
+		@Override
+		public void connected() {
+			LOGGER.debug("The watch connected to OpenShift connection {}", conn.toString());
+		}
+
+		@Override
+		public void disconnected() {
+			LOGGER.debug("The watch disconnected to OpenShift connection {}", conn.toString());
+			IWatcher watcher = watches.remove(key);
```
fbricon: sometimes returns null, causes NPE next line

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -0,0 +1,121 @@
...
+			watcher.stop();
+		}
+
+		@Override
+		public void error(Throwable err) {
+			LOGGER.error("There was an error watching connection {}",err, conn.toString());
+			IWatcher watcher = watches.remove(key);
```
fbricon: sometimes returns null, causes NPE next line

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -0,0 +1,121 @@
...
+			switch(change) {
+			case ADDED:
+				resources.add(resource);
+				newItem = resource;
+				break;
+			case DELETED:
+				oldItem = resources.remove(resources.indexOf(resource));
+				break;
+			case MODIFIED:
+				oldItem = resources.remove(resources.indexOf(resource));
```
fbricon: what happens if oldItem is null?

jcantrill:
@fbricon I think watch when opening a connection would be ideal as well, but I have concerns about the cases of a user who has a large number of projects.  Ideally we could selectively allow a user to start or stop project watches.  That will need to be an enhancement and probably require a project overlay of some kind to properly indicate whats going on.

cc @adietish @maxandersen  review at your first opportunity.

jcantrill:
Updated to use job.schedule instead of sleep

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -0,0 +1,227 @@
...
+		}
+
+		@Override
+		public void error(Throwable err) {
+			OpenShiftCoreActivator.logWarning("Reconnecting. There was an error watching connection " + conn.toString(), err);
+			restart();
+		}
+		
+		private synchronized void restart() {
+			if(state.get() == State.DISCONNECTED) {
```
adietish: any chance a job gets into disconnected state other than your set further below?

PR #818, opened 2015-11-20
JBIDE-21116 - openshift server adapter doesn't need or benefit from 

robstryker:
JBIDE-21116 - openshift server adapter doesn't need or benefit from xml / fileset enhancements

### DIFF
--- a/plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/server/OpenShiftServerExtendedProperties.java
+++ b/plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/server/OpenShiftServerExtendedProperties.java
@@ -24,6 +24,10 @@ public class OpenShiftServerExtendedProperties extends ServerExtendedProperties 
 		super(adaptable);
 	}
 	
+	public boolean allowConvenienceEnhancements() {
+		return false;
+	}
+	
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/server/OpenShiftServerExtendedProperties.java
```diff
@@ -24,6 +24,10 @@ public OpenShiftServerExtendedProperties(IAdaptable adaptable) {
 		super(adaptable);
 	}
 	
+	public boolean allowConvenienceEnhancements() {
```
maxandersen: what does 'convenienceEnhancements" mean ? is that really our name for XML and File features ?

robstryker: Yes, that is really our name for it.  Server Extended Properties was never really meant to be external API. In reality this method shouldnt even be on the extended properties because its more of a UI customization and has nothing to do with the server itself specifically. This method was only added to allow others who DID add an extended-properties to THEN opt out of the fileset/xpaths (seemingly a long time ago actually). 

But yeah, thats the name.

PR #819, opened 2015-11-25
JBIDE-20990 - Introduction of CDK adapter,

robstryker:
with integration to openshift v3 and docker.
Details below:

Renamed Plugin

Added runtime detection

Move startup and shutdown logic into subsystem controllers

Added launch tab group and streamlined some of the startup code to more properly fit a user-friendly paradigm

Bug in Vagrant poller when Vagrantfile is configured to expect credentials via environment variables but server adapter is configured for terminal to prompt for credentials over TTY

Current logic overwrote environment map entirely rather than just the two properties we'd like to add

Lots of cleanup:
  allow cdk server to use new credential framework,
  provide support for virtualbox 'poweroff' state instead of just 'shutoff'

Add a warning alerting users that passwords may not be shown in launch tabs

Fix server adapter to handle changes in credential  model

Add listener for server state change to add an openshift connection if applicable

added warnigns if credentials are missing
added credential composite to wizard fragment
made sure widget enablement in server editor is correct

Isolate code for external launch
add support for adbinfo

Integration with docker


fbricon:
Merged in master

PR #820, opened 2015-11-25
JBIDE-20990 - Introduction of CDK adapter,

robstryker:
with integration to openshift v3 and docker.
Details below:

Renamed Plugin

Added runtime detection

Move startup and shutdown logic into subsystem controllers

Added launch tab group and streamlined some of the startup code to more properly fit a user-friendly paradigm

Bug in Vagrant poller when Vagrantfile is configured to expect credentials via environment variables but server adapter is configured for terminal to prompt for credentials over TTY

Current logic overwrote environment map entirely rather than just the two properties we'd like to add

Lots of cleanup:
  allow cdk server to use new credential framework,
  provide support for virtualbox 'poweroff' state instead of just 'shutoff'

Add a warning alerting users that passwords may not be shown in launch tabs

Fix server adapter to handle changes in credential  model

Add listener for server state change to add an openshift connection if applicable

added warnigns if credentials are missing
added credential composite to wizard fragment
made sure widget enablement in server editor is correct

Isolate code for external launch
add support for adbinfo

Integration with docker

Adding feature, updating category xml, etc


plugins/org.jboss.tools.openshift.egit.ui/src/org/jboss/tools/openshift/egit/ui/util/EGitUIUtils.java
```diff
@@ -43,7 +43,7 @@ public void execute(Repository repository, IProgressMonitor monitor) throws Core
 	};
 	
 	public static String getEGitDefaultRepositoryPath() {
-		return Activator.getDefault().getPreferenceStore().getString(UIPreferences.DEFAULT_REPO_DIR);
+		return RepositoryUtil.getDefaultRepositoryDir();
```
fbricon: this doesn't compile in 4.3.x. How come that change managed to creep here?

fbricon:
Merged in 4.3.x

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
```diff
@@ -0,0 +1,212 @@
...
+	
+	private ADBInfo loadADBInfo(IServer server) {
+		try {
+			ILaunchConfigurationWorkingCopy wc = new CDKLaunchConfigUtility().
+					createExternalToolsLaunchConfig(server, "adbinfo", server.getName() + " adbinfo");
+			ILaunch launch2 = wc.launch("run", new NullProgressMonitor());
+			final IProcess[] processes = launch2.getProcesses();
+			IStreamMonitor mon = processes[0].getStreamsProxy().getOutputStreamMonitor();
+			final ArrayList<String> lines = new ArrayList<String>();
+			mon.addListener(new IStreamListener() {
```
scabanovich: Is it guaranteed that text is always appended in complete lines? Maybe it makes sense to check if appended text ends with line separator, and also split text to lines to add all true lines into the array at this moment. If text does not end with line separator, then the last line is not complete, it should be saved into a temporary string and concatenated with first line in the next contribution.
Or, to make things simple and safe, all text contributions may be added to StringBuilder rather than to List.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
```diff
@@ -0,0 +1,212 @@
...
+			HashMap<String,String> env = new HashMap<String,String>();
+			Iterator<String> lineIterator = lines.iterator();
+			while(lineIterator.hasNext()) {
+				String oneAppend = lineIterator.next();
+				String[] allAppends = oneAppend.split("\n");
+				for( int i = 0; i < allAppends.length; i++ ) {
+					if( allAppends[i].trim().startsWith("export ")) {
+						int eq = allAppends[i].indexOf("=");
+						if( eq != -1 ) {
+							String k = allAppends[i].substring("export ".length(), eq);
```
scabanovich: Check above calls trim() but key extraction does not. Both should either to call trim() or not do it.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
scabanovich: Imho, by the time when a listener is added to the monitor, there may be some output in it already, that can be obtained as IStreamMonitor.getContents().

PR #828, opened 2015-12-02
Update CDK server icon

robstryker:


plugins/org.jboss.tools.openshift.cdk.server/build.properties
```diff
@@ -12,4 +12,5 @@ bin.includes = META-INF/,\
                about.properties,\
                jboss_about.png,\
                pom.xml,\
-               .
+               .,\
```
fbricon: this is not necessary anymore, see previous commit

PR #831, opened 2015-12-03
JBIDE-21175 - ensure more reliable locating of vagrant command

robstryker:


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/CDKConstantUtility.java
```diff
@@ -10,29 +10,143 @@
...
+	private static final String WINDOWS_WHERE_VAGRANT = "where vagrant";
+	
+	
+	private static String discoveredVagrantLocation = null;
+	private static boolean searchFailed = false;
+	
+	
 	public static String getVagrantLocation() {
-		return VAGRANT_LOCATION_LINUX;
+		return findVagrantLocation();
```
maxandersen: im wondering if we are better of launching this similar to what @gercan and @ibuziuk been doing for npm and cordova cli tools.

i.e. call out via bash or cmd.exe making it up to the user to have it properly configured in PATH

robstryker: There are times we call out via Runtime.exec() in which case the existing path is in control.  But there are also times we use the external tools launch configuration, in which case we need to know the full location.  In general, we use the launch config when we want a console to appear with the output, and we use Runtime.exec() when we just want it done in the background.

robstryker: Did some perf-testing on windows and linux.  This method is only a little bit slow when it has to go out to which / where,   and it will only perform this once per workspace, and only if vagrant isn't found in the path already.  Everything's already cached, and it won't try a second time, so the perf. hit is very very very low and the ability to work around path errors is tremendous.  

I believe this patch is for the best.

fbricon:
I think looking up the executable through which/%PATH% lookup.

It's should prolly be refactored in a utility class though, as we already do the same thing to get the oc binary in https://github.com/jbosstools/jbosstools-openshift/blob/29bdf8867d4f8e3623a50ecd7b80d0ef7b775ae6/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/OpenShiftPreferencePage.java#L56

maxandersen:
please check with @ibuziuk and @gercan how they do it for npm/etc.

Might be useful here too.

PR #832, opened 2015-12-03
JBIDE-21175 - ensure more reliable locating of vagrant command

robstryker:


plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
```diff
@@ -0,0 +1,154 @@
...
+			String windowsDefaultLoc, String linuxDefaultLoc) {
+		this.windowsCommand = windowsCommand;
+		this.linuxCommand = linuxCommand;
+		this.linuxDefaultLoc = linuxDefaultLoc;
+		this.windowsDefaultLoc = windowsDefaultLoc;
+	}
+	
+	
+	public  String getLocation() {
+		return findLocation(-1);
```
fbricon: I'd rather not allow infinite timeout

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/CDKConstantUtility.java
```diff
@@ -10,29 +10,49 @@
...
-		return VAGRANT_LOCATION_LINUX;
+		return findVagrantLocation();
+	}
+	
+	private static String findVagrantLocation() {
+		if( locationProvider == null ) {
+			locationProvider = new CommandLocationProvider("vagrant.exe", "vagrant", 
+					VAGRANT_LOCATION_WINDOWS, VAGRANT_LOCATION_LINUX);
+		}
+		return locationProvider.getLocation(10 * 1000);
```
fbricon: search on FS should be real fast. I'd put a much lower timeout, like 1s

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
```diff
@@ -0,0 +1,154 @@
...
+		return searchPath(System.getenv("PATH"), ";", windowsCommand);
+	}
+
+	private  String searchPathLinux() {
+		return searchPath(System.getenv("PATH"), ":", linuxCommand);
+	}
+	
+	private  String searchPath(String path, String delim, String commandName) {
+		String[] roots = path.split(delim);
+		for( int i = 0; i < roots.length; i++ ) {
```
fbricon: indexed based for loops?

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
```diff
@@ -0,0 +1,154 @@
...
+			if( new File(roots[i], commandName).exists()) {
```
fbricon: no need to instanciate same file twice

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
```diff
@@ -0,0 +1,154 @@
...
+	public String getLocation(int timeout) {
+		return findLocation(timeout);
+	}
+	
+	private  String findLocation(int timeout) {
+		if( discoveredLocation != null || searchFailed )
+			return discoveredLocation;
+
+		// Vagrant hasn't been discovered yet. Let's find it. 
+		if (Platform.getOS().equals(Platform.OS_WIN32)) {
```
fbricon: no WIN64??

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/CDKConstantUtility.java
```diff
@@ -10,29 +10,49 @@
...
-			ILaunchConfiguration lc = server.getLaunchConfiguration(false, new NullProgressMonitor());
-			if( lc != null ) {
-				return lc.getAttribute(IExternalLaunchConstants.ATTR_LOCATION, getVagrantLocation());
+		if( server != null ) {
+			try {
+				ILaunchConfiguration lc = server.getLaunchConfiguration(false, new NullProgressMonitor());
+				if( lc != null ) {
+					String ret = lc.getAttribute(IExternalLaunchConstants.ATTR_LOCATION, (String)null);
+					if( ret != null )
+						if( new File(ret).exists())
```
fbricon: if (ret != null && new File(ret).exists())

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
fbricon: You can use SystemUtils.IS_OS_WINDOWS

fbricon:
I think it'd be more elegant to provide a static method CommandLocationProvider.get() that would return the proper lookup strategy, based on the underlying os. Then provider.search(command, defaultCommand)

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
```diff
@@ -0,0 +1,154 @@
...
+	
+	
+	private String discoveredLocation = null;
+	private boolean searchFailed = false;
+	
+	private String linuxCommand, windowsCommand;
+	private String windowsDefaultLoc, linuxDefaultLoc;
+	
+	public CommandLocationProvider(String windowsCommand, String linuxCommand,
+			String windowsDefaultLoc, String linuxDefaultLoc) {
```
fbricon: If we need to have specific lookup strategies for OSX (eg. the binary name is different from linux) you have to add more parameters. That's pretty smelly

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
```diff
@@ -0,0 +1,154 @@
...
+		}
+		
+		if( discoveredLocation == null ) {
+			searchFailed = true;
+		}
+		return discoveredLocation;
+	}
+	
+	private  String findLocationWindows(int timeout) {
+		if( windowsDefaultLoc != null && new File(windowsDefaultLoc).exists()) {
```
fbricon: so if you provide a default location, you always skip the lookup, that sounds wrong. This should happen as last resort

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
```diff
@@ -0,0 +1,154 @@
...
+			return windowsDefaultLoc;
+		}
+		String ret = searchPathWindows();
+		if( ret == null ) {
+			ret = runCommand(WINDOWS_WHERE + windowsCommand, timeout);
+		}
+		return ret;
+	}
+	private  String findLocationLinux(int timeout) {
+		if( linuxDefaultLoc != null && new File(linuxDefaultLoc).exists()) {
```
fbricon: so if you provide a default location, you always skip the lookup, that sounds wrong. This should happen as last resort

robstryker: I disagree completely. If you provide a default location AND it exists, its probably the right location. It's most likely where the binary was installed.  There's no reason to continue searching if it's already in the expected location.

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
robstryker: Yes, I like index-based for-loops ;) Helps me when debugging.

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
robstryker: k good point

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/CDKConstantUtility.java
robstryker: code snob

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/CDKConstantUtility.java
robstryker: The timeout is there because it may run a which / where command if it's not found on the path or at default location,.

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
robstryker: Good point.  I'll rework the solution.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/CDKConstantUtility.java
fbricon: I know, but which/where should return *fast* anyway, so if it takes more than 1 sec, then it's doomed anyway

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
robstryker: I'll allow it but not default it.

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationProvider.java
robstryker: There's no WIN64 constant in Platform.   I'd rather not use SystemUtils as that's an org.apache class and I don't feel like depending on that plugin.  I prefer to use the eclipse-level constants.  I trust them more. shrug.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/OpenShiftPreferencePage.java
```diff
@@ -126,7 +63,16 @@ public String getName() {
 			return extensions;
 		};
 
-		public abstract String getLocation();
+		public String getLocation() {
+			if( locationProvider == null ) {
+				 // TODO fill in windows default loc
+				String[] platforms = new String[]{Platform.OS_LINUX, Platform.OS_WIN32};
+				String[] cmds = new String[]{"oc","oc.exe"};
+				String[] defLocs = new String[]{null, "/usr/bin/oc"};
```
scabanovich: Is "/usr/bin/oc" matched to OS_WIN32 correctly?

robstryker: No.  There is no default location for windows oc tools. In windows, it's just a zip that the user expands wherever they want. For this reason, we have a null default location, which means the path will be searched instead, and, if not found, a 'where oc.exe' will be executed.

PR #835, opened 2015-12-03
[JBIDE-20530] Add docker image lookup to deploy image wizard

fbricon:


adietish:
looks good, +1, LGTM

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
```diff
@@ -102,8 +103,9 @@ public boolean originatedFromDockerExplorer() {
 
 	@Override
 	public IDockerConnection getDockerConnection() {
-		if(dockerConnection == null && getDockerConnections().size() == 1) {
-			setDockerConnection(getDockerConnections().get(0));
+		List<IDockerConnection> all = getDockerConnections();
+		if(dockerConnection == null && all.size() == 1) {
+			setDockerConnection(all.get(0));
 		}
```
scabanovich: Previous way was better in not calling getDockerConnections() when dockerConnection == null.
Taking the best from both,

if(dockerConnection == null) {
    List<IDockerConnection> all = getDockerConnections();
    if(all.size() == 1) {
        setDockerConnection(all.get(0));
    }
}

fbricon:
@scabanovich agreed. Will fix that in a next commit.

PR #838, opened 2015-12-09
[JBIDE-21187] making sure we required tree ownerdraw fixedtree ownerd…

adietish:
…raw fixed org.eclipse.swt 3.104.1


maxandersen:
this PR is incomplete now, right?

adietish:
Yes, incomplete. It's currently only removing the workaround which makes it functional on Mars.1 only. It's missing the logical switch which would use the workaround in Mars.0 while not using it in Mars.1

fbricon:
is this still relevant for JBT 4.4?

adietish:
outdated, closing

PR #839, opened 2015-12-09
Bump DMR lib to 1.3 for openshift-restclient-java updates

jcantrill:


jcantrill:
cc @fbricon

fbricon:
+1 to apply

maxandersen:
what is the related jira for this ? interested in understanding why openshift need to update to dmr 1.3 and how about the other places we have dmr - do they need update too?

jcantrill:
There was not a jira for this. Bump was needed to assist in sanitizing Json
to remove nulls.   Previous version did not have a way to remove entries
from a list. Other places will not need this change unless they have need
to remove from a list. Also, AFAIK, dmr is not part of the TP unless it
gets pulled in with some other dependency.

On Friday, December 11, 2015, Max Rydahl Andersen <notifications@github.com>
wrote:

> what is the related jira for this ? interested in understanding why
> openshift need to update to dmr 1.3 and how about the other places we have
> dmr - do they need update too?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jbosstools/jbosstools-openshift/pull/839#issuecomment-163877013>
> .
>


-- 
--
Jeff Cantrill
Senior Software Engineer, Red Hat Engineering
Red Hat, Inc.
*Office*: 703-748-4420 | 866-546-8970 ext. 8162420
jcantril@redhat.com
http://www.redhat.com

maxandersen:
we've been talking about moving dmr to TP so it would be best we don't diverge too much, but if we do at least have a jira documenting what we started/needed to use. i.e. I assume this fix is related to some specific fix in DMR that would be useful to know.

PR #841, opened 2015-12-10
[JBIDE-21133] implement better docker image name validation

fbricon:

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/DockerImageValidator.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/DockerImageValidator.java
@@ -1,0 +1,100 @@
+/**
+ * Copyright 2013-2015 Docker, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ * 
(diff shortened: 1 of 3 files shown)


jcantrill:
LGTM

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/DockerImageValidator.java
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	public IStatus validateImageName(String repoName) {
+		String uri = repoName;
+		if (StringUtils.isEmpty(uri) || uri.contains("://")) {
+			return ValidationStatus.cancel(ERROR_MSG);
+		}
+		Matcher matcher = IMAGE_PATTERN.matcher(repoName);
+		if (!matcher.matches()) {
+			return ValidationStatus.cancel(ERROR_MSG);
```
scabanovich: It always seems user unfriendly when a validation for a possibly long string with a lot of rules to it just says that there is an error in it. The Matcher does not provide API for looking into the problem it met. Can we try being smarty and if the entire pattern did not match, start here 'prefix matchers', for example 
String hostOnlyRegexp = "^((?:" + hostnameRegexp + "/)?";
String hostAndNameOnlyRegexp = "^((?:" + hostnameRegexp + "/)?" + nameRegexp + ")";
String hostAndNameAndTagOnlyRegexp = "^((?:" + hostnameRegexp + "/)?" + nameRegexp + ")(?:[:](" + tagRegexp +"))?
etc.
to get to the problem more close and return a finer report for the problem?

PR #842, opened 2015-12-10
[JBIDE-20764] implemented server editor for v3 adapter

adietish:


adietish:
@fbricon plz comment

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
```diff
@@ -35,196 +57,445 @@
...
+									model.setDeployProject(deployProject);
+								}
+								if (connection != null) {
+									model.setConnection(connection);
+								}
+								return Status.OK_STATUS;
+							}
+						})
+				.runWhenDone(new DisableAllWidgetsJob(true, container, false, busyCursor))
+				.schedule();
```
scabanovich: Should not the first parameter here be 'false'?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServiceViewModel.java
```diff
@@ -0,0 +1,198 @@
...
+	private void updateConnections(List<Connection> connections) {
+		List<Connection> oldItems = new ArrayList<>(this.connections);
+		// ensure we're not operating on the same list
+		List<Connection> newItems = new ArrayList<>();
+		if (connections != null) {
+			newItems.addAll(connections);
+		}
+		this.connections.clear();
+		this.connections.addAll(newItems);
+		firePropertyChange(PROPERTY_CONNECTIONS, oldItems, this.connections);
```
scabanovich: If concern is that 'this.connections' and 'connections' may be the same object, code still may be more concise:

List<Connection> oldItems = new ArrayList<>(this.connections);
if(connections != this.connections) {
    this.connections.clear();
    if(connections !=null) {
        this.connections.addAll(connections);
    }
}

Besides, if 'connections' and 'this.connections' is the same object then there is no need to fire so that check if(connections != this.connections) can wrap the entire method content, not just copying list.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServiceViewModel.java
```diff
@@ -0,0 +1,198 @@
...
+		List<ObservableTreeItem> oldItems = new ArrayList<>(this.serviceItems);
+		// ensure we're not operating on the same list
+		List<ObservableTreeItem> newItems = new ArrayList<>();
+		if (serviceItems != null) {
+			newItems.addAll(serviceItems);
+		}
+		this.serviceItems.clear();
+		this.serviceItems.addAll(newItems);
+		firePropertyChange(PROPERTY_SERVICE_ITEMS, oldItems, this.serviceItems);
+	}
```
scabanovich: See comment to the previous method.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
fbricon: indeed. good catch

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
```diff
@@ -35,196 +57,445 @@
...
+		ValueBindingBuilder
+				.bind(WidgetProperties.text(SWT.Modify).observe(sourcePathText))
+				.validatingAfterConvert(new IValidator() {
+
+					@Override
+					public IStatus validate(Object value) {
+						if (StringUtils.isEmpty(value)) {
+							return ValidationStatus.cancel("Please provide a source path to deploy to the pod.");
+						}
+						String path = (String) value;
```
scabanovich: If value is not instance of String (e.g. null), StringUtils.isEmpty(Object) will return false which is not what is expected.
Even if we are 100% sure that non-null value cannot be other than String, this cast is unsafe as a pattern.

fbricon:
I applied the PR in 4.3.x and master, after addressing @scabanovich 's remarks and fixing a couple other issues : 
- the server page had no title/description/image
- the initial pod-to-local-dir rsync was never executed, which still caused some permission issues during rsync

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
adietish: it should be true if we editing was implemented (editing was not implemented since it's non-obvious for a server editor to turn dirty - and allow saving - when editing a project and NOT the server). I therefore turned this to false on purpose to disable editing for now. But of course, once editing is supported the "true" will become a "false" so that the widgets may be used to edit values.

adietish:
thx!

PR #845, opened 2015-12-11
[JBIDE-21295] implemented #getWelcomePageUrl in server adapter

adietish:


adietish:
reviewed and tested by @fbricon, +1ed

PR #847, opened 2015-12-14
JBIDE-21313 : fix open in web console

fbricon:


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/OpenInWebConsoleHandler.java
```diff
@@ -35,7 +37,7 @@
 /**
  * @author Fred Bricon
  */
-public class OpenInWebConsoleHandler extends AbstractHandler {
+public class OpenInWebConsoleHandler extends OpenInWebBrowserHandler {
```
adietish: Looking at the reason for inheritance here I see it reusing a dialog for the case that the url is empty and launching the browser. Both are 1-liners so I personnally would prefer composition over inheritance, but that's more a matter of gusto.
Nevertheless the warning message in the superclass seems incorrect, it'll tell the user "No route to open"

adietish: I suggest the following #849

fbricon:
superseded by #849

PR #848, opened 2015-12-14
[JBIDE-19178] Simple deployment view

jcantrill:
This PR provides for a simple deployment view
* Tabbed property view
* Display of service(route)->pods

cc @fbricon @adietish


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/Deployment.java
```diff
@@ -0,0 +1,58 @@
...
+ * A deployment is the collection of resources
+ * that makes up an 'application'
+ * 
+ * @author jeff.cantrill
+ *
+ */
+public class Deployment {
+
+	private IService service;
+	private List<IRoute> routes = new ArrayList<>();
```
fbricon: no need to initialize here. it's done in the constructor.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/Deployment.java
```diff
@@ -0,0 +1,58 @@
...
+	private List<IPod> pods;
+	
+	public Deployment(IService service, List<IRoute> routes, List<IPod> pods) {
```
fbricon: prolly need assert routes/pods are not null here, or at least assigne empty collections if they are

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/Deployment.java
```diff
@@ -0,0 +1,58 @@
...
+ * that makes up an 'application'
+ * 
+ * @author jeff.cantrill
+ *
+ */
+public class Deployment {
+
+	private IService service;
+	private List<IRoute> routes = new ArrayList<>();
+	private List<IPod> pods;
```
fbricon: maybe use Sets instead of lists? so that you avoid potential duplicates when doing add(pod)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/Deployment.java
fbricon: Maybe use Collection parameters, and create new Sets in constructor

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/DeploymentResourceMapper.java
```diff
@@ -0,0 +1,60 @@
...
+			adapters.add(new Deployment(service, appRoutes, pods));
+		}
+		
+		return adapters;
+	}
+	
+	public static Deployment getDeployment(IService service) {
+		List<IRoute> routes = ConnectionsRegistryUtil.getConnectionFor(service).getResources(ResourceKind.ROUTE, service.getNamespace());
+		List<IPod> pods = service.getPods();
+		List<IRoute> appRoutes = routes.stream()
```
fbricon: can probably refactor the filtering part, also used in getDeployments -> filter(routes, serviceName)

jcantrill:
@fbricon Updated PR based on your comments. Removed class that is no longer in use.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/DeploymentResourceMapper.java
```diff
@@ -0,0 +1,60 @@
...
+	
+	public List<Deployment> getDeployments(){
+		List<IService> services = conn.getResources(ResourceKind.SERVICE, project.getName());
+		Collection<IRoute> routes = conn.getResources(ResourceKind.ROUTE, project.getName());
+		
+		List<Deployment> adapters = new ArrayList<>(services.size());
+		for (IService service : services) {
+			adapters.add(getDeployment(service, routes));
+		}
+
```
scabanovich: Can it be done as
List<Deployment> adapters = services.stream().map(service -> getDeployment(service, routes)).collect(Collectors.toList());

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -27,47 +26,38 @@
...
-		ResourceKind.POD,
-		ResourceKind.REPLICATION_CONTROLLER, 
-		ResourceKind.BUILD, 
-		ResourceKind.IMAGE_STREAM, 
-		ResourceKind.ROUTE
-	};
-	
-	private Map<IProject, List<ResourceGrouping>> groupMap = new HashMap<IProject, List<ResourceGrouping>>();
+	private Map<IProject, List<Deployment>> deploymentsByProject = new HashMap<>();
+	private Map<IService, Deployment> deploymentByService = new HashMap<>();
```
scabanovich: Is it expected that two distinct equal instances of IService (service1.equals(service2) == true, service1 != service2) will make one entry in deploymentByService? Is it expected that if deploymentByService has key service1, then deployment can be retrieved with equal service2? Both expectations fail with HashMap. However, if model always provides the same instance for the same service, it is not a problem.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
```diff
@@ -109,14 +113,25 @@ public StyledString getStyledText(Object element) {
...
+		if (element instanceof Deployment) {
+			Deployment d = (Deployment) element;
+			return style(d.getService().getName(), formatRoute(d.getRoutes()));
+		}
 		return super.getStyledText(element);
 	}
 
+	private String formatRoute(Collection<IRoute> routes) {
+		if(routes.size() > 0) {
+			IRoute route = (IRoute)routes.toArray()[0];
```
scabanovich: IRoute route = routes.iterator().next();

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -75,6 +65,58 @@ protected void handleConnectionChanged(IConnection connection, String property,
...
+			IService service = (IService)resource;
+			deploymentByService.remove(service);
+			removeChildrenFromViewer(resource.getProject(), deploymentByService.get(service));
+			break;
+		case ResourceKind.POD:
+			IPod pod = (IPod)resource;
+			for (IService deployedService : deploymentByService.keySet()) {
+				if(selectorsOverlap(deployedService.getSelector(), pod.getLabels())){
+					deploymentByService.get(deployedService).remove(pod);
+					removeChildrenFromViewer(deploymentByService, pod);
```
scabanovich: Should it be 
removeChildrenFromViewer(deployedService, pod);
?

jcantrill: I would ignore this 'refresh' logic for now since its broken in this pull.  It needs to be investigated more as a follow on task.  Your comments may be valid

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
adietish: I would think that #toArray() which is System.arraycopy'ing behind the scenes is far more efficient than iterator creation. 
Beside this I think that micro-optimizations on this level rarely matter. But maybe code-clarity is a better with the iterator. Matter of gusto, dont have a strong opinion on this tbh

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/DeploymentResourceMapper.java
jcantrill: I think it is unlikely that either the number of services or routes in a given project will be large.  The only alternative I can see is to iterate over routes and create a map of service->routes and look up by service name.  I wonder if its worth the optimization at this time...

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
jcantrill: We will have cases where 2 distinct object instances represent the same service, possibly in different states.  Hopefully, we have a hashing algo that will identify they are the same.

We should address the NP, but name and kind SHOULD NEVER be null since it is the fundamental THING that distinguishes the resource.

jcantrill: We should consider fixing the NP scenario, but knowing the impl, it will always return empty string or a value and never null.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
scabanovich: Yes, I agree that this is a micro-optimization which in this concrete case where 'routes' list may not be be large, does not deserve much attention.
But, it still may be a good occasion to assert that list.iterator().next() is _always_ faster than list.toArray()[0], even for the size 1. And if it comes to a habit to convert to list when one element is needed, the habit may play a bad joke in a performance bottleneck case, as in some cases times do not just add but multiply.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
scabanovich: Yes, this is a trivial issue, but just the same, hashCode() and equals() are very related methods, and if one of them checks or does not check some value for null for some good reason, the other should treat that value in the same way.

jcantrill:
Merged as part of feature branch: feature_19178_deployment_view

jcantrill:
Closing since merged into feature branch

PR #851, opened 2015-12-15
JBIDE-21327 - most recent openshift-vagrant requires tls-verify flag …

robstryker:
…to be set properly


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
```diff
@@ -109,12 +109,10 @@ private void configureDocker(IServer server, ADBInfo adb) {
 						.name(getNextName(server, mgr)).tcpHost(dockerHost);
 				String tlsVerifyString = adb.env.get("DOCKER_TLS_VERIFY");
 				boolean tlsVerify = (Integer.parseInt(tlsVerifyString) != 0);
```
maxandersen: afaik just the presence of DOCKER_TLS_VERIFY is enough - the value does not matter.

maxandersen:
looks safe and valid to apply

maxandersen:
+1 for branch and master.

robstryker:
pushed after rebase

PR #852, opened 2015-12-15
JBIDE-21170 - fix martin's issue regarding launching eclipse with .app having no path at all

robstryker:


robstryker:
PR seems for wrong branch. give me a minute to fix.

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationLookupStrategy.java
```diff
@@ -114,42 +144,52 @@ private String searchPath(String path, String delim, String commandName) {
...
+				p = Runtime.getRuntime().exec(cmd[0]);
 			try {
 				p.waitFor();
 			} catch(InterruptedException ie) {
 				// Ignore, expected
 			}
+			InputStream is = null;
 			if(p.exitValue() == 0) {
-				InputStream is = p.getInputStream();
+				is = p.getInputStream();
```
fbricon: shouldn't that stream be closed at the end?

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationLookupStrategy.java
```diff
@@ -82,11 +91,32 @@ private  String findLocation(String defaultLoc, String cmd, String which, String
...
+	
+	public String ensureFolderOnPath(String existingPath, String folder) {
+		existingPath = (existingPath == null ? "" : existingPath);
+		String[] roots = null;
+		if( existingPath.isEmpty() && pathCommand != null) {
+			// Path should never be empty, we have to discover the path (OSX when eclipse launched via .app)
+			String pathresult = runCommand(pathCommand, false);
+			roots = pathresult.split(delim);
+		} else
+			roots = existingPath.split(delim);
```
scabanovich: With empty string, "".split(delim) returns array of length 1 with empty string as the element.
Then, adding to the list a folder "myFolder" and joining will result in ",myFolder" - with leading comma.
Maybe it is not important in this case, but please check it.

scabanovich: And what is funny about split(String regex) is that
"".split(delim).length == 1
delim.split(delim).length == 0
(delim + delim).split(delim).length == 0
("myFolder" + delim).split(delim).length == 1
(delim + "myFolder").split(delim).length == 2

Very strange arithmetic...

robstryker: we already check for isEmpty() 5 lines above

robstryker: Gave a further look and worst case here is that our new path starts with delim,  which isn't an issue for any OS to be able to handle and is still valid.

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationLookupStrategy.java
robstryker: Good catch fred

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/internal/common/core/util/CommandLocationLookupStrategy.java
scabanovich: Ok, good.
And well, Javadoc to split method describes its logic in details, it covers all cases that I mentioned above, and that is something worth to remember when it may impact results.

robstryker:
PR targeted to wrong branch since it's old.

PR #853, opened 2015-12-15
[JBIDE-21318] use proper OpenShift version in server labels

fbricon:


adietish:
+1 for the quick fix for now.

scabanovich:
May I mention here a potential NPE in Connection.equals(). Comparing for clients goes  
		if (client == null) {
			if (other.client != null)
				return false;
		} else if (!client.getBaseURL().toString().equals(other.client.getBaseURL().toString()))
			return false;
When this.client is not null, other.client is not checked for null though checks above imply that it is possible.

adietish:
+1, goo catch, thanks!

adietish:
PR'ed in #856

PR #854, opened 2015-12-15
[JBIDE-21325] allow to disable password prompt when in conn dialog

adietish:


adietish:
@fbricon plz review

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
```diff
@@ -380,6 +382,9 @@ public void setNotConnected() {
 	
 	private void setConnectedStatus(IStatus status) {
 		firePropertyChange(PROPERTY_CONNECTED_STATUS, this.connectedStatus, this.connectedStatus = status);
+		if (isConnected()) {
+			this.connection.enablePromptCredentials(true);
```
fbricon: can the connection be null here?

fbricon: probably not if isConnected returns true

fbricon:
tested it works.
+1

adietish:
great, thx!

adietish:
merged and pushed to upstream/jbosstools-4.3.1.Beta1x

PR #856, opened 2015-12-16
[JBIDE-21318] fixed possible NPE in Connection#equals (as pointed out…

adietish:
… by VKabanovich)


adietish:
merged and pushed to jbosstools-4.3.1.Beta1x
cherry-picked and pushed to jbosstools-4.3.x, master

PR #857, opened 2015-12-16
JBIDE-21335 - when server has no credential set (or missing credentia…

robstryker:
…l), server gets stuck in starting... mode

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
@@ -154,7 +154,14 @@ public class CDKLaunchController extends AbstractSubsystemController implements 
 		}
 		
 		String args = configuration.getAttribute(ATTR_ARGS, (String)null);
-		ILaunchConfigurationWorkingCopy wc = new VagrantLaunchUtility().createExternalToolsLaunchConfig(s, args, getStartupLaunchName(s));
+		ILaunchConfigurationWorkingCopy wc = null;
+		try {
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
```diff
@@ -154,7 +154,14 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
 		}
 		
 		String args = configuration.getAttribute(ATTR_ARGS, (String)null);
-		ILaunchConfigurationWorkingCopy wc = new VagrantLaunchUtility().createExternalToolsLaunchConfig(s, args, getStartupLaunchName(s));
+		ILaunchConfigurationWorkingCopy wc = null;
+		try {
+			wc = new VagrantLaunchUtility().createExternalToolsLaunchConfig(s, args, getStartupLaunchName(s));
+		} catch(CoreException ce) {
+			CDKCoreActivator.pluginLog().logError(ce);
```
maxandersen: okey so this fix catch an error and set server stopped  if error happens.

I assume this is just a temporary fix since having credentials is not a full requirement is it ?

fbricon:
+1

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
robstryker: For CDK server? Currently having credentials is a hard requirement. Especially in the openshift-vagrant repository. But yes, it's likely this will be re-visited when the raw cdk is supported.

PR #860, opened 2015-12-16
[JBIDE-21340] existing proj validation: removed dirty check & git rem…

adietish:
…ote HTTPS is now warning


adietish:
we're missing the git source property validation in this PR

fbricon:
PR applied in master and 4.3.x. We can further enhance the validation in a subsequent commit

PR #864, opened 2015-12-18
[JBIDE-21352] server adapter wiz: using the selected project in the visible project view

adietish:


plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
```diff
@@ -289,18 +295,63 @@ public static boolean isSingleSelection(ISelection selection) {
...
+	/**
+	 * Returns the visible workbench parts which match the given ids. If no ids
+	 * are given all visible parts are returned.
+	 * 
+	 * @param partIds
+	 * @return
+	 * 
+	 * @see IWorkbenchPart
+	 */
+	public static List<IViewPart> getVisibleViewParts(String... partIds) {
```
fbricon: if partIds == null or length==0 return empty list

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
```diff
@@ -289,18 +295,63 @@ public static boolean isSingleSelection(ISelection selection) {
...
+		IWorkbenchWindow workbenchWindow = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
+		IWorkbenchPage page = workbenchWindow.getActivePage();
+		List<IViewPart> parts = new ArrayList<>();
```
fbricon: List<IViewPart> parts = new ArrayList<>(partIds.length); //list size will be at most == to params length

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
```diff
@@ -289,18 +295,63 @@ public static boolean isSingleSelection(ISelection selection) {
...
+		List<String> partIdsList = null;
```
fbricon: List<String> partIdsList = Arrays.asList(partIds); //given we should check partIds is not null at the beginning

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
```diff
@@ -289,18 +295,63 @@ public static boolean isSingleSelection(ISelection selection) {
...
+	 * <ul>
+	 * <li>package explorer</li>
+	 * <li>project explorer</li>
+	 * <li>resource navigator</li>
+	 * </ul>
+	 * @return
+	 */
+	public static IViewPart getVisibleProjectsView() {
+		List<IViewPart> parts = getVisibleViewParts(PACKAGE_EXPLORER_ID, PROJECT_EXPLORER_ID, RESOURCE_NAVIGATOR_ID);
+		if (parts.isEmpty()) {
```
scabanovich: Ah, sorry, I am with a trifle.
List<IViewPart> getVisibleViewParts(String... partIds)
is good and may be even put to common.ui to be of use to all JBoss Tools.
But here, since we need only the first one visible of the listed parts, I would suggest one more method
IViewPart getFirstVisibleViewPart(String... partIds)
which would not build a list but return the first found visible part.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
scabanovich: Comments to the method declare 'If no ids are given all visible parts are returned' - and that is implemented.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
```diff
@@ -289,18 +295,63 @@ public static boolean isSingleSelection(ISelection selection) {
...
+			return null;
+		}
+		return parts.get(0);
```
scabanovich: Returned list does not keep the precedence set by the order of ids. Several views can be visible at the same time in different regions of window. Method getVisibleViewParts should sort the built list to match the order of ids.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
adietish: +1

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
adietish: +1 slightly more performant since no array resizing

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
adietish:
```
		if (partIds != null) {
			partIdsList = Arrays.asList(partIds);
			parts = new ArrayList<>(partIds.length);
		} else {
			parts = new ArrayList<>();
		}
```
since calling getVisibleViewParts() would result in a null varargs array and is perfectly valid.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
adietish: +1

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
adietish: +1

PR #874, opened 2016-01-04
JBIDE-21362 - process.waitFor needs a timeout when running adbinfo

robstryker:


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +126,104 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
+		msg.append("\n\nProcess Error:\n");
+		lines = err;
+		for( int i = 0; i < lines.length; i++ ) {
+			msg.append("    ");
+			msg.append(lines[i]);
+		}
+		return msg.toString();
+	}
 	
+	private static class StreamGobbler implements Runnable {
```
fbricon: should implement AutoCloseable

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +126,104 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
+		public synchronized boolean isCanceled() {
+			return canceled;
+		}
+		public synchronized void cancel() {
+			canceled = true;
+		}
+		public String[] getLines() {
+			return (String[]) lines.toArray(new String[lines.size()]);
+		}
+		public void cleanup() {
```
fbricon: rename as close() to match AutoCloseable

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +126,104 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
-		BufferedReader buff = new BufferedReader(
-				new InputStreamReader(p.getInputStream()));
+		final Process p = Runtime.getRuntime().exec(cmd.toArray(new String[0]), envp, vagrantDir);
+
+		InputStream errStream = p.getErrorStream();
+		InputStream inStream = p.getInputStream();
+
+		StreamGobbler inGobbler = new StreamGobbler(inStream);
+		StreamGobbler errGobbler = new StreamGobbler(errStream);
+
```
fbricon: could instanciate in latter try(...), if  StreamGobbler is AutoCloseable

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +126,104 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
+				// Timeout reached
+				inGobbler.cancel();
+				errGobbler.cancel();
+				throw new TimeoutException(getTimeoutError(inGobbler.getLines(), errGobbler.getLines()));
 			}
-		} catch(InterruptedException ie) {
-			// ignore
+			
+			return inGobbler.getLines();
+		} finally {
```
fbricon: unnecessary finally block if StreamGobbler is AutoCloseable

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +126,104 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
+			inGobbler.cleanup();
+			errGobbler.cleanup();
 		}
-		return result.toArray(new String[0]);
 	}
 	
+	private static String getTimeoutError(String[] output, String[] err) {
+		StringBuffer msg = new StringBuffer();
```
fbricon: StringBuilder

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +126,104 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
+				lines.add(line);
+			}
+		}
+		public synchronized boolean isCanceled() {
+			return canceled;
+		}
+		public synchronized void cancel() {
+			canceled = true;
+		}
+		public String[] getLines() {
```
fbricon: Why an array? List<String> is just as fine

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +126,104 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
+			inGobbler.cleanup();
+			errGobbler.cleanup();
 		}
-		return result.toArray(new String[0]);
 	}
 	
+	private static String getTimeoutError(String[] output, String[] err) {
+		StringBuffer msg = new StringBuffer();
+		msg.append("Process output:\n");
+		String[] lines = output;
```
fbricon: if output was a list, then 
```java
lines.forEach(line -> msg.append("   ").append(line));
```

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +126,104 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
+		return msg.toString();
+	}
 	
+	private static class StreamGobbler implements Runnable {
+		private Scanner inScanner;
+		private ArrayList<String> lines;
+		private boolean canceled = false;
+		private BufferedInputStream is;
+		public StreamGobbler(InputStream inStream) {
+			lines = new ArrayList<String>();
```
fbricon: new ArrayList<>();

scabanovich:
Method VagrantLaunchUtility.convertEnvironment(Map<String, String> env) can be enhanced:

	public static String[] convertEnvironment(Map<String, String> env) {
		if (env == null || env.size() == 0)
			return null;

		// Create a new map based on pre-existing environment of Eclipse
		Map<String, String> original = new HashMap<>(System.getenv());
		// Add new environment on top of existing
		original.putAll(env);		
		// Convert the combined map into a form that can be used to launch
		// process
		return original.entrySet().stream().map(e->e.getKey() + "=" + e.getValue()).collect(Collectors.toList()).toArray(new String[original.size()]);
	}

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +126,104 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
+			is = new BufferedInputStream(inStream);
+			inScanner = new Scanner(is);
+		}
+
+		@Override
+		public void run() {
+			while (inScanner.hasNextLine() && !isCanceled()) {
+				String line = inScanner.nextLine();
+				lines.add(line);
```
scabanovich: When cancel() is called, this thread can be right before adding line to lines, and that may cause concurrent exception while the other thread calls getLines() or works with list if getLines() will return list. I suggest wrap it with

		synchronized (this) {
			if(!isCanceled()) {
				lines.add(line);
			}
		}

and call cancel() as the first command in getLines().

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
adietish: array makes sure that consumer is not operating on the same list. Can be achieved via array or Collections.unmodifiableList or ...

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
adietish:
```
if (!isCanceled()) {
     synchronized (lines) {
           lines.add(line);
     }
}
```

and the corresponding lock when getting lines.

or simply use a concurrent collection: CopyOnWriteArrayList, ConcurrentLinkedQueue etc.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
```diff
@@ -221,14 +217,13 @@ private ADBInfo loadADBInfo(IServer server) {
...
-			ce.printStackTrace(); // TODO log
+		} catch(IOException | TimeoutException ce) {
+			CDKCoreActivator.pluginLog().logError("Unable to successfully complete a call to vagrant adbinfo. ", ce);
 		}
 		return null;
 	}
 	
 	private static class ADBInfo {
-		// Mocked
 		private int openshiftPort = 8443;
```
adietish: static final?

adietish: static final?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
```diff
@@ -221,14 +217,13 @@ private ADBInfo loadADBInfo(IServer server) {
...
 		private String openshiftHost = "https://10.1.2.2";
```
adietish: static final?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
```diff
@@ -240,7 +235,6 @@ public ADBInfo(HashMap<String,String> env) {
 				URI url = new URI(dockerHost);
 				String h = url.getHost();
 				openshiftHost = "https://" + h;
-				//openshiftPort = url.getPort();  // i don't believe this is the correct port. This is for docker
 			} catch(URISyntaxException murle) {
```
adietish: Why should this be catched? Shouldn't this make it to the caller that then can react accordingly?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
robstryker: What are the benefits to making the inner static class final? I don't imagine there are many here. Since it's private, no subclass can override it anyway.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
scabanovich: Or return new ArrayList(lines)

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -120,32 +119,93 @@ private ILaunchConfigurationWorkingCopy findLaunchConfig(IServer s, String launc
...
+		return msg.toString();
+	}
 	
+	private static class StreamGobbler implements Runnable, AutoCloseable {
+		private Scanner inScanner;
+		private CopyOnWriteArrayList<String> lines;
+		private boolean canceled = false;
+		private BufferedInputStream is;
+		public StreamGobbler(InputStream inStream) {
+			lines = new CopyOnWriteArrayList<>();
```
scabanovich: CopyOnWriteArrayList makes complete copy of its array when adding each element. If the number of lines may be more than 100, it will impact performance.
If getLines() be started with cancel() before return lines, then no item will be added to the list after getLines() is called.

robstryker: I'd rather not design the class such that calling getLines() cancels further reading.

scabanovich: Then getLines() may be synchronized and return copy of List - it will be only one copy instead of copy per adding an item.

PR #876, opened 2016-01-05
[JBIDE-21399] keep the editor dirty if resource save/update failed

fbricon:


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/OpenShiftResourceDocumentProvider.java
```diff
@@ -85,8 +85,9 @@ public IClient visit(IClientCapability cap) {
 			@Override
 			protected IStatus run(IProgressMonitor monitor) {
 				try {
-					client.update(newResource);//should run in job as well
+					client.update(newResource);
 				} catch (OpenShiftException e) {
+					setDirty(element);
```
scabanovich: As it notifies ui, probably setDirty should be called with Display.getDefault().asyncExec().
In this case, it is guaranteed to be called after AbstractDocumentProvider.SaveOperation sets editor state saved and setting it dirty will not be overrun.

PR #879, opened 2016-01-06
[JBIDE-21413] Update for changes to Service#getTargetPort returning string

jcantrill:


jcantrill:
@fbricon @adietish  was planning to just merge this with the change to the restclient so it doesnt hold up anyone.  should only be changes to make things work correctly

PR #884, opened 2016-01-07
[JBIDE-21415] Update deployment mapper to properly populate tab content

jcantrill:


jcantrill:
cc @adietish @fbricon

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
```diff
@@ -64,4 +65,24 @@ private static boolean inCollection(String item, final Collection<String> texts)
...
+	/**
+	 * Determine if the source map overlaps the target map (i.e. Matching a service to a pod). There
+	 * is  match iff the target includes all the the keys from the source and those keys have
+	 * matching values
+	 * 
+	 * @param source
+	 * @param target
+	 * @return true if there is overlap; false; otherwise
+	 */
+	public static boolean selectorsOverlap(Map<String, String> source, Map<String, String> target) {
```
fbricon: this is a public method so better perform some null checks, or at least use assert statements

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
```diff
@@ -64,4 +65,24 @@ private static boolean inCollection(String item, final Collection<String> texts)
...
+		if(!target.keySet().containsAll(source.keySet())) {
+			return false;
+		}
+		for (String key : source.keySet()) {
+			if(!target.get(key).equals(source.get(key))) {
```
fbricon: Use Objects.deepEquals(target.get(key), source.get(key)), it's null safe

PR #889, opened 2016-01-08
[JBIDE-21428] look up oc binary in preferences and fall back to $PATH

adietish:


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
```diff
@@ -35,20 +48,30 @@ public String getName() {
...
+	 * Returns the location from preferences or looks it up on the path.
+	 * 
+	 * @return
+	 */
+	public String getLocation() {
+		String location = OpenShiftCorePreferences.INSTANCE.getOCBinaryLocation();
+		if (StringUtils.isBlank(location)) {
+			location = findLocation();
 		}
+		return location;
```
scabanovich: Then OpenShiftPreferencePage and OpenShiftCorePreferenceInitializer cannot use that method, they need 'pure' search for default in $PATH.

adietish: I did that, the OpenShiftCorePreferenceInitializer now calls 
```
		String location = OCBinary.getInstance().findLocation();
```
right?

adietish:
I now also included your changes to OpenShiftPreferencePage.
@scabanovich plz review

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OpenShiftCorePreferenceInitializer.java
```diff
@@ -22,9 +23,9 @@
 	@Override
 	public void initializeDefaultPreferences() {
 		IEclipsePreferences defaultPreferences = ((IScopeContext)DefaultScope.INSTANCE).getNode(OpenShiftCoreActivator.PLUGIN_ID);
-		
-		String location = OCBinaryName.getInstance().getLocation();
```
adietish: as you suggested, we're now using #findLocation (which only looks up in $PATH)

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
scabanovich: Oh, yes. I would vote for getDefaultLocation(), name findLocation() remained obscure to me. But it is ok, as is done.

scabanovich:
It looks good to me.

adietish:
ok, thanks!

robstryker:
What was the logic of only searching the $PATH directly? Sometimes (such as on mac) the binary may be placed in a folder that's not on the default path, or that's only on the path when using a command line shell but not eclipse's path. 

Eclipse gets a path when it is run. If the user then goes and changes the path (After discovering an error for example), do we really want him to reboot his workspace? Or do we just want to be able to find it seamlessly by running a bash command in the background to see if it's on the path now? 

The full logic of CommandLocationBinary will check the PATH passed to eclipse environment, but if it still can't find it, will run bash in the background. This allows the user to update the PATH via export PATH=etc on the command line without having to reboot eclipse.

PR #895, opened 2016-01-12
[JBIDE-21349] Browse for Workspace source folder

fbricon:
Workspace based paths now use the workspace_loc variable.

Opening the filesystem browser converts a workspace relative path to an absolute one
Opening the workspace browser converts an absolute relative path to a workspace relative one
If a filesystem path that doesn't match anything in the workspace is used to open the workspace browser, it falls back on selecting the selected project's path

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardFragment.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardFragment.java
@@ -25,9 +25,11 @@ import org.eclipse.core.databinding.property.list.IListProperty;
 import org.eclipse.core.databinding.property.list.MultiListProperty;
 import org.eclipse.core.databinding.validation.IValidator;
 import org.eclipse.core.databinding.validation.ValidationStatus;
+import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Path;
@@ -40,6 +42,7 @@ import org.eclipse.jface.databinding.viewers.ObservableListContentProvider;
 import org.eclipse.jface.databinding.viewers.ObservableListTreeContentProvider;
 import org.eclipse.jface.databinding.viewers.ViewerProperties;
 import org.eclipse.jface.dialogs.Dialog;
+import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.jface.layout.GridLayoutFactory;
 import org.eclipse.jface.viewers.ColumnLabelProvider;
@@ -48,7 +51,9 @@ import org.eclipse.jface.viewers.StructuredViewer;
 import org.eclipse.jface.viewers.StyledCellLabelProvider;
 import org.eclipse.jface.viewers.StyledString;
 import org.eclipse.jface.viewers.TreeViewer;
+import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.jface.viewers.ViewerCell;
+import org.eclipse.jface.viewers.ViewerFilter;
 import org.eclipse.jface.wizard.IWizardContainer;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.ModifyEvent;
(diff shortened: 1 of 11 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardFragment.java
```diff
@@ -356,22 +366,78 @@ public void widgetSelected(SelectionEvent e) {
...
+		});
+		dialog.setAllowMultiple( false );
+		org.eclipse.core.resources.IResource res = model.getDeployProject();
+		if (org.apache.commons.lang.StringUtils.isNotBlank(selectedFile)) {
+			String prefix = "${workspace_loc:";
+			String path = selectedFile;
+			if (selectedFile.startsWith(prefix) && selectedFile.endsWith("}")) {
+				path = path.substring(prefix.length(), path.length()-1);
+			}
+			res = ResourcesPlugin.getWorkspace().getRoot().findMember(path);
```
scabanovich: Maybe, if the member is not found, 'res' is better to remain model.getDeployProject()?

fbricon: fixed

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardFragment.java
```diff
@@ -356,22 +366,78 @@ public void widgetSelected(SelectionEvent e) {
...
+	}
+
+	private ElementTreeSelectionDialog createWorkspaceFolderDialog(Shell shell, String selectedFile) {
+		ElementTreeSelectionDialog dialog = new ElementTreeSelectionDialog(
+				shell,
+				new WorkbenchLabelProvider(),
+				new WorkbenchContentProvider()
+		);
+		dialog.setTitle("Select an workspace folder");
+		dialog.setMessage("Select a workspace folder to deploy");
```
scabanovich: Either both 'a workspace', or both 'an workspace'. :)

fbricon: yeah I fixed that already

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardFragment.java
```diff
@@ -356,22 +366,81 @@ public void widgetSelected(SelectionEvent e) {
...
+
+			@Override
+			public boolean select(Viewer viewer, Object parentElement, Object element) {
+				return element instanceof IContainer;
+			}
+		});
+		dialog.setAllowMultiple( false );
+		org.eclipse.core.resources.IResource res = model.getDeployProject();
+		if (org.apache.commons.lang.StringUtils.isNotBlank(selectedFile)) {
+			String prefix = "${workspace_loc:";
```
scabanovich: "workspace_loc" is used 3 times. Maybe it deserves a constant. Then for prefix it may be "${" + WORKSPACE_LOC.
It is a pity that StringSubstitutionEngine.VARIABLE_START is a private constant.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardFragment.java
```diff
@@ -356,22 +366,81 @@ public void widgetSelected(SelectionEvent e) {
...
 	}
 
+	private String getWorkspaceRelativePath(String sourcePath) {
+		if (org.apache.commons.lang.StringUtils.isBlank(sourcePath) || sourcePath.contains("{")) {
+			return sourcePath;
+		}
+		IPath absolutePath = new Path(sourcePath);
+		IContainer container = ResourcesPlugin.getWorkspace().getRoot().getContainerForLocation(absolutePath);
+		if (container != null) {
+			return container.getFullPath().toPortableString();
```
scabanovich: I am not sure that toPortableString() is preferred here to just toString().
IPath.toString() makes only separator '/' for all platforms and it is what we need.
Docs say that if toPortableString() is used it has to be matched with fromPortableString() when converting back to Path.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardFragment.java
```diff
@@ -356,22 +366,81 @@ public void widgetSelected(SelectionEvent e) {
 		};
 	}
 
+	private String substitueVariables(String rawSourcePath) {
```
jcantrill: spelling: 'substitute'  or maybe 'replace'

scabanovich: @fbricon , sorry that I did not mention in my comment above about toPortableString() that I meant all its occurrences. There are 3 more of them, they all are applied to container full path implying the same arguments.

fbricon:
PR updated

scabanovich:
Looks good. 
I just got an idea, if VariablesHelper could have one more method getWorkspacePath(IResource) that would take full path of the resource and wrap it with the prefix.

fbricon:
we can add that when it's needed. I'll merge the PR so I can get rid of it now

fbricon:
Applied in master / 4.3.x

PR #900, opened 2016-01-14
[JBIDE-21429] Initial support for LiveReload

fbricon:
This initial support is pretty hacky. The LiveReload command requires IServerModules (that wraps an IServer and an IModule) to determine the host to proxy and the url to open
Current implementation of OpenShift server adapter doesn't contain any IModule, so we provide a stub IModule implementation, even though IModules should be created via IModuleFactories and never be implemented directly.


fbricon:
Applied in master / 4.3.x

PR #901, opened 2016-01-14
JBIDE-21371 - add custom auto-publish thread    

robstryker:
and ensure publishing for special project happens
   automatically

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/resources/OpenshiftResourceChangeListener.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/resources/OpenshiftResourceChangeListener.java
@@ -1,0 +1,254 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Red Hat Inc..
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat Incorporated - initial API and implementation
+ *******************************************************************************/
+package org.jboss.tools.openshift.internal.core.server.resources;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IResourceChangeEvent;
+import org.eclipse.core.resources.IResourceChangeListener;
+import org.eclipse.core.resources.IResourceDelta;
+import org.eclipse.core.resources.IResourceDeltaVisitor;
+import org.eclipse.core.resources.IResourceRuleFactory;
+import org.eclipse.core.resources.IncrementalProjectBuilder;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+import org.eclipse.core.runtime.jobs.Job;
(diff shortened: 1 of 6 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/resources/OpenshiftResourceChangeListener.java
```diff
@@ -0,0 +1,244 @@
+package org.jboss.tools.openshift.core.server.resources;
```
fbricon: missing copyright header

fbricon: should be internal package

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/NewApplicationHandler.java
```diff
@@ -32,6 +32,10 @@
 	public Object execute(ExecutionEvent event) throws ExecutionException {
 		try {
 			WizardUtils.openWizardDialog(new NewApplicationWizard(), HandlerUtil.getActiveShell(event));
+
+			// This technically sends the wrong event for the wrong 'command' id,
+			// but the refresh resource handler only looks at the selection... 
+			new RefreshResourceHandler().execute(event);
```
fbricon: how is that related to the autopublish issue?

robstryker: It's not, but I just got tired of the viewer not being updated when i exited the wizard. plz don't make me open new issue :(

adietish: This workaround will only refresh the selected resource, so this might help you when coding but I dont think that this is even a valid general purpose workaround. I would remove it.

fbricon: I believe the proper fix for refresh will be available from the new openshift explorer layout feature branch, so this hack should be removed

fbricon:
+1 to merge after removing new RefreshResourceHandler().execute(event); and squashing the commits

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/resources/OpenshiftResourceChangeListener.java
```diff
@@ -0,0 +1,254 @@
...
+	public class MagicProjectChangeJob extends Job {
+		private IResourceDelta delta;
+		private OpenShiftServer openshiftServer;
+		private IResourceChangeEvent event;
+		private List<IResource> changes;
+
+		public MagicProjectChangeJob(OpenShiftServer openshiftServer, IResourceDelta delta, List<IResource> changes) {
+			super(NLS.bind(Messages.jobUpdateServer, openshiftServer.getServer().getName()));
+			this.openshiftServer = openshiftServer;
+			this.delta = delta;
```
scabanovich: IResourceDelta passed to MagicProjectChangeJob is never used which is good because as docs warn the resource delta object 'is valid only for the duration of the invocation of' resourceChanged(IResourceChangeEvent), it may not be valid when job is runned. I suggest removing this parameter.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/resources/OpenshiftResourceChangeListener.java
```diff
@@ -0,0 +1,254 @@
...
+			((Server)server).setServerPublishState(newState);
+			
+			// TODO set the delta?
+			
+			
+			if (server.getServerState() != IServer.STATE_STOPPED && behaviourDelegate != null)
+				behaviourDelegate.handleResourceChange();
+			
+			if (server.getServerState() == IServer.STATE_STARTED)
+				autoPublish(event);
```
scabanovich: Object event was not passed to MagicProjectChangeJob.

PR #902, opened 2016-01-15
JBIDE-21454 - error in locating git repository for a given project

robstryker:


scabanovich:
There is also method EGitUtils.hasDotGitFolder(IProject), should not it use the same logic?

scabanovich:
This class in 3 places uses 3 different methods:
1) project.getLocation().toOSString() in hasDotGitFolder()
2) project.getLocation().toFile() in createRepository()
3) project.getLocationURI().getPath() in getRepositoryPathFor()
Imho, such a variety in choice cannot have a deep meaning.
The way (2) might do well for all 3 cases as File object can be passed as parent to new File(File, String)

robstryker:
I don't think it's smart to change all those other methods without knowing and understanding how each one is used and called. It's possible hasDotGitFolder() is meant only specifically to check whether that specific folder has a .git folder in it or if the folder is elsewhere. If that's the intent, then the logic is currently correct. 

I think changing only this one is the appropriate fix unless we come up with other usecases where the behavior is incorrect and the behavior matches the method name / expected behavior.

scabanovich:
You may be quite right in this case. I see that hasDotGitFolder() is used by ImportNewProject (OpenShift 2) in order to set up .gitignore, modify .settings, set markers for errors detected at import. Then, should this be done for the imported project if .git is not in its root but some folders above? At least, it is a question to ask. 
When a specific  problem is fixed and as a result similar pieces get different though they might remain similar,  it may be the very way for a large application to become too complex and buggy, too dangerous to touch anything that goes outside of the imminent problem.

plugins/org.jboss.tools.openshift.egit.core/src/org/jboss/tools/openshift/egit/core/EGitUtils.java
```diff
@@ -229,7 +229,11 @@ public static Repository createRepository(IProject project,
 	}
 
 	public static File getRepositoryPathFor(IProject project) {
-		return new File(project.getLocationURI().getPath(), Constants.DOT_GIT);
+		File f = new File(project.getLocationURI().getPath());
```
adietish: This is exactly what EGit RepositoryFinder#indInDirectoryAndParents already does in a more sophisticated manner:
https://github.com/eclipse/egit/blob/master/org.eclipse.egit.core/src/org/eclipse/egit/core/project/RepositoryFinder.java#L170

robstryker: Somehow the code you linked looks *way* more complicated... 

The real question is whether my patch is the way the method is intended to behave. If the two are really clones of each other, one should simply call the other, to keep logic isolated and convenience methods obvious.

fbricon:
Applied in master and 4.3.x

plugins/org.jboss.tools.openshift.egit.core/src/org/jboss/tools/openshift/egit/core/EGitUtils.java
adietish: It's not a matter of complexity but a matter of consistency. If EGit does a search for the repo the way it does then there are valid reasons for it. I dont see why we should do an alternative approach as EGit is the authoritative opinion here tbh.

PR #905, opened 2016-01-15
[JBIDE-21410] corrected source path validation and added decorations

adietish:


scabanovich:
@adietish , method OpenShiftServerEditorSection.createSourcePathControls also has similar constants, maybe they should be the same as in ServerSettingsWizardFragment.createSourcePathControls

PR #908, opened 2016-01-19
[JBIDE-21457] Show builds in explorer tree view

jcantrill:
cc @fbricon


fbricon:
+1 to apply after cleanup

PR #910, opened 2016-01-20
JBIDE-21481 - initial commit for wtp-style project support

robstryker:
This PR is an intiial implementation for zipping / assembling wtp-style projects for deployment / rsync over to the remote pod. This patch is the best we can do for beta2, but it does suffer from several problems, which require much larger architectural changes to be 100% functional. 

Discussion on the architectural issues should be on the jira, while this can review the code specifically.


robstryker:
This PR seems to encompass the commit from https://github.com/jbosstools/jbosstools-openshift/pull/902  because I couldn't get the usecase to work without that commit.

fbricon:
ok to apply in after squashing the last 4 commits

PR #926, opened 2016-01-27
[JBIDE-21587] moved common openshift databinding classes to jbt.common

adietish:


adietish:
wont happen for 4.3.x maintenance. keeping the PR in 4.4.x only

robstryker:
@adietish  can you link to the 4.4.x PR?   

It's also worth noting things really shouldn't be moved to jbt.common anymore.  Foundation is a much better place for api's that are to be consumed by a wider audience.

PR #929, opened 2016-01-28
JBIDE-21205 - allow cdk to use the new API in runtimes

robstryker:


fbricon:
OK to apply as long as upstream API is available too (dunno if it also requires applying another PR in base)

fbricon:
+1 to apply when upstream changes are available

PR #930, opened 2016-01-28
JBIDE-21610 [CDK] - Allow .bat extensions when locating a given command

robstryker:


alexeykazakov:
Added JBIDE-21610 to the comment's message, rebased, and pushed to 4.3.x. Also cherry-picked to master.

PR #938, opened 2016-02-02
JBIDE-21622 - increase timeout for CDK server to 800 seconds or 13 min.

robstryker:


fbricon:
might as well go up to  900sec, that's 15min, a good ol' round number

PR #945, opened 2016-02-03
[JBIDE-21624] Trigger a new deployment from deploymentconfig or deplo…

jcantrill:
…yment(service)

cc @fbricon @adietish


plugins/org.jboss.tools.openshift.ui/plugin.xml
```diff
@@ -590,6 +608,16 @@
      <command
            commandId="org.eclipse.ui.edit.delete"
            style="push" />
+     <separator
+           name="org.jboss.tools.openshift.ui.separator1"
+           visible="true">
+     </separator>
+     <command
```
fbricon: I'd put the menu before the common edit/delete menus

fbricon:
So after deploying from the explorer, I tried to deploy a 2nd time from the properties tab, but got an exception:

      Unable to perform a deployment for config {0}
      Could not update resource nodejs-mongodb-example: {
        "kind": "Status",
        "apiVersion": "v1",
        "metadata": {},
        "status": "Failure",
        "message": "deploymentConfig \"nodejs-mongodb-example\" cannot be updated: the object has been modified; please apply your changes to the latest version and try again",
        "reason": "Conflict",
        "details": {
          "name": "nodejs-mongodb-example",
          "kind": "deploymentConfig"
        },
        "code": 409
      }

fbricon:
Also, the pods list in properties view is not updated when triggering a deploy from the explorer

plugins/org.jboss.tools.openshift.ui/plugin.xml
jcantrill: updated.

jcantrill:
@fbricon made a change to how the event is fired which would affect deployments. Deployments are replicationControllers and should fire this event but was firing 'replicationcontrollers'

fbricon:
Everything seems to work fine now. +1 to apply

PR #947, opened 2016-02-03
[JBIDE-19506] added tests for ConnectionWizardPage

adietish:


tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/common/core/connection/ConnectionsFactoryTest.java
```diff
@@ -0,0 +1,109 @@
...
+import java.io.IOException;
+
+import org.jboss.tools.openshift.common.core.connection.ConnectionsFactory;
+import org.jboss.tools.openshift.common.core.connection.IConnection;
+import org.jboss.tools.openshift.common.core.connection.IConnectionFactory;
+import org.jboss.tools.openshift.test.core.connection.ConnectionFake;
+
+/**
+ * @author Andre Dietisheim
+ */
```
fbricon: this is not a test class, as it has no test methods.

Tycho fails with 

      initializationError(org.jboss.tools.openshift.test.common.core.connection.ConnectionsFactoryTest)  Time elapsed: 0.002 sec  <<< ERROR!
      java.lang.Exception: No runnable methods

Either make the class abstract or you need to write actual tests

adietish: no tests yet, this is so far only the infrastructure for it.

PR #957, opened 2016-02-12
[JBIDE-21670] Remove watch on unwatchable kind

jcantrill:


jcantrill:
@adietish I think this will affect the server adapter code and the imagestreamtag logic.

PR #958, opened 2016-02-15
[JBIDE-21665] Refresh edit resource document after update

jcantrill:
@fbricon @adietish  simplest fix to the referenced issue.  It is missing prompting the user to accept the refresh if they have made modifications.  I'm not certain how to determine that I have an edit and need to prompt the user to accept the update.  Key listener to know my local copy has changed?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/OpenShiftResourceDocumentProvider.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/OpenShiftResourceDocumentProvider.java
@@ -10,15 +10,17 @@
  ******************************************************************************/
 package org.jboss.tools.openshift.internal.ui.property;
 
-import java.util.Collection;
-import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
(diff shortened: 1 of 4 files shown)


adietish:
@jcantrill IEditorPart#isDirty will tell you if there's any unsaved data in the editor. But then you'd need to get hold of the editor part within the resource document provider which I dont see how you could achieve. Looking further at the API I detected IDocumentProvider#canSaveDocument(element) which should be equivalent to 
```
OpenShiftResourceDocumentProvicer.canSaveDocument(OpenShiftResourceInput)
```

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/OpenShiftResourceDocumentProvider.java
```diff
@@ -53,9 +51,27 @@ protected IDocument createDocument(Object element) throws CoreException {
...
 		}
 		return document;
 	}
+	
+	
+	
+	private synchronized void registerConnectionListener(OpenShiftResourceInput input) {
+		listener = new ConnectionListener(input, this);
+		ConnectionsRegistrySingleton.getInstance().addListener(listener);
+	}
```
scabanovich: I am not quite sure about lifecycle of this document provider, but in general, one instance of document provider can be used to manage documents for many elements. If this is the case, the listener for one element will be overwritten at creating another document for another element and will never be removed from connection registry.

As a variant, connection listeners can be managed by overriding methods createElementInfo(element) and disposeElementInfo(element, info) and be stored in Map<ElementInfo, ConnectionListener>

jcantrill:
@fbricon  updated to include nag notification if the content changes from underneath you.  addressed changes recommend by @scabanovich

fbricon:
mmm when I edit the # of replicas on a deployment config and save, nothing happens. The resource version is incremented as expected but, no new pod is spun up

fbricon:
Works perfectly. Applied in master / 4.3.x

PR #960, opened 2016-02-16
[JBIDE-21703] Retrieve istags associated with a resource

jcantrill:
@adietish can you test to see if this resolves the issue.  I'm not certain what the actions are to verify.

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
@@ -369,6 +369,17 @@ public class Connection extends ObservablePojo implements IConnection, IRefresha
 		}
 	}
 
+	@Override
+	public <T extends IResource> T getResource(String kind, String namespace, String name) {
+		try {
+			if(client.getAuthorizationStrategy() == null) {
+				client.setAuthorizationStrategy(getAuthorizationStrategy());
+			}
+			return client.get(kind, name, namespace);
+		} catch (UnauthorizedException e) {
+			return retryGet("Unauthorized.  Trying to reauthenticate", e, kind, name, namespace);
+		}
+	}
 	/**
 	 * Get or refresh a resource
 	 * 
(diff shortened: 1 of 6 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
```diff
@@ -382,16 +393,16 @@ public String toString() {
 			}
 			return client.get(resource.getKind(), resource.getName(), resource.getNamespace());
 		} catch (UnauthorizedException e) {
-			return retryGet("Unauthorized.  Trying to reauthenticate", e, resource);
+			return retryGet("Unauthorized.  Trying to reauthenticate", e, resource.getKind(), resource.getName(), resource.getNamespace());
```
adietish: trying to understand: how is this related to the inability to retrieve istags? Did the backend change?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/IOpenShiftConnection.java
```diff
@@ -31,6 +31,15 @@
...
 	<T extends IResource> List<T> getResources(String kind, String namespace);
+
+	/**
+	 * Retrieve a resource by name
+	 * @param kind
+	 * @param namespace
+	 * @param name
+	 * @return
+	 */
+	<T extends IResource> T getResource(String kind, String namespace, String name);
```
adietish: ok, so we now need this method?

adietish:
great, that was quick, thanks!
+1

adietish:
unfortunately this still isnt working for me, an empty set of istags is retrieved. Looking into it.

fbricon:
Works for me, the deployment directory is properly set in the server wizard and the server editor in PR #944

adietish:
I intentified the problem: if you get back to the connection page and choose a different one, the deployment resource mappers are not repopulated correctly, causing this to fail in this particular case.

adietish:
merged to jbosstools-4.3.x and master.
filing a new issue for the lacking reloading of the mappers in case you change the connection

PR #961, opened 2016-02-17
[JBIDE-21686] reduced snchronized blocks to prevent deadlocks

adietish:


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftProjectCache.java
```diff
@@ -50,79 +52,103 @@ public OpenShiftProjectCache() {
...
+				&& (newValue instanceof List)) {
+			IOpenShiftConnection openshiftConnection = (IOpenShiftConnection) connection;
+			String key = getCacheKey(openshiftConnection);
+			Set<IProjectAdapter> adapters = getOrCreateEntry(key);
+			addAndRemoveAdaptersFor((List<IProject>) newValue, (List<IProject>) oldValue, openshiftConnection, adapters);
+		}
+	}
+
+	private Set<IProjectAdapter> getOrCreateEntry(String key) {
+		Set<IProjectAdapter> adapters = cache.get(key);
```
jcantrill: Do we need to synchronize on the cache here so we dont have 2 threads trying to map to a key at the same time

jcantrill:
Outside of my comments, LGTM if it resolves the deadlocks

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftProjectCache.java
adietish: yes you're right. We discussed this and I then got the wrong assumption that the concurrent hashmap would guarantee this. But reading the javadoc clearly states that retrieval- and update-operations may overlaps since retrieval is not locking at all, only updates do.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftProjectCache.java
```diff
@@ -136,35 +162,45 @@ public void connectionRemoved(IConnection connection) {
...
+		if (adapter == null) {
+			return;
+		}
 		synchronized (listeners) {
 			try {
-				for (IProjectAdapter a : adapters) {
-					listeners.forEach(l->l.handleAddToCache(this, a));
-				}
-			}catch(Exception e) {
+				listeners.forEach(l -> l.handleAddToCache(this, adapter));
```
adietish: I kept those catches but for sinle updates only so that notifying the other listeners would still be performed even if 1 of them throws.

adietish:
@fbricon plz test my PR

fbricon:
wrapping ConcurrentHashMap calls in synchronized blocks defeats the purpose of using that class http://stackoverflow.com/questions/25998536/does-a-concurrenthashmap-need-wrapped-in-a-synchronized-block

fbricon:
Anyways, I can reproduce the bug without the fix. When the fix is applied, the problem goes away

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftProjectCache.java
```diff
@@ -50,79 +52,119 @@ public OpenShiftProjectCache() {
...
+				notifyAdd(adapter);
 			}
-			cache.put(key, adapters);
-			notifyAdd(adapters);
+		} else {
+			adapters = getEntry(key);
 		}
-		return new ArrayList<>(cache.get(key));
+
+		return new ArrayList<>(adapters);
```
scabanovich: The last little hole. Creating list should also be synchronized on adapters.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftProjectCache.java
```diff
@@ -50,79 +52,119 @@ public OpenShiftProjectCache() {
 	}
 	
 	@Override
-	public synchronized Collection<IProjectAdapter> getProjectsFor(final IOpenShiftConnection conn) {
+	public Collection<IProjectAdapter> getProjectsFor(final IOpenShiftConnection conn) {
 		String key = getCacheKey(conn);
-		if(!cache.containsKey(key)) {
+		Set<IProjectAdapter> adapters = null;
```
scabanovich: Most safe is 
Set<IProjectAdapter> adapters = getEntry(key);
if(adapters == null) {
   ...
}
Otherwise, in a moment between contains(key) and adapters = getEntry(key), entry may disappear. 
This is not very probable, but the one thing which is a bit more probable is that this entry may be removed while projects are iterated and listeners are notified of an adapter added.

scabanovich: On the second glance it seems that this logic is not protected against concurrently creating 2 entries by two threads that call this method at approximately the same time and both passed !contains(key). Cache will finally have only one entry, but listeners will be notified of adapters created for both collections, in random order.
Possible code that fully protects this method against this case may be
boolean newAdapters = false;
Set&lt;IProjectAdapter> adapters = null;
synchronized(cache) {
    adapters = getEntry(key);
    if(adapters == null) {
        adapters = createEntry(key);
        newAdapters = true;
    }
}
if(newAdapters) {
    ....  //iterate projects.
}
However, that is not all. This method may be concurrently called with other methods that modify cache. Synchronization on cache and adapters prevents concurrent modification exceptions, but it does not save from listeners being notified of changes in an unpredictable order.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftProjectCache.java
adietish: I dont think so. Modifying the adapters is only done on an new (cache-) entry. Creating the entry is locking, thus 2 thread wont be able to both see the entry as being new.

scabanovich: There is adapters.remove() in addAndRemoveAdaptersFor() called on connectionChanged().

PR #988, opened 2016-02-29
[JBIDE-21787] Added wizard title to New OpenShift Connection

mlabuda:


plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/OpenShiftExplorerView.java
```diff
@@ -123,7 +123,8 @@ private Control createExplanationPane(Control connectionsPane, PageBook pageBook
 	private SelectionAdapter onExplanationClicked(final Control connectionsPane, final Control explanationPane) {
 		return new SelectionAdapter() {
 			public void widgetSelected(SelectionEvent e) {
-				ConnectionWizard wizard = new ConnectionWizard();
+				ConnectionWizard wizard = new ConnectionWizard();				
+				wizard.setWindowTitle("New OpenShift Connection");
```
adietish: @mlabuda this is the right API-call but I would directly put into the constructor of the ConnectionWizard (https://github.com/jbosstools/jbosstools-openshift/blob/master/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizard.java#L41) so that this also gets fixed for all the other cases

mlabuda: @adietish  I'd thought about putting it into the constructor, but then I spotted this wizard is used also for modification (Edit) of an existing connection, where is different title, so I put it there to let it be in responsibilities of the one who is using this wizard to put correct title for it. But if you want, I could put it to constructor to let it be default and who wanna reuse the wizard would have to call set title method.

adietish: @mlabuda ok, then I would still have the call in the contructor, but then have the caller pass the title in as a parameter to the constructor

mlabuda:
@adietish  please review

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizard.java
```diff
@@ -21,24 +21,32 @@
...
 	
+	public static final String NEW_CONNECTION_TITLE = "New OpenShift Connection";
+	public static final String EDIT_CONNECTION_TITLE = "Edit OpenShift Connection";
+	
 	/**
 	 * Constructor to use when connecting with the default connection.
+	 * 
+	 * @param connection wizard shell title
 	 */
-	public ConnectionWizard() {
```
fbricon: I'd keep the default constructor that'd send ConnectionWizard.NEW_CONNECTION_TITLE as title, way less code changes that way

mlabuda:
Ready for review.

fbricon:
Applied in master / 4.3.x (after changing the commit message format).
Thanks @mlabuda !

PR #989, opened 2016-02-29
[JBIDE-21789] filtering non-v2 connections for v2 server wizard

adietish:


adietish:

PR #990, opened 2016-02-29
JBIDE-21170 - issue finding vagrant in mac when launched via .app

robstryker:
Adds -i -l options to CommandLocationLookupStrategy for Mac

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
@@ -20,6 +20,7 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Platform;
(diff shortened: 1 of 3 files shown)


mmalina:
This doesn't improve anything as far as I can see. There are three places where the PATH was causing problems:
1. Inability to find vagrant - I think this worked even before this patch and from discussion with Rob this is the only thing the patch was intended to fix
2. Once the vagrant process is launched, it fails to find VBoxManage, because the process lacks proper PATH
3. Even if vagrant up successfully starts the box, the launch then fails, because the call to "vagrant status" fails to find vagrant.

So from I can see, 2. and 3. still don't work.

mmalina:
I just tried the latest iteration of the fix (c6c5333a84f99eb84e990f3f1eb493196811f806) and now even the vagrant location field in launch config is empty, so it won't work at all. 
And even if I entered the path to vagrant manually, an attempt to start cdk failed like before:
`Vagrant could not detect VirtualBox! Make sure VirtualBox is properly installed.
Vagrant uses the VBoxManage binary that ships with VirtualBox, and requires
this to be available on the PATH. If VirtualBox is installed, please find the
VBoxManage binary and add it to the PATH environmental variable.`

robstryker:
Martin is on PTO now, but, we worked togethre on this issue last night and he says it fixed hte issue.

[3:52:31 PM] Martin Malina: wow, it's there now
[3:53:11 PM] Martin Malina: wait just a second
[3:53:19 PM] Martin Malina: it would be good if I actually started the server
[4:00:33 PM] Martin Malina: it's starting now
[4:02:19 PM] Martin Malina: it's started now. so looks good

alexeykazakov:
LGTM

robstryker:
merged after rebase

PR #994, opened 2016-03-02
JBIDE-21808 - outdated "import package" declaration in "org.jboss.tools.openshift.test"

xcoulon:
replaced import-packages directive with additional entries in the required-bundles


adietish:
+1, thanks!
Beside cleaning up. adds consistency where we now only have Required-Bundle and not a mix with Import-Package any more

adietish:
merged into jbosstools-4.3.x, cherry-picked to master

PR #995, opened 2016-03-02
[JBIDE-21590] Wait for server to reconcile project delete

jcantrill:
Provides a wait for the server to reconcile the project deletion.

cc @mlabuda @fbricon @adietish


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/OpenShiftJobs.java
```diff
@@ -39,6 +44,15 @@ private OpenShiftJobs() {}
...
 	public static DeleteResourceJob createDeleteProjectJob(final IProject project) {
+		return createDeleteProjectJob(project, null);
+	}
+	
+	/**
+	 * Creates a {@link DeleteResourceJob} to delete an OpenShift {@link IProject}.
+	 * @param project
+	 * @param callIfNotOK - A callback to execite if the job does not return OK status 
+	 */
+	public static DeleteResourceJob createDeleteProjectJob(final IProject project, Runnable callIfNotOK) {
```
adietish: the Runnable "callIfNotOK" looks abit redundant/duplicate to what IJobChangeListener offers: The one that creates the delete job could attach a job change listener to it and implement #done, inspecting the IStatus that's in the IJobChangeEvent he'll get.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/OpenShiftJobs.java
```diff
@@ -56,11 +71,39 @@ protected IStatus doRun(IProgressMonitor monitor) {
...
+				OpenShiftException e = null;
+				long sleep = 0;
+				do {
+					try {
+						conn.getResource(project);
+						Thread.sleep(PROJECT_DELETE_DELAY);
+						sleep = sleep + PROJECT_DELETE_DELAY;
+					} catch (InterruptedException e1) {
+					} catch(OpenShiftException ex) {
+						e = ex;
```
adietish: shouldn't we make sure that the exception we get really is an exception caused by the fact that the project isnt there any longer?

adietish:
other than my comments, code seems legit. It even handles multiple invocation from context menu (adapter is set to 'deleting' while it's deleting) and adds tests. Nice!

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/OpenShiftJobs.java
jcantrill: You are saying we should query if it is a 401?

jcantrill:
@adietish made an update but I see THIS event sometimes doesnt work based on the new logic.  I'm thinking it would actually be better to have a blocking progress modal that times out after a max time.  I noticed another situation the disabled delete doesnt address where i was able to 'expand' a deleted project which threw a different exception.  WDYT about making delete project 'blocking' from the context  menu?

adietish:
@jcantrill Do you know why exactly it doesnt work? Frequently this happens if the listeners gets attached after the job was launched and teminates. But I dont see this being true for your code. Maybe you should add brk-points to the other methods (#sleeping, #scheduled, #awake etc.) to further understand when this wont get into #done.
I believe that the job should terminate itself for a given timeout, at least that seems legit to me in terms of separations of concerns (the thread terminates itself vs gets terminated) - this is imho also the logic in jdk with the deprecation of Thread#stop.
For the case where throwing a different exception was not preventing a concurrent deleting thread I'd think that we then should maybe catch all exceptions?
Blocking seems like a bad idea - but maybe I get you wrong - since this would also block the UI thread, something we want to avoid.
Maybe we should discuss this in more details in bj?

jcantrill:
Updated to reduce contention points of using sync blocks.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftProjectCache.java
```diff
@@ -88,27 +86,26 @@ public void connectionChanged(IConnection connection, String property, Object ol
 				&& (oldValue instanceof List)
 				&& (newValue instanceof List)) {
 			IOpenShiftConnection openshiftConnection = (IOpenShiftConnection) connection;
-			String key = getCacheKey(openshiftConnection);
-			Set<IProjectAdapter> adapters = getOrCreateEntry(key);
+			Set<IProjectAdapter> adapters = getCacheFor(openshiftConnection);
```
adietish: i see a possible race-condition where 2 threads concurrently query the cache and create entries here since only #getCacheFor is locking the cache but the latter #createEntry isnt.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftProjectCache.java
```diff
@@ -53,20 +53,18 @@ public OpenShiftProjectCache() {
...
-		Set<IProjectAdapter> adapters = getOrCreateEntry(key);
-		if (adapters.isEmpty()) {
+		Set<IProjectAdapter> adapters = getCacheFor(conn);
+		if(adapters == null) {
 			// new entry
 			Collection<IProject> projects = conn.<IProject>getResources(ResourceKind.PROJECT);
-			for (IProject project : projects) {
-				IProjectAdapter adapter = addNewProjectAdapter(adapters, conn, project);
-				notifyAdd(adapter);
+			adapters = createEntry(conn);
```
adietish: here too (see below): what if 2 threads get in here concurrently, query the cache and create entries? checking if entry exists and creating an entry are not atomic, thus not thread safe I believe

adietish:
@jcantrill looks good now since querying cache and creating a new entry is atomic. Will test tomorrow

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/OpenShiftJobs.java
```diff
@@ -49,18 +55,43 @@ protected IStatus doRun(IProgressMonitor monitor) {
...
+				boolean deleted = false;
+				long sleep = 0;
+				do {
+					try {
+						conn.getResource(project);
+						Thread.sleep(PROJECT_DELETE_DELAY);
+						sleep = sleep + PROJECT_DELETE_DELAY;
+					} catch (InterruptedException e1) {
+					} catch(OpenShiftException ex) {
+						if(ex.getStatus() != null) {
```
adietish: testing this I see the DeleteResourceJob enlessly looping. When breaking in I see ex.getStatus() being null it thus never gets into the termnal condition. "ex" is com.openshift.restclient.authorization.ResourceForbiddenException, the message is 'User "adietish@redhat.com" cannot get projects in project "testproj"'.

adietish: The actual reason for the missing status is an issue in the client lib:

DefaultClient#createOpenShiftException()
```
			if(status.getCode() == STATUS_FORBIDDEN) {
				if(StringUtils.isNotBlank(token)) { //truly forbidden
					return new ResourceForbiddenException(status.getMessage(), e);
```
The resourceForbiddenException that is being constructed simply wont add the status, it only adds the message.

jcantrill:
This https://github.com/openshift/openshift-restclient-java/commit/62cadaba3e17651f113c036d9cd523f539f2d9e5 fixed that problem.  I will update to add a check against null status to be safe.

jcantrill:
There is a different issue in the total sleep never is advanced because of the exception so you never reach the max value.  I will fix here.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/OpenShiftJobs.java
```diff
@@ -49,18 +55,43 @@ protected IStatus doRun(IProgressMonitor monitor) {
...
+							final int code = ex.getStatus().getCode();
+							if(code == IHttpConstants.STATUS_NOT_FOUND || code == IHttpConstants.STATUS_FORBIDDEN) {
+								deleted = true;
+							}
+						}
+					}
+				}while(!deleted || sleep > MAX_PROJECT_DELETE_DELAY);
```
adietish: Shouldnt the condition to stop retrying be:
```
!deleted && sleep < MAX_PROJECT_DELETE_DELAY
```
?

jcantrill:
merged.

adietish:
tested it, it works for me. +1

PR #1003, opened 2016-03-03
JBDS-3662 CDK Launch config has broken PATH env variable value set

dgolovin:
do PATH workaround for Mac OS X only


alexeykazakov:
@robstryker please review. But if this PR breaks anything else we should bring back Windows support first ASAP and then work on solution which will fix Mac too.

robstryker:
This is the wrong solution and breaks windows for users not launching from cygwin

PR #1004, opened 2016-03-03
JBIDE-21797 - vagrant 1.8.x has different output to vagrant status --…

robstryker:
…machine-readable, adds line with no vagrant id breaking our parse code expectations


fbricon:
+1 to apply after squashing the commits

PR #1007, opened 2016-03-04
JBIDE-21761 - Port Mapping validation is incorrect

xcoulon:
Fixed the validator logic
Also:
 - move the explanation label under the 'Pod Port' text field
 - reduced the dialog margin to fit with the Eclipse UI

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicePortDialog.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicePortDialog.java
@@ -38,44 +38,46 @@ import org.jboss.tools.openshift.internal.common.ui.wizard.OkCancelButtonWizardD
 
 import com.openshift.restclient.model.IServicePort;
 
+/**
+ * Dialog to configure the port mapping for a Docker Image to be deployed.
+ */
 public class ServicePortDialog extends AbstractOpenShiftWizardPage {
 
 	static final String PROPERTY_SERVICE_PORT = "port";
 	static final String PROPERTY_POD_PORT = "targetPort";
 	
-	private IServicePort model;
-	private List<IServicePort> ports;
-	private final int servicePort;
-	private final String podPort;
+	private final IServicePort model;
+	private final List<IServicePort> ports;
 
-	public ServicePortDialog(IServicePort model, String message, List<IServicePort> ports) {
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicePortDialog.java
```diff
@@ -140,26 +141,23 @@ public IStatus validate(Object value) {
 	}
 	class PodPortValidator implements IValidator{
 		
-		private final int MAXLENGTH = 63;
-		private final Pattern REGEXP = Pattern.compile("[a-z0-9]([a-z0-9-]*[a-z0-9])*");
-
-		private final IStatus ERROR = ValidationStatus.error("The target port must be at most 15 characters, matching regex [a-z0-9]([a-z0-9-]*[a-z0-9])*, and hyphens cannot be adjacent to other hyphens): e.g. \"http\"");
```
adietish: as far as I understood ports can be numerical but also alphanumeric names? https://issues.jboss.org/browse/JBIDE-21761?focusedCommentId=13169586&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13169586

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicePortDialog.java
```diff
@@ -140,26 +141,23 @@ public IStatus validate(Object value) {
...
 			if(StringUtils.isEmpty(value)) {
 				return ERROR;
 			}
-			String newPort = (String) value;
-			if(!newPort.equals(podPort)) {
-				if(newPort.length() > MAXLENGTH || !REGEXP.matcher(newPort).matches()) {
+			try {
+				final long portNumber = Long.valueOf(value.toString());
+				// validate range
+				if(portNumber < 0 || portNumber > 65535) {
```
adietish: we still only validate to numbers? Shouldn't we keep the old validator but change the error-message? ports on the pod can be numbers or alphanum names.

xcoulon:
Pull request updated. 
@adietish would you mind taking another look at it ?

fbricon:
Applied in master / 4.3.x, after removing the stale "Pod port is linked to service port changes" label

PR #1012, opened 2016-03-07
[JBIDE-21802] ignore <none>:<none> Docker images

fbricon:
Not really pretty way to ignore <none>:<none> images. @xcoulon, I'd be curious to know if there's a proper way to get rid of the noise.

Also made the code more Java 8'y


xcoulon:
@fbricon thanks for this patch. Using lambda expressions is nice. There's actually a more direct way to exclude intermediate and dangling images, using the `Image#isDangling()` and `Image#isIntermediateImage()` methods when initializing the list of image names in the `DeployImageWizardModel` class.
Please see https://github.com/jbosstools/jbosstools-openshift/pull/1013 instead.

fbricon:
Superseded by #1013

PR #1016, opened 2016-03-09
[JBIDE-21846] fix deploy image name validator when set by setting the  image

jcantrill:
This PR
* Renames deployimagename validator to servicenamevalidator
* Adds validation tests
* Fixes name validation when model#setName is called by model#setImage


jcantrill:
cc @adietish  @fbricon @scabanovich

fbricon:
LGTM but I'm not a databinding expert. I'd rather prefer if @adietish or @scabanovich could comment

adietish:
did not test but in reading code +1

PR #1017, opened 2016-03-09
JBIDE-14828 - allow user to skip the clone step

robstryker:


fbricon:
I'd prefer if we could select a folder actually containing an existing .git repo (instead of the parent), but we can still refine it later.
As for now, the patch does the job and LGTM.
+1

adietish:
cherry-picked to master and merged into jbosstools-4.3.x

PR #1021, opened 2016-03-10
[JBIDE-21559] Allow access to specific pod container logs

jcantrill:
cc @fbricon @adietish

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/PodLogsHandler.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/PodLogsHandler.java
@@ -10,13 +10,19 @@
  ******************************************************************************/
 package org.jboss.tools.openshift.internal.ui.handler;
 
+import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.ArrayUtils;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.viewers.ISelection;
+import org.eclipse.jface.viewers.LabelProvider;
+import org.eclipse.jface.window.Window;
 import org.eclipse.osgi.util.NLS;
+import org.eclipse.ui.dialogs.ElementListSelectionDialog;
 import org.eclipse.ui.handlers.HandlerUtil;
 import org.jboss.tools.openshift.core.OpenShiftAPIAnnotations;
 import org.jboss.tools.openshift.core.connection.Connection;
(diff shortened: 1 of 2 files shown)


scabanovich:
In ConsoleListener.consolesRemoved(IConsole[] consoles) the line REGISTRY.remove(pod) has to be changed to remove Key object as key in the map.

It may be better to create Key instance right in the constructor of PodLogsJob, and make one field Key key that gives access to pod and containerName.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/PodLogsHandler.java
```diff
@@ -59,7 +66,25 @@ private void showLogs(IPod pod, ExecutionEvent event) {
...
-		new PodLogsJob(pod).schedule();
+		Collection<IContainer> containers = pod.getContainers();
+		String containerName = null;
+		if(containers.size() > 1) {
+			List<String> names = containers.stream().map(c->c.getName()).collect(Collectors.toList());
+			Collections.sort(names);
+			ElementListSelectionDialog dialog = new ElementListSelectionDialog(HandlerUtil.getActiveShell(event), new LabelProvider());
+			dialog.setElements(names.toArray());
+			dialog.setTitle("Pod Containers");
+			dialog.setMessage("Which container logs do you need?");
```
scabanovich: Could message be like "Select a container log" ?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/PodLogsHandler.java
```diff
@@ -59,7 +66,25 @@ private void showLogs(IPod pod, ExecutionEvent event) {
...
+			dialog.setMultipleSelection(false);
+			int result = dialog.open();
+			if(Window.CANCEL == result) {
+				return;
+			}
```
scabanovich: Can user have an option to decline container selection but not to cancel the job and continue as in the case containers.size() == 0 ?

jcantrill:
@scabanovich updated pr per your suggested changes

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/PodLogsHandler.java
```diff
@@ -59,7 +66,25 @@ private void showLogs(IPod pod, ExecutionEvent event) {
...
+			containerName = (String) dialog.getFirstResult();
+		}else if (containers.size() == 1){
+			containerName = containers.iterator().next().getName();
+		}
+		new PodLogsJob(pod, containerName).schedule();
```
fbricon: technically if containers.size() == 0, then containerName == null, so no need to schedule the job

jcantrill: @fbricon added check for no container though I would imagine the server won't allow this case.

fbricon:
+1 to apply once https://github.com/openshift/openshift-restclient-java/pull/127 is merged and the client pushed to nexus

jcantrill:
merged.

PR #1023, opened 2016-03-11
[JBIDE-21449] fix multiple callbacks from WatchManager

jcantrill:
fix multiple notifications by reusing listener.

cc @fbricon @adietish


adietish:
thanks!
rebased and merged into jbosstools-4.3.x, cherry-picked to master

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -25,6 +28,7 @@
 import org.jboss.tools.openshift.core.connection.Connection;
 import org.jboss.tools.openshift.core.connection.ConnectionProperties;
 import org.jboss.tools.openshift.core.connection.ConnectionsRegistryUtil;
+import org.omg.PortableInterceptor.DISCARDING;
```
fbricon: This import got lost

PR #1026, opened 2016-03-14
JBIDE-21828 - OpenShift Explorer view flickers when refreshing a service

xcoulon:
Refactoring the DeploymentResourceMapper and its parent class
 OpenShiftProjectUIModel to avoid keeping reference to IResource
 in both places.

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
+++ b/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
@@ -23,6 +23,7 @@ import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jface.viewers.ITreeContentProvider;
+import org.eclipse.jface.viewers.TreePath;
 import org.eclipse.jface.viewers.TreeViewer;
 import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.swt.widgets.Control;
@@ -224,11 +225,14 @@ public abstract class BaseExplorerContentProvider implements ITreeContentProvide
 		asyncViewerRefresh(new Runnable() {
 			public void run() {
 				synchronized (viewer) {
-					if(object != null) {
+					//final TreePath[] treePaths = viewer.getExpandedTreePaths();
+					if (object != null) {
 						viewer.refresh(object);
-					}else {
+					} else {
 						viewer.refresh();
 					}
+					// This line causes an endless load on linux when a node has zero children. (JCantrill)
+					//viewer.setExpandedTreePaths(treePaths);
 				}
 			}
 		});
@@ -278,9 +282,6 @@ public abstract class BaseExplorerContentProvider implements ITreeContentProvide
 		
 		private List<Object> children = new ArrayList<Object>();
 
-		public LoadingStub() {
-		}
-		
 		public void add(Exception e) {
 			children.add(e);
 		}
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
@@ -15,7 +15,9 @@ import java.beans.PropertyChangeListener;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.eclipse.core.databinding.observable.Diffs;
@@ -32,6 +34,7 @@ import org.jboss.tools.openshift.internal.ui.models.IAncestorable;
 import org.jboss.tools.openshift.internal.ui.models.IProjectAdapter;
 import org.jboss.tools.openshift.internal.ui.models.IProjectCache;
 import org.jboss.tools.openshift.internal.ui.models.IResourceUIModel;
+import org.jboss.tools.openshift.internal.ui.models.IResourcesUIModel;
 import org.jboss.tools.openshift.internal.ui.models.OpenShiftProjectCache;
 
 import com.openshift.restclient.OpenShiftException;
@@ -50,6 +53,9 @@ public class OpenShiftExplorerContentProvider extends BaseExplorerContentProvide
 	private static final List<String> TERMINATED_STATUS = Arrays.asList("Complete", "Failed", "Error", "Cancelled");
 	private static final IProjectCache cache = new OpenShiftProjectCache();
 	
+	/**
+	 * Constructor.
+	 */
 	public OpenShiftExplorerContentProvider() {
 		super();
 		cache.addListener(this);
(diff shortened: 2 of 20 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
```diff
@@ -80,7 +84,9 @@ public String getText(Object element) {
 
 	@Override
 	public Image getImage(Object element) {
-		if (element instanceof Deployment) {
+		if (element instanceof OpenShiftProjectUIModel) {
```
jcantrill: This should be the ProjectAdapter interface this model implements instead of the concrete class.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -192,38 +199,41 @@ public void handleAdd(int index, Object element) {
...
-				if(child instanceof IResourceUIModel && ResourceKind.ROUTE.equals(((IResourceUIModel) child).getResource().getKind())) {
-					updateChildrenFromViewer(parent);
-				}
-				//HACK to fix JBIDE-21458
-				if(newList.size() > 0 && (parent instanceof IProjectAdapter || parent instanceof Deployment)) {
-					refreshViewer(parent);
-				}else {
-					addChildrenToViewer(parent, child);
-				}
+//				if(child instanceof IResourceUIModel && ResourceKind.ROUTE.equals(((IResourceUIModel) child).getResource().getKind())) {
```
jcantrill: Is this code intended to still be commented?  This particular line explicitly checks routes because we haven chosen to display in the tree 'service - route'.  On first glance, i'm not certain the UI will reflect the case where a service exists and then a route is added.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -192,38 +199,41 @@ public void handleAdd(int index, Object element) {
...
+//					updateChildrenFromViewer(parent);
+//				}
+//				//HACK to fix JBIDE-21458
+//				if(newList.size() > 0 && (parent instanceof IProjectAdapter || parent instanceof Deployment)) {
```
jcantrill: This was added to address a 'phantom' node issue where if there was only one child the UI didn't refresh properly.  Is that addressed elsewhere?

jcantrill:
Initial pass looks good. I will test tomorrow to see if i notice any issues with regards to the expected functionality.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
xcoulon: ok, fixed.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
xcoulon: @jcantrill That's right. I reverted the comment on lines 204-206.
To verify, I applied changes from CLI using the following command : `oc delete route -l app=jee-sample` and the route disappeared from the OpenShift Explorer view. I obtained the same result when deleting the route from the 'Routes' tab in the properties view when the service was selected.

Then I used the `oc expose service jee-sample` command to expose again the service and the OpenShift Explorer view immediately displayed the route as a qualifier/decorator of the service.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
xcoulon: @jcantrill I was not able to reproduce the 'phantom' node issue you're referring to. Could you verify it or give me steps to reproduce ?

fbricon:
tested against the latest version in 4.3.x, I created a new nodejs + mongodb app. I can see the 2 services in my project, and I briefly (<1sec) saw a build pod then it disappeared. 
I had to refresh the project to see the running pod.
I then clicked Deploy Latest on the service. The pod disappeared again from the explorer.
Having the properties view opened, I can see the new pods show up there, just not in the explorer view.

jcantrill:
@xcoulon adding my issues here:

1. BaseExplorerContentProvder#refreshViewer is still causing me problems when my connection has no children.  Commenting out the last line of expanding the nodes resolves it for me on linux
2. Latest changes no longer allow provide the same right-click context menu items I had before (e.g. new app, deploy image, etc)

xcoulon:
@jcantrill I updated the patch to restore all missing commands in the context menu, both at the connection level and at the project level.

> BaseExplorerContentProvder#refreshViewer is still causing me problems when my connection has no children. Commenting out the last line of expanding the nodes resolves it for me on linux
Can you give me more details on how to reproduce this issue ? I can't reproduce this bug on MacOSX. @adietish, would you have time to try the patch and let us know if this problem also occurs on your machine ?

jcantrill:
image
@xcoulon I see the context menu items are back but so is the dummy node problem.  This is the view on linux after the pod appeared and the build node disappeared.

adietish:
current impl flickers for me if there's no project:
# EXEC: removed my existing project in Explorer

Result: "Loading..." child appears for the connection and it stays there endlessly. If I move the mouse over it I notice the mouse pointer turning hourglass/pointer/hourglass/pointer endlessly. Similary a selection on it is reverted immediately.

https://youtu.be/F78vlYmviR8

jcantrill:
Changes in the content provider property change removed from the original version that gets rid of the dummy node issue: https://gist.github.com/jcantrill/37cf76925b3c5bcb4292

xcoulon:
@jcantrill ok, I included your gist in the patch. Can you give it another try. Frankly, I'm not fully happy with this patch because I can still see some flickering (although, maybe less than before) and moreover, all service nodes get collapsed upon refresh completion.
I guess we could fix that later if we want to move on with this patch, though...
@adietish @jcantrill @fbricon please let me know if this patch is acceptable now.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
```diff
@@ -224,11 +225,13 @@ protected void refreshViewer(final Object object) {
...
 				synchronized (viewer) {
-					if(object != null) {
+					final TreePath[] treePaths = viewer.getExpandedTreePaths();
+					if (object != null) {
 						viewer.refresh(object);
-					}else {
+					} else {
 						viewer.refresh();
 					}
+					viewer.setExpandedTreePaths(treePaths);
```
jcantrill: This line causes an endless load on linux when a node has zero children.

jcantrill:
@fbricon @xcoulon  I think it needs a rebase, but it appears to address the issues I identified about.  LGTM if you want to merge, but I would understand if we wanted to wait until after code freeze.

xcoulon:
@fbricon I rebased the patch again, as @jcantrill suggested.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
```diff
@@ -224,11 +225,14 @@ protected void refreshViewer(final Object object) {
...
-					if(object != null) {
+					//final TreePath[] treePaths = viewer.getExpandedTreePaths();
+					if (object != null) {
 						viewer.refresh(object);
-					}else {
+					} else {
 						viewer.refresh();
 					}
+					// This line causes an endless load on linux when a node has zero children. (JCantrill)
+					//viewer.setExpandedTreePaths(treePaths);
```
scabanovich: Without it, expand state is lost. Can we check which tree paths are safe to set expanded and call this method with filtered list?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftProjectUIModel.java
```diff
@@ -90,48 +91,85 @@ public IProject getProject() {
...
+
+	@Override
+	public Collection<IResourceUIModel> getBuilds() {
+		return this.mapper.getDeployments().stream().flatMap(deployment -> deployment.getBuilds().stream())
+				.collect(Collectors.toList());
+	}
+
+	@Override
+	public Collection<IResourceUIModel> getImageStreams() {
+		return this.mapper.getDeployments().stream().flatMap(deployment -> deployment.getImageStreams().stream())
```
scabanovich: This may be not thread safe. Yes, getImageStreams() returns synchronized list, but iterator is not thread safe, and java docs say that operations with iteration should be manually synchronized. The same is true for all other similar getters.

fbricon:
Closing as it's superseded by PR #1217

PR #1027, opened 2016-03-14
[JBIDE-21449] fix multiple callbacks from WatchManager

jcantrill:
Second attempt at fixing multiple callbacks.  This changes provides for a single listener per kind per project.  This will resolve issues where one listener is notified multiple times.

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
@@ -12,9 +12,10 @@ package org.jboss.tools.openshift.internal.core;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -51,7 +52,7 @@ public class WatchManager {
 			ResourceKind.ROUTE
 	};
 	
-	private Map<IProject, IWatcher> watches = Collections.synchronizedMap(new HashMap<>());
+	private Map<IProject, ConcurrentMap<String,IWatcher>> watches = new ConcurrentHashMap<>();
 	
 	private static class Holder {
 		static WatchManager instance = new WatchManager();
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -61,22 +62,26 @@ public static WatchManager getInstance() {
...
+			watchList.values().forEach(w->w.stop());
 		}
 	}
 
 	public void startWatch(final IProject project) {
-		if(!watches.containsKey(project)){
-			startWatch(project, 0, 0);
+		ConcurrentMap<String, IWatcher> kindMap = new ConcurrentHashMap<>();
+		if(watches.putIfAbsent(project, kindMap) == null){
+			final Connection conn = ConnectionsRegistryUtil.getConnectionFor(project);
```
fbricon: can't conn be null at that point?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -124,25 +132,26 @@ public void disconnected() {
...
+			Trace.warn("Reconnecting. There was an error watching connection {0}: ", err, conn.toString());
 			restart();
 		}
 		
-		private synchronized void restart() {
-			if(state.get() == State.DISCONNECTED) {
-				Trace.debug("Restart called but returning early because already disconnected");
+		private void restart() {
+			if(state.get() == State.STARTING) {
+				Trace.debug("Resturning early from restart.  Already staring for project {0} and kind {1}", project.getName(), kind);
```
fbricon: s/Resturning/Returning
s/staring/starting ???

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -124,25 +132,26 @@ public void disconnected() {
...
-		
+
+
+
 		private class RestartWatchJob extends Job{
 			private IClient client;
 
 			RestartWatchJob(IClient client){
-				super("");
+				super("WatchMananger Job");
```
fbricon: OpenShift WatchManager Job

adietish: good catch, the missing job name was resulting in quite a few unnamed jobs in the progress view - reported by QE, found out via jstack what it was

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
jcantrill: I will add a check here but if it's null then we have a problem.  It would mean that someone created a project without using our factory

jcantrill:
@fbricon @adietish updated pr

fbricon:
I confirm that on websocket reconnect, we have only 1 set of listeners (1 listener per resource kind)  per project.
So LGTM

jcantrill:
merged.

PR #1030, opened 2016-03-15
JBIDE-21868 - Performance issue on Deploy Docker Image wizard

xcoulon:
Moving the code that pull image metadata (sometimes using the
 imagestream capabilities) when the user inputs the Image Name
 to the page validation phase, when the user clicks on the
 'Next >' button.

Also renamed some fields to reflect their type and UI name:
- image -> imageName
- name -> resourceName

This also seems to fix JBIDE-21815

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
@@ -18,7 +18,6 @@ import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -30,11 +29,9 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.IJobChangeEvent;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.runtime.jobs.JobChangeAdapter;
-import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.linuxtools.docker.core.DockerConnectionManager;
 import org.eclipse.linuxtools.docker.core.IDockerConnection;
 import org.eclipse.linuxtools.docker.core.IDockerImageInfo;
-import org.eclipse.swt.widgets.Display;
 import org.jboss.dmr.ModelNode;
 import org.jboss.tools.openshift.common.core.connection.ConnectionsRegistrySingleton;
 import org.jboss.tools.openshift.core.connection.Connection;
@@ -43,7 +40,6 @@ import org.jboss.tools.openshift.internal.common.ui.wizard.IKeyValueItem;
 import org.jboss.tools.openshift.internal.ui.OpenShiftUIActivator;
 import org.jboss.tools.openshift.internal.ui.wizard.common.EnvironmentVariable;
 import org.jboss.tools.openshift.internal.ui.wizard.common.ResourceLabelsPageModel;
-import org.jboss.tools.openshift.internal.ui.wizard.deployimage.search.DockerHubRegistry;
 
 import com.openshift.restclient.OpenShiftException;
 import com.openshift.restclient.ResourceKind;
@@ -67,8 +63,8 @@ public class DeployImageWizardModel
 	private static final int DEFAULT_REPLICA_COUNT = 1;
 	private Connection connection;
 	private IProject project;
-	private String name;
-	private String image;
+	private String resourceName;
+	private String imageName;
 	private Collection<IProject> projects = Collections.emptyList();
 
 	private List<EnvironmentVariable> environmentVariables = Collections.emptyList();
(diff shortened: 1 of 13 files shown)


adietish:
@fbricon @jcantrill could you please review this since you're far more intimate with the code and usecase?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
```diff
@@ -205,90 +200,53 @@ public void setProject(IProject project) {
...
-				portSpecs.add(new PortSpecAdapter(spec));
-			}catch(IllegalArgumentException e) {
-				OpenShiftUIActivator.getDefault().getLogger().logError(e);
-			}
+	public boolean initializeContainerInfo() {
+		this.imageMeta = lookupImageMetadata();
+		if (this.imageMeta == null) {
+			return false;
 		}
+		final List<EnvironmentVariable> envVars = this.imageMeta.env().stream().map(env -> env.split("="))
```
scabanovich: Is it possible that value of variable is empty: "name="? In this case "name=".split("=").length = 1 and next map() will throw array out of index.

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/ui/models/DeployImageWizardModelTest.java
```diff
@@ -58,28 +58,28 @@ public void setUp() {
...
-		//has registry+port and tag
-		model.imageExistsLocally("host:1234/foo/bar:asf34fs");
-		verify(dockerConnection).hasImage("host:1234/foo/bar", "asf34fs");
-		
-		//
-		model.setDockerConnection(null);
-		assertFalse(model.imageExistsLocally("foo/bar"));
-		
-	}
+//	@Test
```
fbricon: why is this test commented? 
- if it's temporary, please just comment the Test annotation and add a comment explaining why
- if the code needs to be removed, then please remove it, why have git to go back in time and find previous revisions

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/IDeployImagePageModel.java
```diff
@@ -67,35 +67,55 @@
...
-	 * Checks if an image with the given name exists in a remote registry
-	 * @param imageName the full name of the image to search remotely
-	 * @return true if the image exists in a remote registry
+	 * Sets the name of the Docker Image to use
+	 * @param imageName the name of the Docker Image to use
 	 */
-	boolean imageExistsRemotely(String imageName);
+	void setImageName(String imageName);
 	
+
```
fbricon: I'd rather prefer you deleted the code instead of commenting it

fbricon:
Any chance you could add more tests pretty please?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
xcoulon: @scabanovich that's right ! Im adding `.filter(env -> env.indexOf('=') != -1)` to filter out any env variable that would not match the expected `name=value` pattern.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/IDeployImagePageModel.java
xcoulon: @fbricon I can do that

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/ui/models/DeployImageWizardModelTest.java
xcoulon: @fbricon right, it was meant to be removed because the `IDeployImagePageModel.imageExistsLocally` has been removed.

xcoulon:
@fbricon I just updated this pull request by taking all comments into account, including your kind request for JUnit tests ;-)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizardModel.java
scabanovich: @xcoulon , good that my comment helped you to find one more case to consider, but my idea was not that '=' is absent but that name=value has empty value so that literally string is 'name='. In this case split("=") returns array of length 1. 

If this case can never happen, then there is nothing to do for it, but it is good that an incorrect question may help bring to light a real problem.

fbricon:
Everything works fine for me. Applied in master / 4.3.x
Thanks @xcoulon!

xcoulon:
Great ! thanks, @fbricon ;-)

PR #1034, opened 2016-03-16
JBIDE-21616 - Main tab for CDK launch configuration is not whole by default

snjeza:
https://issues.jboss.org/browse/JBIDE-21616
Main tab for CDK launch configuration is not whole by default


fbricon:
Applied in master / 4.3.x. Thanks @snjeza !

PR #1040, opened 2016-03-17
Fix a bug in my wizard (iscomplete) and switch openshift to noop shutdown controller

robstryker:


plugins/org.jboss.tools.openshift.core/plugin.xml
```diff
@@ -120,15 +120,9 @@
          </subsystemMapping>
 
      <!-- Shutdown -->
-         <subsystem
-            system="shutdown"
-            id="shutdown.automatic.fail"
-            name="Cannot Shutdown Server"
-            class="org.jboss.tools.openshift.core.server.behavior.UnstoppableShutdownController">
```
fbricon: UnstoppableShutdownController class should be removed also

fbricon:
+1

PR #1042, opened 2016-03-18
JBIDE-21908 - legacy launch code was still in effect, didn't delegate…

robstryker:
… to launch controller


fbricon:
Works alright.
+1 to apply

PR #1043, opened 2016-03-18
JBIDE-21905 - make startup controller a bit more logical with a quick…

robstryker:
… poll to see if the connection can even be reached

Update test


fbricon:
+1 to apply

PR #1045, opened 2016-03-18
JBIDE-21901 - ensure new server adapter gets started after creation, …

robstryker:
…which should always work since the wizards connect to the service

Add suppression of automatic startup for tests

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizard.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizard.java
@@ -17,6 +17,7 @@ import org.eclipse.jface.wizard.Wizard;
 import org.eclipse.wst.server.core.IServer;
 import org.eclipse.wst.server.core.IServerWorkingCopy;
 import org.jboss.tools.openshift.core.connection.Connection;
+import org.jboss.tools.openshift.core.server.OpenShiftServerUtils;
 import org.jboss.tools.openshift.internal.ui.OpenShiftUIActivator;
 
 import com.openshift.restclient.model.IService;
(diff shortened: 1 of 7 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizard.java
```diff
@@ -59,7 +60,9 @@ public void addPages() {
 	public boolean performFinish() {
 		this.serverSettingsWizardPage.getModel().updateServer();
 		try {
-			this.createdServer = this.serverSettingsWizardPage.getModel().getServer().save(true, new NullProgressMonitor());
+			IServerWorkingCopy wc = this.serverSettingsWizardPage.getModel().getServer();
+			wc.setAttribute(OpenShiftServerUtils.SERVER_START_ON_CREATION, true);
```
adietish: I see this being made in the wizard and the page with twice the same code. Why not do it upon saving the server in the model?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardFragment.java
```diff
@@ -106,6 +107,7 @@ public boolean hasComposite() {
 	public void performFinish(IProgressMonitor monitor) throws CoreException {
 		if(serverSettingsWizardPage != null) {
 			serverSettingsWizardPage.getModel().updateServer();
+			serverSettingsWizardPage.getModel().getServer().setAttribute(OpenShiftServerUtils.SERVER_START_ON_CREATION, true);
```
adietish: I see this being made in the wizard and the page with twice the same code. Why not do it upon saving the server in the model?

robstryker: The goal here was to make it so that not having the attribute set would not start the server. If I simply added this to the model or the OpenShiftServerUtils,  then our test cases would not be able to use these methods or models without then overwriting the value to false. The goal is of course to test what the behavior is when the flag isn't set at all, and the way to do that is to ensure the model itself isn't setting the value.

PR #1049, opened 2016-03-21
[JBIDE-21920] ensure the oc binary is set when necessary

fbricon:


adietish:
+1, but it seems to me that we now dont restore the location in the context after we ran the oc binary.

fbricon:
@adietish please look at the new iteration

adietish:
ok, great! LGTM

PR #1050, opened 2016-03-21
[JBIDE-21844] fix DeployImageJob to create dc that points to imagestream

jcantrill:
cc @fbricon @adietish 

This fixes the issue where some images deploy and others do not.  It is related to how we are creating the indirection from the imagechangetrigger to the imagestream and the resource name length.  The trigger#from is of the form imagestreamname.tag and is not necessarily equivalent to dockerimagename.tag.


fbricon:
works like a charm
+1

fbricon:
Applied in master / 4.3.x

PR #1051, opened 2016-03-22
Jbide 21934

robstryker:


plugins/org.jboss.tools.openshift.express.ui/src/org/jboss/tools/openshift/express/internal/ui/server/CommitDialog.java
```diff
@@ -1299,7 +1299,7 @@ private static Status getFileStatus(String path, IndexDiff indexDiff) {
 	}
```
fbricon: these changes look very suspicious

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ServerRSyncUtil.java
```diff
@@ -0,0 +1,49 @@
...
+import org.eclipse.wst.server.core.IServer;
+import org.jboss.tools.openshift.common.core.utils.StringUtils;
+import org.jboss.tools.openshift.core.server.OpenShiftServerUtils;
+import org.jboss.tools.openshift.core.server.RSync;
+import org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator;
+import org.jboss.tools.openshift.internal.core.preferences.OCBinary;
+
+import com.openshift.restclient.model.IService;
+
+public class ServerRSyncUtil {
```
fbricon: looks like a good candidate for some testing

robstryker: Seems pretty difficult... it'd require having actual projects in the workspace.  It all comes back to getDeployProject where a number of settings are stored..., so unless I can mock the entire workspace for the call to locate the project, I don't think any of this will work.  

OpenShiftServerUtils is basically one huge blob of untestable static methods.

PR #1052, opened 2016-03-22
[JBIDE-21829] fix retryList which was not getting namespace

jcantrill:
cc @fbricon 

This fixes the issue where upon retrying to list a set of resource we did not send the namespace


fbricon:
+1 nice catch

jcantrill:
merged into 4.3.x and master

PR #1053, opened 2016-03-22
JBIDE-21326 - Server Adapter: Publishing is not telling me what it's doing

xcoulon:
Writing the lines returned by the `rsync` operation in the client in the console using an IServerConsoleWriter.

This patch is a joined effort with @robstryker and requires that the work on branch https://github.com/robstryker/jbosstools-server/tree/JBIDE-21326_maint on `jbosstools-server` be merged, first.

The rsync logs are streamed in the console during both `publishStart` and `publishFinish` method invocations, but the later could probably be avoided.

@fbricon @adietish @jcantrill please let me know if this patch works for you.


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
```diff
@@ -42,22 +39,20 @@
 
 public class OpenShiftPublishController extends StandardFileSystemPublishController implements IPublishController {
 
-	private RSync rsync = null;
-	private RSync createRSync(IServer server, IProgressMonitor monitor) throws CoreException {
-		String location = OCBinary.getInstance().getLocation();
+	private static RSync createRSync(IServer server, IProgressMonitor monitor) throws CoreException {
+		final String location = OCBinary.getInstance().getLocation();
```
adietish: I feel like this should be checked in RSync (or even better in any OCBinaryOperation) so that we dont have to repeat ourselves.

adietish:
Other than this +1, looks good and removes a deprecated class

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
xcoulon: ok, let me refactor this code

xcoulon:
@adietish patch updated after your review ;-)

PR #1054, opened 2016-03-23
[JBIDE-21948] fix ClassCastException for SourceBuildStrategy

jcantrill:
@fbricon fixes class cast exception introduced by rest client changes.


fbricon:
+1

PR #1057, opened 2016-03-24
[JBIDE-21625] Enable debug mode for OpenShift 3 server adapter

fbricon:

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
@@ -40,6 +40,7 @@ import org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator;
 import org.osgi.service.prefs.BackingStoreException;
 
 import com.openshift.restclient.ResourceKind;
+import com.openshift.restclient.model.IDeploymentConfig;
 import com.openshift.restclient.model.IService;
 import com.openshift.restclient.model.route.IRoute;
 
@@ -47,6 +48,8 @@ import com.openshift.restclient.model.route.IRoute;
  * @author Andre Dietisheim
  */
 public class OpenShiftServerUtils {
+
+	private static final String DEPLOYMENT_CONFIG_KEY = "deploymentconfig";
 
 	private static final String LIVERELOAD_PORT_KEY = "port";//Key to the port # of the host the LiveReload server need to proxy
 
@@ -326,5 +329,26 @@ public class OpenShiftServerUtils {
 	 */
 	public static String getProjectAttribute(String name, String defaultValue, IProject project) {
 		return ServerUtils.getProjectAttribute(name, defaultValue, SERVER_PROJECT_QUALIFIER, project);
+	}
+	
+	public static IDeploymentConfig getDeploymentConfig(IServerAttributes attributes) {
+		IService service = getService(attributes);
+		if (service == null) {
+			return null;
+		}
+		//TODO use annotations instead of labels
+		String dcName = service.getPods().stream().filter(p -> p.getLabels().containsKey(DEPLOYMENT_CONFIG_KEY))
+				.findFirst()
+				.map(p -> p.getLabels().get(DEPLOYMENT_CONFIG_KEY))
+				.orElse(null);
+		if (dcName == null) {
+			return null;
+		}
+		Connection connection = getConnection(attributes);
+		if (connection == null) {
+			return null;
+		}
+		IDeploymentConfig dc = connection.getResource(ResourceKind.DEPLOYMENT_CONFIG, service.getNamespace(), dcName);
+		return dc;
 	}
 }
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
@@ -10,13 +10,21 @@
  *******************************************************************************/
 package org.jboss.tools.openshift.core.server.behavior;
 
+import java.io.IOException;
+import java.util.Optional;
+import java.util.Set;
+
+import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.debug.core.ILaunch;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.launching.SocketUtil;
 import org.eclipse.wst.server.core.IServer;
 import org.eclipse.wst.server.core.ServerUtil;
 import org.jboss.ide.eclipse.as.wtp.core.server.behavior.AbstractSubsystemController;
@@ -24,15 +32,25 @@ import org.jboss.ide.eclipse.as.wtp.core.server.behavior.ControllableServerBehav
 import org.jboss.ide.eclipse.as.wtp.core.server.behavior.IControllableServerBehavior;
 import org.jboss.ide.eclipse.as.wtp.core.server.behavior.ILaunchServerController;
 import org.jboss.ide.eclipse.as.wtp.core.server.behavior.ISubsystemController;
+import org.jboss.tools.foundation.core.plugin.log.StatusFactory;
 import org.jboss.tools.openshift.core.server.OpenShiftServerUtils;
 import org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator;
+import org.jboss.tools.openshift.internal.core.portforwarding.PortForwardingUtils;
+import org.jboss.tools.openshift.internal.core.server.debug.DebuggingContext;
+import org.jboss.tools.openshift.internal.core.server.debug.IDebugListener;
+import org.jboss.tools.openshift.internal.core.server.debug.OpenShiftDebugUtils;
 
 import com.openshift.restclient.OpenShiftException;
+import com.openshift.restclient.capability.resources.IPortForwardable;
+import com.openshift.restclient.capability.resources.IPortForwardable.PortPair;
+import com.openshift.restclient.model.IDeploymentConfig;
+import com.openshift.restclient.model.IPod;
 import com.openshift.restclient.model.IService;
 
 public class OpenShiftLaunchController extends AbstractSubsystemController
 		implements ISubsystemController, ILaunchServerController {
 
+	private static final String DEBUG_MODE = "debug";
 
 	/**
 	 * Get access to the ControllableServerBehavior
@@ -50,20 +68,87 @@ public class OpenShiftLaunchController extends AbstractSubsystemController
 	@Override
 	public void launch(ILaunchConfiguration configuration, String mode,
 			ILaunch launch, IProgressMonitor monitor) throws CoreException {
-		IControllableServerBehavior beh = getServerBehavior(configuration);
-		if( beh != null ) {
-			((ControllableServerBehavior)beh).setServerStarting();
-			int state = pollState();
-			if( state == IServer.STATE_STARTED) {
-				((ControllableServerBehavior)beh).setServerStarted();
-				((ControllableServerBehavior)getControllableBehavior()).setRunMode(mode);
-			} else {
-				((ControllableServerBehavior)beh).setServerStopped();
-				((ControllableServerBehavior)getControllableBehavior()).setRunMode(null);
-			}
-		} else {
-			// TODO throw error
+		IControllableServerBehavior serverBehavior = getServerBehavior(configuration);
+		if( !(serverBehavior instanceof ControllableServerBehavior )) {
+			throw toCoreException("Unable to find a IControllableServerBehavior instance");
 		}
+		ControllableServerBehavior beh = (ControllableServerBehavior) serverBehavior;
+		
+		beh.setServerStarting();
+		
+		IServer server = beh.getServer();
+		
+		IDeploymentConfig dc = OpenShiftServerUtils.getDeploymentConfig(server);
+		if (dc == null) {
+			throw toCoreException("No deployment config was found for "+server.getName());
+		}
+		
+		DebuggingContext debugContext = OpenShiftDebugUtils.get().getDebuggingContext(dc);
+		if( DEBUG_MODE.equals(mode)) {
+			startDebugging(server, dc, debugContext, monitor);
+		} else {//run, profile
+			stopDebugging(dc, debugContext, monitor);
+		}
+		
+		int state = pollState();
+		if( state == IServer.STATE_STARTED) {
+			beh.setServerStarted();
+			beh.setRunMode(mode);
+		} else {
+			beh.setServerStopped();
+			((ControllableServerBehavior)getControllableBehavior()).setRunMode(null);
+		}
+	}
+
+
+	private void startDebugging(IServer server, IDeploymentConfig dc, DebuggingContext debugContext,
+			IProgressMonitor monitor) throws CoreException {
+		int remotePort = debugContext.getDebugPort();
+		if( remotePort == -1 ) {
+			debugContext.setDebugPort(8787);//TODO get default port from server settings?
+		}
+		IDebugListener listener = new IDebugListener() {
+			
+			public void onDebugChange(DebuggingContext debuggingContext, IProgressMonitor monitor)
+					throws CoreException {
+				if (debuggingContext.getPod() == null) {
+					throw toCoreException("Unable to connect to remote pod");
+				}
+				int localPort = mapPortForwarding(debuggingContext, monitor);
+				
+				if(isJavaProject(server)) {
+					attachRemoteDebugger(server, localPort, monitor);
+				}					
+			}
+
+			@Override
+			public void onPodRestart(DebuggingContext debuggingContext, IProgressMonitor monitor)
+					throws CoreException {
+				onDebugChange(debuggingContext, monitor);
+			}
+		};
+		debugContext.setDebugListener(listener);
+		OpenShiftDebugUtils.get().enableDebugMode(dc, debugContext, monitor);
+	}
+
+
+	private void stopDebugging(IDeploymentConfig dc, DebuggingContext debugContext, IProgressMonitor monitor)
+			throws CoreException {
+		IDebugListener listener = new IDebugListener() {
+			
+			public void onDebugChange(DebuggingContext debuggingContext, IProgressMonitor monitor)
+					throws CoreException {
+				//that will automatically kill the remote debugger
+				unMapPortForwarding(debuggingContext.getPod());
+			}
+
+			@Override
+			public void onPodRestart(DebuggingContext debuggingContext, IProgressMonitor monitor)
+					throws CoreException {
+			}
+		};
+		debugContext.setDebugListener(listener);
+		OpenShiftDebugUtils.get().disableDebugMode(dc, debugContext, monitor);
 	}
 
 	protected int pollState() {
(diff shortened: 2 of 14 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
```diff
@@ -327,4 +332,26 @@ public static String getProjectAttribute(String name, String defaultValue, IServ
...
 		return ServerUtils.getProjectAttribute(name, defaultValue, SERVER_PROJECT_QUALIFIER, project);
 	}
+	
+	public static IDeploymentConfig getDeploymentConfig(IServerAttributes attributes) {
+		String dcKey = "deploymentconfig";
+		IService service = getService(attributes);
+		if (service == null) {
+			return null;
+		}
+		//TODO use annotations instead of labels
```
jcantrill: Are you able to retrieve this value from the annoations and move the constant to OpenShiftAPIAnnotations

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -50,20 +67,83 @@ protected static IControllableServerBehavior getServerBehavior(ILaunchConfigurat
...
+			throw toCoreException("Unable to find a IControllableServerBehavior instance");
+		}
+		ControllableServerBehavior beh = (ControllableServerBehavior) serverBehavior;
+		
+		beh.setServerStarting();
+		
+		IServer server = beh.getServer();
+		
+		IDeploymentConfig dc = OpenShiftServerUtils.getDeploymentConfig(server);
+		DebuggingContext debugContext = OpenShiftDebugUtils.get().getDebuggingContext(dc);
```
jcantrill: Don't we need a null check here in case the DC is not found

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -50,20 +67,83 @@ protected static IControllableServerBehavior getServerBehavior(ILaunchConfigurat
...
+		if( "debug".equals(mode)) {
+			startDebugging(server, dc, debugContext, monitor);
+		} else {//run, profile
+			stopDebugging(dc, debugContext, monitor);
+		}
+		
+		int state = pollState();
+		if( state == IServer.STATE_STARTED) {
```
jcantrill: Is there a need to synchronize on something here since we are evaluating state and then doing something?

fbricon: that piece is @robstryker 's code, I believe he can answer that.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/DeploymentConfigListenerJob.java
```diff
@@ -0,0 +1,101 @@
...
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.Job;
+import org.jboss.tools.openshift.common.core.connection.IConnection;
+import org.jboss.tools.openshift.common.core.connection.IConnectionsRegistryListener;
+import org.jboss.tools.openshift.internal.core.util.ResourceUtils;
+
+import com.openshift.restclient.model.IDeploymentConfig;
+import com.openshift.restclient.model.IPod;
+
+public class DeploymentConfigListenerJob extends Job implements IConnectionsRegistryListener {
```
jcantrill: You could add an inner instance of ConnectionsRegistryAdapter if you wanted to remove those empty methods

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
```diff
@@ -327,4 +332,26 @@ public static String getProjectAttribute(String name, String defaultValue, IServ
 	public static String getProjectAttribute(String name, String defaultValue, IProject project) {
 		return ServerUtils.getProjectAttribute(name, defaultValue, SERVER_PROJECT_QUALIFIER, project);
 	}
+	
+	public static IDeploymentConfig getDeploymentConfig(IServerAttributes attributes) {
+		String dcKey = "deploymentconfig";
```
xcoulon: make a constant ?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
xcoulon: no, the `public DebuggingContext getDebuggingContext(IDeploymentConfig)` already checks if the argument is `null`.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -50,20 +67,83 @@ protected static IControllableServerBehavior getServerBehavior(ILaunchConfigurat
...
+		}
+		ControllableServerBehavior beh = (ControllableServerBehavior) serverBehavior;
+		
+		beh.setServerStarting();
+		
+		IServer server = beh.getServer();
+		
+		IDeploymentConfig dc = OpenShiftServerUtils.getDeploymentConfig(server);
+		DebuggingContext debugContext = OpenShiftDebugUtils.get().getDebuggingContext(dc);
+		if( "debug".equals(mode)) {
```
xcoulon: use a constant ?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -50,20 +67,83 @@ protected static IControllableServerBehavior getServerBehavior(ILaunchConfigurat
...
+			IProgressMonitor monitor) throws CoreException {
+		int remotePort = debugContext.getDebugPort();
+		if( remotePort == -1 ) {
+			debugContext.setDebugPort(8787);//TODO get default port from server settings?
 		}
+		IDebugListener listener = new IDebugListener() {
+			
+			public void onDebugChange(DebuggingContext debuggingContext, IProgressMonitor monitor)
+					throws CoreException {
+				if (debuggingContext.getPod() == null) {
```
xcoulon: can `debuggingContext` be null ?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -92,5 +172,131 @@ public void setupLaunchConfiguration(
...
+		}
+	}
+	
+	/**
+	 * Map the remote port to a local port. 
+	 * Return the local port in use, or -1 if failed
+	 * @param server
+	 * @param remotePort
+	 * @return the local debug port
+	 */
```
xcoulon: .. or `-1` if port forwarding did not start or was cancelled.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -92,5 +172,131 @@ public void setupLaunchConfiguration(
...
+		
+		long elapsed = 0;
+		long increment = 1_000;
+		long maxWait =  60_000; //TODO Get from server settings?
+		boolean launched = false;
+		monitor.subTask("Waiting for remote debug port availability");
+		while(!launched && elapsed < maxWait) {
+			try {
+				//TODO That's fugly. ideally we should see if socket on debug port is responsive instead
+				debuggerLaunchConfig.launch("debug", new NullProgressMonitor());
```
xcoulon: same as above: use a constant for 'debug'

xcoulon: are you launching the debugger multiple times until it works ?

fbricon: yup

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/DeploymentConfigListenerJob.java
```diff
@@ -0,0 +1,101 @@
...
+			}
+			return;
+		}
+		//Wait for new pod once deployment is done
+		if (newValue instanceof IPod) {
+			IPod candidate = (IPod) newValue;
+			String podName = candidate.getName();
+			if (!oldPods.contains(podName) &&
+				"Running".equals(candidate.getStatus()) &&
+				!podName.endsWith("-deploy")) {
```
jcantrill: A more exact way of matching the pod would be to check if the selector from the DC would select the pod.  The DC has a spec that is used to create the replicationcontroller which monitors pods to ensure the replica count reaches the desired state.  AFAIK 'deploy' pods do not have the dc selector labels.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
xcoulon: However, `OpenShiftDebugUtils.get().getDebuggingContext(IDeploymentConfig)` can return `null`, which means that a `null` `debugContext` will be passed to the `startDebugging` and `stopDebugging` methods, which will cause NPEs

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -0,0 +1,260 @@
...
+			updateDebugConfig(deploymentConfig, debugContext, monitor);
+		}
+		return debugContext;
+	}
+
+	public DebuggingContext disableDebugMode(IDeploymentConfig deploymentConfig, DebuggingContext debugContext, IProgressMonitor monitor) throws CoreException {
+		assert deploymentConfig != null;
+		assert debugContext != null;
+
+		IDebugListener listener = debugContext.getDebugListener();
```
jcantrill: There is some (mostely) repeated code here between enable and disable debug mode

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -0,0 +1,260 @@
...
+			debugContext.setDebugEnabled(false);
+			updateDebugConfig(deploymentConfig, debugContext, monitor);
+		}
+		return debugContext;
+	}
+	
+	public void updateDebugConfig(IDeploymentConfig deploymentConfig, DebuggingContext debugContext, IProgressMonitor monitor) throws CoreException {
+		monitor.subTask("Updating Deployment Configuration");
+		if (deploymentConfig == null
+				|| deploymentConfig.getEnvironmentVariables() == null) {
```
jcantrill: I dont think this will ever return null for ENV vars, but its worth keeping the check.  What about a check or empty too?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -50,20 +67,83 @@ protected static IControllableServerBehavior getServerBehavior(ILaunchConfigurat
...
+
+
+	private void stopDebugging(IDeploymentConfig dc, DebuggingContext debugContext, IProgressMonitor monitor)
+			throws CoreException {
+		IDebugListener listener = new IDebugListener() {
+			
+			public void onDebugChange(DebuggingContext debuggingContext, IProgressMonitor monitor)
+					throws CoreException {
+				//that will automatically kill the remote debugger
+				unMapPortForwarding(debuggingContext.getPod());
```
xcoulon: potential NPE

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -50,20 +67,83 @@ protected static IControllableServerBehavior getServerBehavior(ILaunchConfigurat
...
+			}
+
+			@Override
+			public void onPodRestart(DebuggingContext debuggingContext, IProgressMonitor monitor)
+					throws CoreException {
+			}
+		};
+		debugContext.setDebugListener(listener);
```
xcoulon: potential NPE (see above)

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -0,0 +1,260 @@
...
+					added = ports.add(newPort);
+				}
+			}
+			if (added) {
+				container.setPorts(ports); 
+				deploymentConfig.setContainers(containers);
+			}
+		}
+		//TODO the list of env var to set in debug mode should probably be defined in the server settings instead
+		deploymentConfig.setEnvironmentVariable("DEBUG_PORT", String.valueOf(debugContext.getDebugPort()));
```
jcantrill: Atleast move them to constants in this class

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
xcoulon: should we call the `ILaunch#terminate()` method instead ? 
The `ILaunch` can be provided by the calling `launch(ILaunchConfiguration, String, ILaunch, IProgressMonitor )` method.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
```diff
@@ -194,4 +196,19 @@ public static String imageRef(IBuild build) {
...
 
 	}
+	
+	/**
+	 * Find the collection of pods for the given deployment config
+	 * @param deploymentConfig
+	 * @param pods
+	 * @return
+	 */
+	public static Collection<IPod> getPodsForDeploymentConfig(IDeploymentConfig deploymentConfig) {
```
jcantrill: Similiar to my comment above is this will match only if the annotation is there but a better approach would probably be to try and match the dc.selector to the pod.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
```diff
@@ -194,4 +196,19 @@ public static String imageRef(IBuild build) {
...
+		String dcName = deploymentConfig.getName();
+		List<IPod> projectPods = deploymentConfig.getProject().getResources(ResourceKind.POD);
+		List<IPod> deploymentConfigPods = projectPods.stream()
+				.filter(p -> dcName.equals(p.getAnnotation("openshift.io/deployment-config.name")))
```
jcantrill: or use the constant from the OpenShiftAnnotations

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -0,0 +1,260 @@
...
+	public static OpenShiftDebugUtils get(ILaunchManager launchManager) {
+		return new OpenShiftDebugUtils(launchManager);
+	}
+	
+	private OpenShiftDebugUtils(ILaunchManager launchManager) {
+		this.launchManager = launchManager;
+	}
+	
+	public DebuggingContext enableDebugMode(IDeploymentConfig deploymentConfig, DebuggingContext debugContext, IProgressMonitor monitor) throws CoreException {
+		assert deploymentConfig != null;
```
xcoulon: Shouldn't we avoid the `asset` keyword ? Are we sure that Eclipse is started with the `-ea` switch ?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -0,0 +1,260 @@
...
+		assert debugContext != null;
+
+		IDebugListener listener = debugContext.getDebugListener();
+		if (debugContext.isDebugEnabled() && listener != null) {
+			IPod pod = getFirstPod(deploymentConfig);
+			debugContext.setPod(pod);
```
xcoulon: what happens if the pod is null ? potential NPE somewhere else ?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
robstryker: No need to synchronize anything here.  It's our job to set the server state here. The IServer is waiting for us to beh.setServerStarted() on success or beh.setServerStopped() on a startup-fail.   Since our initial poll indicates it's already up, we set it to started.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
fbricon: no

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
fbricon: true, but the behavior / flow is slightly different, so not obvious to refactor. This is something I can look at post 4.3.x

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
fbricon: no, current ENV vars can be empty and we want to add new stuff in it

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
fbricon: debuggingContext can't be null here

fbricon: we can certainly improve the detection mechanism. Right now I'd like to merge it as is, so people can use it

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
fbricon: no debugContext is guaranteed to be not null

PR #1059, opened 2016-03-24
JBIDE-21949 - Allow user to turn on/off showing console when output changes

xcoulon:
Using the new method in IServerConsoleWriter to avoid activating the console
when new content is written. The console is only opened when it is created,
ie, at the first 'rsync' operation.


fbricon:
+1 to merge

xcoulon:
merging now, since the required patch on `jbosstools-server` was merged, too: https://github.com/jbosstools/jbosstools-server/pull/405

PR #1064, opened 2016-03-29
JBIDE-21981 - Error during rsync

xcoulon:
Using the OpenShiftBinaryOptions to pass some options to the 'oc'
command that will be run by the client, to make sure that the '.git'
folder is still excluded during rsync and that the server certificates
are not verified during the TLS connection setup.


xcoulon:
This PR depends on https://github.com/openshift/openshift-restclient-java/pull/138 which brings changes at the client level.

xcoulon:
Closing an opening with a different JIRA issue number.

PR #1065, opened 2016-03-29
[JBIDE-21902] fix saving token not needing to remember it

jcantrill:
Fix setting token to null on save when we can still us it.


adietish:
thanks!

PR #1070, opened 2016-03-30
JBIDE-21942 - prevent NPE on adbinfo

robstryker:
Also tries to fix some workflow issues when openshift cannot be reached after cdk is marked started by opening a dialog to let the user know the situation

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/VagrantPoller.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/VagrantPoller.java
@@ -12,6 +12,10 @@ package org.jboss.tools.openshift.cdk.server.core.internal.adapter;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.SocketTimeoutException;
+import java.net.URL;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
@@ -20,6 +24,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.TimeoutException;
+
+import javax.net.ssl.HttpsURLConnection;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IStatus;
@@ -32,6 +38,13 @@ import org.jboss.tools.openshift.cdk.server.core.internal.CDKConstants;
 import org.jboss.tools.openshift.cdk.server.core.internal.CDKCoreActivator;
 import org.jboss.tools.openshift.cdk.server.core.internal.adapter.controllers.VagrantLaunchUtility;
 import org.jboss.tools.openshift.cdk.server.core.internal.listeners.CDKLaunchEnvironmentUtil;
+import org.jboss.tools.openshift.cdk.server.core.internal.listeners.ServiceManagerEnvironment;
+import org.jboss.tools.openshift.core.OpenShiftCoreUIIntegration;
+
+import com.openshift.internal.restclient.http.HttpClientException;
+import com.openshift.internal.restclient.http.UrlConnectionHttpClientBuilder;
+import com.openshift.restclient.ISSLCertificateCallback;
+import com.openshift.restclient.http.IHttpClient;
 
 public class VagrantPoller implements IServerStatePoller2 {
 	private IServer server;
@@ -59,7 +72,7 @@ public class VagrantPoller implements IServerStatePoller2 {
 			public void run() {
 				pollerRun();
 			}
-		}, "Vagrant Poller"); //$NON-NLS-1$
+		}, "CDK Poller"); //$NON-NLS-1$
 		t.start();
 	}
 	
@@ -73,7 +86,7 @@ public class VagrantPoller implements IServerStatePoller2 {
 		setStateInternal(false, state);
     	Map<String,String> env = CDKLaunchEnvironmentUtil.createEnvironment(server);
 		while(aborted == null && !canceled && !done) {
-			IStatus stat = onePing(server, env);
+			IStatus stat = onePingSafe(server, env);
 			int status = stat.getSeverity();
 			boolean completeUp = ( status == IStatus.OK && expectedState);
 			boolean completeDown = (status == IStatus.ERROR && !expectedState);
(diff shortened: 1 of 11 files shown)


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ServiceManagerEnvironment.java
```diff
@@ -78,13 +119,20 @@ public static ADBInfo loadADBInfo(IServer server) {
...
 		} catch( URISyntaxException urise) {
 			CDKCoreActivator.pluginLog().logError("Environment variable DOCKER_HOST is not a valid uri:  " + env.get("DOCKER_HOST"), urise);
 		} catch(IOException | TimeoutException ce) {
-			CDKCoreActivator.pluginLog().logError("Unable to successfully complete a call to vagrant adbinfo. ", ce);
+			CDKCoreActivator.pluginLog().logError("Unable to successfully complete a call to vagrant service-manager. ", ce);
 		}
-		return null;
+		return null;	}
+	
+	@Deprecated
```
fbricon: why don't you remove it completely?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ServiceManagerEnvironment.java
```diff
@@ -17,25 +17,24 @@
...
 	
-	public static final String SHARED_INFO_KEY = "cdk.sharedinfo.adbInfo";
+	public static final String SHARED_INFO_KEY = "cdk.sharedinfo.serviceManagerEnvironment";
 	
 	int openshiftPort = 8443;
 	String openshiftHost = "https://10.1.2.2";
 	
 	HashMap<String,String> env;
-	public ADBInfo(HashMap<String,String> env) throws URISyntaxException {
+	public ServiceManagerEnvironment(HashMap<String,String> env) throws URISyntaxException {
```
fbricon: please use a Map interface instead

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
```diff
@@ -227,24 +230,47 @@ private void processTerminated(IServer server,IProcess process, IDebugEventSetLi
...
+			} else {
+				beh.setServerStopped();
+			}
+		}
+	}
+	
+	private void handleOpenShiftUnavailable(final IControllableServerBehavior beh, final OpenShiftNotReadyPollingException osnrpe) {
+		// Log error?  Show dialog?  
+		((ControllableServerBehavior)beh).setServerStarted();
+		new Job(osnrpe.getMessage()) {
```
fbricon: why a job to log that message?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
```diff
@@ -227,24 +230,47 @@ private void processTerminated(IServer server,IProcess process, IDebugEventSetLi
 		final ControllableServerBehavior beh = (ControllableServerBehavior)JBossServerBehaviorUtils.getControllableBehavior(server);
 		new Thread() {
 			public void run() {
-				try {
```
fbricon: handleProcessTerminated is launched from a thread, but spawns a new thread itself

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/VagrantPoller.java
```diff
@@ -132,32 +135,77 @@ private File getWorkingDirectory(IServer s) throws PollingException {
...
-		return CDKCoreActivator.statusFactory().infoStatus(CDKCoreActivator.PLUGIN_ID, "Vagrant status indicates the CDK is starting.");
+    	String[] lines = VagrantLaunchUtility.call(vagrantcmdloc, args,  getWorkingDirectory(server), env);
+		IStatus vmStatus = parseOutput(lines);
+		if (vmStatus.isOK()) {
+			checkOpenShiftHealth(server, 4000); // throws OpenShiftNotReadyPollingException
+		}
+		return vmStatus;
+	}
+	
+	private boolean checkOpenShiftHealth(IServer server, int timeout) throws OpenShiftNotReadyPollingException {
```
fbricon: I'd make that method take a health url as parameter, so it can be ~unit tested

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
robstryker: It doesn't log the message.  It pops it up in an error dialog. I did that to avoid having to make a UI extension point, or somehow mixing my core / ui code together.

robstryker:
I'm not able to verify this patch works. Checking the health of https://10.1.2.2:8443/healthz/ready fails with a sockettimeoutexception.  From curl, however, it fails with an issue about the certificate being unsigned.  I suspect these two are related. 

If I switch the url to https://www.google.com, it works as expected with no socket timeout exception.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
fbricon: I see. interesting approach.
You smarty pants ;-)

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/VagrantPoller.java
```diff
@@ -132,32 +143,84 @@ private File getWorkingDirectory(IServer s) throws PollingException {
...
+		
+		ServiceManagerEnvironment adb = ServiceManagerEnvironment.getOrLoadServiceManagerEnvironment(server, true);
+		if( adb == null ) {
+			return false;
+		}
+		String url = adb.getOpenShiftHost() + ":" + adb.getOpenShiftPort() + "/healthz/ready";
+		return checkOpenShiftHealth(url, timeout);
+	}
+	protected boolean checkOpenShiftHealth(String url,  int timeout) throws OpenShiftNotReadyPollingException {
+		ISSLCertificateCallback sslCallback = OpenShiftCoreUIIntegration.getInstance().getSSLCertificateCallback();
```
fbricon: sslCallback could be a method param, to ease unit testing

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDKLaunchController.java
scabanovich: I would think that it both shows the message in the error dialog _and_ logs it. 
If it does not log it, I would see it both as a bug in Eclipse and at the same time the most ingenious way to call UI from core without cumbersome callback interfaces.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/CDKDockerUtility.java
```diff
@@ -54,7 +54,7 @@ public String getNextName(IServer server) {
 		return name;
 	}
 	
-	public IDockerConnection findDockerConnection(ADBInfo adb) {
+	public IDockerConnection findDockerConnection(ServiceManagerEnvironment adb) {
```
scabanovich: Just a trifle remark. ADBInfo is renamed into ServiceManagerEnvironment but (here and in other places) variable name is 'adb' which will put before future generations a mystery to guess over: 'why adb'? :)

fbricon:
So I tested it works, that's pretty cool. Certificate dialog shows up earlier than before because of the health check, but that's perfectly acceptable.

I also tested that when the url is unreachable (by injecting a different server url in debug mode), an error pops up and is the error is also logged.

I wish we could use something else than an internal openshift-restclient package, but that'll do it for this release.

+1 to apply after squashing the commit. Nice job @robstryker

PR #1071, opened 2016-03-31
[JBIDE-21984] server adapter wizard: set project/service according to invocation

adietish:
setting the project and the service in a different manner whether the wizard
was started from explorer or "new server adapter"-wizard


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServiceViewModel.java
```diff
@@ -87,9 +87,9 @@ private void updateServiceItems(List<ObservableTreeItem> newServiceItems) {
 	}
 
 	protected IService updateService(final IService service, final List<ObservableTreeItem> serviceItems) {
-		if (!isLoaded) {
-			return service;
-		}
+//		if (!isLoaded) {
```
fbricon: commenting these lines breaks bootstrapping the service in the wizard.
If you have 2 services in a project, select the 2nd service, click server adapter, then the service is set to null on the 1st call to this method, because other serviceItems are not loaded and later the service will be initialized to the 1st service

adietish: +1

PR #1073, opened 2016-03-31
[JBIDE-22033] Terminate remote debugger when stopping the server

fbricon:
also fixed server status in case an error occurs during startup


fbricon:
@robstryker do you mind taking a look?

robstryker:
I can't find anything to nitpick here at all. It's almost as if you pre-nitpicked your own PR.... kinda unfair... like cheating.  

+1

PR #1074, opened 2016-03-31
JBIDE-21995 - always set the route if it's available

robstryker:
I'm not 100% convinced this is the right solution... 

However, I do know trying to use only a 'default route' means that we get no route url, which means we get no server host. 

Furthermore, the existing code was already wrong. The existing code returned null if the model did NOT select 'default route'.  Something was very strange here. 

I can understand if no route is passed in, we can assume it's the default route to be determined later. However a route is being passed in, and the server ceases to function with no route selected. So if a route is provided, I'm definitely going to use its URL for server host.

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPageModel.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPageModel.java
@@ -362,9 +362,9 @@ public class ServerSettingsWizardPageModel extends ServiceViewModel {
 	}
 
 	private String getRoute(boolean isDefaultRoute, IRoute route) {
-		if (!isDefaultRoute || route == null) {
+		// We must return a route here, because setting the server host depends on the route
+		if( route == null )
 			return null;
-		}
 		return route.getURL();
 	}
 


robstryker:
It's worth noting that this is really only a partial fix. If no route is passed in,  the host will remain localhost. I couldn't quite figure out where to change or resolve the 'default route' in this block of code. I'm assuming that the routes were resolved in the background and so the passed in route is actually the default already.  This is probably a bad assumption, but it's the best I can do at this time.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPageModel.java
```diff
@@ -362,9 +362,9 @@ private String getProfileId() {
 	}
 
 	private String getRoute(boolean isDefaultRoute, IRoute route) {
-		if (!isDefaultRoute || route == null) {
+		// We must return a route here, because setting the server host depends on the route
```
scabanovich: This just removes the feature, which allows user to select route later at starting Show In Browser. Does server adapter logic allow for such a feature? If it does, then the solution may be just to fill host with empty string or a message reminding user of his choice to choose always (or until he selects 'remember' option) the route in a dialog.

adietish:
as @scabanovich pointed out this suggested solution would remove the feature where we want the user to be prompted when the user in UI checks "Prompt for route when multiple routes available to show in browser". 
So imho this fix wont solve the issue properly.

adietish:
@robstryker the logic is correct, isDefaultRoute is "true" if a default route was chosen, or in other words if the checkbox is unchecked.

adietish:
I'd do the following to keep the wizard intact: https://github.com/jbosstools/jbosstools-openshift/pull/1077

adietish:
replaced by #1077

PR #1077, opened 2016-03-31
[JBIDE-21995] null-routeUrl->"" to get non-localhost host in server a…

adietish:
…dapter

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerExtendedProperties.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerExtendedProperties.java
@@ -23,6 +23,7 @@ import org.eclipse.wst.server.core.IServerAttributes;
 import org.eclipse.wst.server.core.IServerWorkingCopy;
 import org.jboss.ide.eclipse.as.core.server.internal.extendedproperties.ServerExtendedProperties;
 import org.jboss.tools.openshift.common.core.utils.StringUtils;
+import org.jboss.tools.openshift.common.core.utils.UrlUtils;
 import org.jboss.tools.openshift.core.IRouteChooser;
 import org.jboss.tools.openshift.core.OpenShiftCoreUIIntegration;
 import org.jboss.tools.openshift.core.connection.Connection;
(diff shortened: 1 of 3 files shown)


adietish:
@fbricon please review

fbricon:
+1

robstryker:
This makes no sense to me at all.  The entire problem here is that there's no host for the server. The server adapter is supposed to have a host. Several other features don't work if there's no host, or if the host is "localhost".   This behavior is just wrong.  This feature is wrong-headed.

adietish:
@robstryker so that would mean that we have to fixate the server adapter to a route? What is affected by the missing host exactly?

adietish:
as discussed with @robstryker we should always use a host for the adapter. We shouldn't not set it if the user has chosen to get prompted since other wtp facilities might depend on the host. The change that rob suggested and we were complaining about erroneously has no effect on the users ability to get prompted. The change is limited to the area where the adapter is created.

scabanovich:
@adietish , but it seems that computed routeURL is not only used to extract host, it is saved to ATTR_ROUTE. So, I still suspect that change will influence the feature. Then me may need two methods and two parameters, one is routeURL to be saved to ATTR_ROUTE, and it may be null, the other for computing host.

adietish:
yes, that's right, I was not explaning well enough. We need the route as before (blanked if the user decided to get prompted for it) and a host derived from any valid route there is (regardless of the user wanting to be prompted)

scabanovich:
Maybe OpenShiftServerExtendedProperties.fireUpdateRoute() should also reset host computed from the selected route, if hosts of different routes may be different.

adietish:
@scabanovich good idea

scabanovich:
Otherwise, there is one concern. By default, promptRouteButton is true. If user will not open Advanced, will route in the model be always initialized? If not (even because of GTK problems), in case of route == null at computing host, we may check if list of routes is not empty and pick the first.

adietish:
@scabanovich I checked this. The model always has the 1st possible route selected since that's what we have to select in the combo in the advanced section even if it's not visible.

adietish:
@fbricon good to go?

scabanovich:
Ok, I agree, method updateRoute() does it.

adietish:
@scabanovich yes exactly :)

adietish:

PR #1078, opened 2016-03-31
JBIDE-21994 - Cannot show OpenShift 3 application via live reload

xcoulon:
the (fake) RootModule needs to return a (fake) IWebModule to
be able to retrieve load the adapter and get a proper context root
for the fake IModule

Signed-off-by: Xavier Coulon <xcoulon@redhat.com>

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServer.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServer.java
@@ -64,9 +64,11 @@ public class OpenShiftServer extends DeployableServer implements IURLProvider, I
 	private String getContextRoot(IProject moduleProject, IProject deployProject) {
 		String contextRoot = null;
 		if (OpenShiftServerUtils.isIgnoresContextRoot(getServer()) 
-				&& moduleProject.equals(deployProject)) {
+				&& (moduleProject == null // case of the fake RootModule whose project is null
+				|| moduleProject.equals(deployProject))) {
 			contextRoot = "";
 		}
+		
 		return contextRoot;
 	}
 	
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerAdapterFactory.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerAdapterFactory.java
@@ -10,10 +10,12 @@
  ******************************************************************************/
 package org.jboss.tools.openshift.internal.ui.server;
 
+import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.IAdapterFactory;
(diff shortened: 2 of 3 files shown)


xcoulon:
@fbricon can you review the PR, please ?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServer.java
```diff
@@ -53,23 +53,10 @@ public void setDefaults(IProgressMonitor monitor) {
...
 
 	@Override
-	public URL getModuleRootURL(IModule module) {
-		IProject moduleProject = module.getProject();
-		String deployProjectName = OpenShiftServerUtils.getDeployProjectName(getServer());
-		IProject deployProject = ProjectUtils.getProject(deployProjectName);
-		String contextRoot = getContextRoot(moduleProject, deployProject);
-		return JBossServer.getModuleRootURL(module, getServer().getHost(), 80, contextRoot);
+	public URL getModuleRootURL(final IModule module) {
+		return getExtendedProperties().getModuleRootURL(module);
```
robstryker: This really shouldn't be moved. Other classes (like publishing) use these methods, and they are overriding behavior from the superclass DeployableServer.

You'd be better adapting your server to DeployableServer and using the API directly.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerExtendedProperties.java
```diff
@@ -150,4 +154,25 @@ protected IStatus run(IProgressMonitor monitor) {
 		}.schedule();
 	}
 
+
```
robstryker: Again, I don't think this should really belong here. My concern is that extended properties simply becomes a dumping ground for anything we want to expose. It's better to simply use the method where it currently exists.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerAdapterFactory.java
```diff
@@ -111,5 +116,44 @@ public Object loadAdapter(Class adapter, IProgressMonitor monitor) {
 		}
 
 	}
+
```
robstryker: I'm really confused by this. What's the goal here? Why are we adapting to stubs?  If a wst.web project already exists, wouldn't it be better to get it's ACTUAL IWebModule from org.eclipse.jst.server.core  ??

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServer.java
xcoulon: Thanks for your feedback, @robstryker. I'm about to propose an updated version of this PR which keeps this method as-is and avoids changing the ExtendedProperties, as you requested.

fbricon:
+1

robstryker:
This was merged

PR #1080, opened 2016-03-31
JBIDE-21981 - fix error during rsync

robstryker:
cleanup

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
@@ -29,6 +29,7 @@ import org.eclipse.wst.server.core.IServerWorkingCopy;
 import org.eclipse.wst.server.core.model.ServerBehaviourDelegate;
 import org.jboss.ide.eclipse.as.wtp.core.console.ServerConsoleModel;
 import org.jboss.ide.eclipse.as.wtp.core.server.behavior.IPublishController;
+import org.jboss.ide.eclipse.as.wtp.core.server.behavior.ServerProfileModel;
 import org.jboss.tools.as.core.server.controllable.subsystems.internal.StandardFileSystemPublishController;
 import org.jboss.tools.common.util.FileUtils;
 import org.jboss.tools.openshift.common.core.utils.ProjectUtils;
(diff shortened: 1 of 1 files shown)


fbricon:
Unfortunately when we ignore the rsync error, we replace the remote archive by a folder BUT, it means we need to restart the module or else none of the changes (web pages) are loaded by EAP

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
```diff
@@ -78,6 +86,23 @@ public void publishStart(final IProgressMonitor monitor)
...
+	@Override
+	protected boolean supportsJBoss7Markers() {
+		return isEapProfile();
+	}
+	
+	protected boolean isEapProfile() {
+		// quick and dirty
+		String profile = ServerProfileModel.getProfile(getServer());
+		if( "openshift3.eap".equals(profile)) {
+			return false;
```
fbricon: nope, return true

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
```diff
@@ -78,6 +86,23 @@ public void publishStart(final IProgressMonitor monitor)
...
+		}
+		return true;
```
fbricon: nope, return false

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
```diff
@@ -78,6 +86,23 @@ public void publishStart(final IProgressMonitor monitor)
...
+
+	@Override
+	protected boolean supportsJBoss7Markers() {
+		return isEapProfile();
+	}
+	
+	protected boolean isEapProfile() {
+		// quick and dirty
+		String profile = ServerProfileModel.getProfile(getServer());
+		if( "openshift3.eap".equals(profile)) {
```
fbricon: return "openshift3.eap".equals(profile) would have avoided that problem

fbricon:
+1 to apply

PR #1081, opened 2016-04-01
[JBIDE-22059] make sure route chooser "no route error" has a shell

adietish:


adietish:
@xcoulon can you please review this small patch?

jcantrill:
LGTM

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/route/RouteChooser.java
```diff
@@ -41,11 +41,7 @@ public RouteChooser(Shell shell) {
 	public IRoute chooseRoute(List<IRoute> routes) {
 		final IRoute[] selectedRoute = new IRoute[1];
 		Display.getDefault().syncExec(() -> {
-			if (shell == null) {
```
xcoulon: what's the point in having a `shell` field in this class ? Why not just call this `PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell()` when a shell is needed, instead ?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/route/RouteChooser.java
```diff
@@ -55,6 +51,14 @@ public IRoute chooseRoute(List<IRoute> routes) {
...
 	}
 
+	private void ensureHasShell() {
+		if (shell == null) {
+			shell = PlatformUI.getWorkbench()
+					.getActiveWorkbenchWindow()
+					.getShell();
+		}
+	}
+
```
xcoulon: from comment above, this method could return the active workbench window's shell, to be used when necessary.

adietish: +1 you're absolutely right.

PR #1083, opened 2016-04-01
[JBIDE-21869] Fix encoding when retrieving logs

jcantrill:
fix log encoding:

image

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/PodLogsJob.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/PodLogsJob.java
@@ -25,11 +25,10 @@ import org.eclipse.ui.console.IConsole;
 import org.eclipse.ui.console.IConsoleListener;
 import org.eclipse.ui.console.MessageConsole;
 import org.eclipse.ui.console.MessageConsoleStream;
-import org.jboss.tools.openshift.core.server.OpenShiftServerBehaviour;
 import org.jboss.tools.openshift.internal.common.core.job.AbstractDelegatingMonitorJob;
-import org.jboss.tools.openshift.internal.common.ui.OpenShiftCommonUIActivator;
 import org.jboss.tools.openshift.internal.common.ui.console.ConsoleUtils;
 import org.jboss.tools.openshift.internal.core.OCBinaryOperation;
+import org.jboss.tools.openshift.internal.ui.OpenShiftUIActivator;
 
 import com.openshift.restclient.OpenShiftException;
 import com.openshift.restclient.capability.CapabilityVisitor;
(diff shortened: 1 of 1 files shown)


jcantrill:
cc @fbricon @adietish

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/PodLogsJob.java
```diff
@@ -158,26 +158,34 @@ public void run() {
...
 			try {
 				final InputStream logs = new BufferedInputStream(capability.getLogs(true, key.container, OpenShiftBinaryOption.SKIP_TLS_VERIFY));
-				int c;
-				while(running && (c = logs.read()) != -1){
-					if(!stream.isClosed()) {
-						stream.write(c);
-					}
+				byte [] data = new byte [1024];
+				int read = 0;
+				while(running && !os.isClosed() && (read = logs.read(data)) != -1){
```
scabanovich: It is more safe to have !os.isClosed() as the last condition since os can be closed while logs is queried. 
If it may be more efficient to check !os.isClosed() before reading to avoid unnecessary work, then two checks, both before and after may be done.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/PodLogsJob.java
```diff
@@ -158,26 +158,34 @@ public void run() {
...
+					os.write(data, 0, read);
 				}
 			} catch (OpenShiftException e) {
-				stream.println(e.getMessage());
+				OpenShiftUIActivator.getDefault().getLogger().logError(e);
+				try {
+					if(stream != null)
```
scabanovich: Local variable stream could not be set to null. But os can be closed and write will fail.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/PodLogsJob.java
```diff
@@ -158,26 +158,34 @@ public void run() {
...
 			} catch (IOException e) {
 				if(!DOCUMENT_IS_CLOSED.equals(e.getMessage())) {
-					OpenShiftCommonUIActivator.log("Exception reading pod log inputstream", e);
+					OpenShiftUIActivator.getDefault().getLogger().logError("Exception reading pod log inputstream", e);
 				}
 			} finally {
 				try {
-					stream.close();
+					if(stream != null)
+						stream.close();
```
scabanovich: Local variable stream could not be set to null.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/PodLogsJob.java
```diff
@@ -158,26 +158,34 @@ public void run() {
 		protected void runOCBinary(MultiStatus multiStatus) {
 			ConsoleUtils.registerConsoleListener(new ConsoleListener(this));
 			final MessageConsole console = showConsole();
-			final MessageConsoleStream stream = console.newMessageStream();
+			final MessageConsoleStream os = console.newMessageStream();
+			os.setEncoding("UTF-8");
+			BufferedOutputStream stream = new BufferedOutputStream(os);
```
scabanovich: I wonder if we need the buffer here. Below, data is read by big enough portions up to 1024 bytes, which is by itself a good buffer. If data is stored in the buffer output stream until the limit (8192) is reached, os can be closed before any output is shown. I cannot be sure, at least it is worth testing with/without buffer, whatever behaves nicer (and if no difference is noticed, the buffer is not needed too).

robstryker:
merged after rebase. See JIRA

PR #1087, opened 2016-04-04
[JBIDE-21991] remove 'Server Adapter...' menu from Routes context menu

fbricon:
Already applied in master. Consider merging in 4.3.x in case we respin.

cc @alexeykazakov


alexeykazakov:
+1 to merge to 4.3.x IF we decide to respin AND approved by QE.

PR #1093, opened 2016-04-06
[JBIDE-21307] Server Adapter: ensure oc binary that's used for rsync …

jeffmaury:
…is v1.1.1+

Check is implemented when server adapter is created and when the oc binary path is configured

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
@@ -12,6 +12,7 @@ package org.jboss.tools.openshift.internal.core.preferences;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.SystemUtils;
+import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.Platform;
 import org.jboss.tools.openshift.core.preferences.OpenShiftCorePreferences;
 import org.jboss.tools.openshift.internal.common.core.util.CommandLocationBinary;
@@ -78,11 +79,13 @@ public enum OCBinary {
 	/**
 	 * Checks if the oc binary is compatible for rsync publishing.
 	 * 
+	 * @param monitor the progress monitor
+	 * 
 	 * @return true if the oc binary is compatible
 	 * @see https://issues.jboss.org/browse/JBIDE-21307
 	 * @see https://github.com/openshift/origin/issues/6109
 	 */
-	public boolean isCompatibleForPublishing() {
-	    return new OCBinaryValidator(getLocation()).isCompatibleForPublishing();
+	public boolean isCompatibleForPublishing(IProgressMonitor monitor) {
+	    return new OCBinaryValidator(getLocation()).isCompatibleForPublishing(monitor);
 	}
 }
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
@@ -22,7 +22,7 @@ public enum OCBinary {
 	OTHER("oc", new String[] {});
 	
 	private static final String OC_DEFAULTLOCATION_LINUX = "/usr/bin/oc";
-
+	
 	public static OCBinary getInstance() {
 		if (SystemUtils.IS_OS_WINDOWS) {
 			return WINDOWS;
@@ -73,5 +73,16 @@ public enum OCBinary {
 			location = getSystemPathLocation();
 		}
 		return location;
+	}
+	
+	/**
+	 * Checks if the oc binary is compatible for rsync publishing.
+	 * 
+	 * @return true if the oc binary is compatible
+	 * @see https://issues.jboss.org/browse/JBIDE-21307
+	 * @see https://github.com/openshift/origin/issues/6109
+	 */
+	public boolean isCompatibleForPublishing() {
+	    return new OCBinaryValidator(getLocation()).isCompatibleForPublishing();
 	}
 }
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
@@ -1,0 +1,70 @@
+/******************************************************************************* 
+ * Copyright (c) 2016 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
+ ******************************************************************************/ 
+package org.jboss.tools.openshift.internal.core.preferences;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class OCBinaryValidator {
+	
+	/*
+	 * The regular expression use to match line with version of the oc executable
+	 */
+	private static final Pattern OC_VERSION_LINE_PATTERN = Pattern.compile("oc (v.*)");
+	
+	private static final String OC_MINIMUM_VERSION_FOR_RSYNC = "v1.1.1";
+
+	private String path;
+	
+	public OCBinaryValidator(String path) {
+		this.path = path;
+	}
+
+	
+	/**
+	 * Returns the version of the OC binary by running the version command.
+	 * The returned string is formatted as vx.y.z.
+	 * 
+	 * @return the string version
+	 */
+	public String getVersion() {
+        String version = null;
+		try {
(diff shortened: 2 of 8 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/OpenShiftPreferencePage.java
```diff
@@ -128,6 +128,9 @@ private boolean validateLocation(String location) {
 			cliLocationEditor.setErrorMessage(NLS.bind("{0} does not have execute permissions.", file));
 			return false;
 		}
+		if (!new OCBinaryValidator(location).isCompatibleForPublishing()) {
+	        setMessage("Your Openshift 3 client executable is pre-1.1.1 and may cause rsync problems", WARNING);
```
maxandersen: afaics you don't know if it is pre-1.1.1 match could also hit post-1.1.1. 

Suggest to have warning be like: "OpenShift client is not 1.1.1 and may cause rsync problems if it is pre-1.1.1".

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
```diff
@@ -0,0 +1,70 @@
...
+	
+	/**
+	 * Checks if the oc binary is compatible for rsync publishing.
+	 * 
+	 * @return true if the oc binary is compatible
+	 * @see https://issues.jboss.org/browse/JBIDE-21307
+	 * @see https://github.com/openshift/origin/issues/6109
+	 */
+	public boolean isCompatibleForPublishing() {
+	    String version = getVersion();
```
fbricon: You should use org.osgi.framework.Version to perform comparisons

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/OpenShiftPreferencePage.java
fbricon: @maxandersen I don't grok your comment

I'd just say something like "OpenShift client version 1.1.1 or higher is required to avoid rsync issues"

fbricon:
We should display the current oc version (asynchronously) in the page.
Also the warning message is too long, doesn't fit the default preference page width. Maybe we should display the error/warning in the body of the page too:

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
```diff
@@ -0,0 +1,70 @@
...
+		this.path = path;
+	}
+
+	
+	/**
+	 * Returns the version of the OC binary by running the version command.
+	 * The returned string is formatted as vx.y.z.
+	 * 
+	 * @return the string version
+	 */
```
fbricon: Be careful. This should never be called from the UI thread

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
```diff
@@ -74,4 +74,15 @@ public String getLocation() {
...
 	}
+	
+	/**
+	 * Checks if the oc binary is compatible for rsync publishing.
+	 * 
+	 * @return true if the oc binary is compatible
+	 * @see https://issues.jboss.org/browse/JBIDE-21307
+	 * @see https://github.com/openshift/origin/issues/6109
+	 */
+	public boolean isCompatibleForPublishing() {
```
fbricon: potentially long running operation, might add an iprogressmonitor

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
```diff
@@ -0,0 +1,70 @@
...
+	public String getVersion() {
```
fbricon: potentially long running operation, might add an IProgressMonitor parameter

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
```diff
@@ -0,0 +1,70 @@
...
+        String version = null;
+		try {
+            ProcessBuilder builder = new ProcessBuilder(path, "version");
+            Process process = builder.start();
+            String line;
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
```
fbricon: need to handle timeouts / cancelled monitor

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
```diff
@@ -0,0 +1,80 @@
...
+	 * @return true if the oc binary is compatible
+	 * @see https://issues.jboss.org/browse/JBIDE-21307
+	 * @see https://github.com/openshift/origin/issues/6109
+	 */
+	public boolean isCompatibleForPublishing(IProgressMonitor monitor) {
+	    return isCompatibleForPublishing(getVersion(monitor));
+	}
+	
+	public static boolean isCompatibleForPublishing(String version) {
+        return version != null && Version.parseVersion(version).compareTo(OC_MINIMUM_VERSION_FOR_RSYNC) >= 0;
```
fbricon: infortunately, osgi Version doesn't like the oc pattern at all

jeffmaury: Where did you get this number from as I downloaded all OC binaries and did not met this kind of version ?
If we switch back to string compare, then this version is considered less than any of the 1.1.x versions except 1.1

fbricon:
I got it from a Dec. 15th build from https://ci.openshift.redhat.com/jenkins/job/devenv_ami/lastSuccessfulBuild/artifact/origin/artifacts/release/

fbricon:
We're getting there, but I think we should still:
- always display the version of the selected oc, i.e. when opening the preferences page (cf my previous annotated image)
- put the warning on a separate line than the version
- only display major.minor.micro elements of the version (1.1.0.459 is wrong here)
- reduce the width of the page. It seems way too large here

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
```diff
@@ -0,0 +1,94 @@
...
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
+                while (!monitor.isCanceled() && (version == Version.emptyVersion) && ((line = reader.readLine()) != null)) {
+                    Matcher matcher = OC_VERSION_LINE_PATTERN.matcher(line);
+                    if (matcher.matches()) {
+                        version = Version.parseVersion(matcher.group(1));
+                        if ((matcher.groupCount() > 1) && (version.getQualifier().length() == 0)) {
+                            version = new Version(version.getMajor(),
+                                                  version.getMinor(),
+                                                  version.getMicro(),
+                                                  matcher.group(2));
```
fbricon: regexp doesn't seem to properly capture the remainder of the version with v1.1-459-g1904cc3-1904cc3-mac.

But since we don't need that part, we can probably skip it

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/OpenShiftPreferencePage.java
```diff
@@ -128,6 +153,26 @@ private boolean validateLocation(String location) {
...
+            @Override
+            protected IStatus run(IProgressMonitor monitor) {
+                version = new OCBinaryValidator(location).getVersion(monitor);
+                return Status.OK_STATUS;
+            }
+
+            @Override
+            protected IStatus updateUI(IProgressMonitor monitor) {
+                if (!ocMessageComposite.isDisposed() && !monitor.isCanceled()) {
+                    ocMessageLabel.setText(NLS.bind("Your client version is {0}. OpenShift client version 1.1.1 or higher is required to avoid rsync issues.", version));
```
fbricon: Should be "Selected client version: {0}", on a separate line

fbricon:
So, there are still a couple issues: http://screencast.com/t/46pNXrrPp
- for some reason, the layout is completely borked, you need to scroll wayyy down to see the apply button
- when executing the oc command in the background, we need to display a "Checking version..." label instead of the current/last version. Also, the apply button should probably not be enabled until the command returned without errors

fbricon:
Also, the 1st `Version` label is redundant. 


The "Your client version .../Checking version..." should move to the left instead of that redundant label

jeffmaury:
When you say that the apply button should be enabled back when the command returns without errors, do you mean that if the required version is not there, we should leave the button disabled ?

fbricon:
Yes, that's my thinking. However I don't have a *very* strong feeling about it. Consider that optional. Fixing the layout issues right now should be the priority.

fbricon:
Squashed and merged as https://github.com/jbosstools/jbosstools-openshift/commit/8ef3736bc74cb34d5cbd10c519771f43edf4d689

I made a few last modifications:
- ocVersionLabel is set to `Checking OpenShift client version...` when checking
- ocVersionLabel uses an italic font
- labels are reset when changing the binary
- eliminated flicker when moving the cursor in the file field

PR #1099, opened 2016-04-07
[JBIDE-20597] Add connection management to deploy image wizard

jeffmaury:
Add a first page to manage Openshift connections

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizard.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizard.java
@@ -19,10 +19,14 @@ import org.eclipse.core.runtime.jobs.JobChangeAdapter;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.widgets.Display;
 import org.jboss.tools.common.ui.JobUtils;
+import org.jboss.tools.openshift.common.core.connection.IConnection;
 import org.jboss.tools.openshift.common.ui.wizard.AbstractOpenShiftWizard;
+import org.jboss.tools.openshift.core.connection.Connection;
 import org.jboss.tools.openshift.internal.common.core.UsageStats;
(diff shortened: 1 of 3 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizard.java
```diff
@@ -48,6 +52,30 @@ public DeployImageWizard(IDeployImageParameters wizardModel) {
 
 	@Override
 	public void addPages() {
+	    if (getModel().originatedFromDockerExplorer()) {
+	        addPage(new ConnectionWizardPage(this, new IConnectionAware<IConnection>() {
```
scabanovich: Can we pass getModel() instead of wrapper?

jeffmaury: I had trouble casting a IConnectionAware<IConnection> to a IConnectionAware<Connection> so if your solution works, let's adopt it

scabanovich: In this snippet cast gives a compilation error:

	void test1() {
		class A {};
		class B extends A {};
		class C<A> {};
		C<A> c = new C<A>();
		C<B> cb = new C<B>();
		c = (C<B>)cb;
	}

This one is correct:

	void test2() {
		class A {};
		class B extends A {};
		class C<A> {};
		C<? extends A> c = new C<A>();
		C<B> cb = new C<B>();
		c = (C<B>)cb;
	}

Sometimes it is very annoying, for example, if I want to store and operate a List<MyImpl> but have to return List<MyInterface> (MyImpl implements MyInterface), a cannot return the list as is, it does not cast.

Still, there is a sense in it. ConnectionWizardPage does not mind accepting any implementation of IConnection, but it should be unambiguously stated with <? extends IConnection> (or with a type parameter).

PR #1101, opened 2016-04-07
[JBIDE-22129] fix trying to reconnect when project is deleted

jcantrill:
cc @fbricon @adietish 

This is really a partial fix to check for project existence.  Additional work should really be done to only check project existence once instead of once per kind.  Deferring that work because of pending release.

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
@@ -96,7 +96,8 @@ public class WatchManager {
 	private static enum State {
 		STARTING,
 		CONNECTED,
-		DISCONNECTED
+		DISCONNECTED, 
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -96,7 +96,8 @@ private void startWatch(final IProject project, int backoff, long lastConnect, W
 	private static enum State {
 		STARTING,
 		CONNECTED,
-		DISCONNECTED
+		DISCONNECTED, 
+		STOPPING
```
scabanovich: I cannot see a reference to new state STOPPING

fbricon:
Except for Slava's remark, everything looks good. Successfully tested.
+1 (bound to dealing with State.STOPPING)

adietish:
rebased and merged into jbosstools-4.3.x, cherry-picked into master

PR #1102, opened 2016-04-07
JBIDE-22057 - fix to ensure a full publish occurs after restart

robstryker:


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -105,14 +109,31 @@ public void launch(ILaunchConfiguration configuration, String mode,
...
 		int state = pollState();
+		
+		if (!Objects.equals(oldMode, mode)) {
+			IModule[] modules = getServer().getModules();
+			for( int i = 0; i < modules.length; i++ ) {
+				((Server)getServer()).setModulePublishState(new IModule[]{modules[i]}, IServer.PUBLISH_STATE_FULL);
+			}
+			
+			// Kick publish server job
+			new Job("Publishing server " + getServer().getName()) {
```
fbricon: you prolly don't want to publish the server when it's stopped

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -105,14 +109,31 @@ public void launch(ILaunchConfiguration configuration, String mode,
 			mode = currentMode;
 			throw e;
 		} finally {
-			checkServerState(beh, mode);
+			checkServerState(beh, currentMode, mode);
```
fbricon: So it seems to me currentMode is always null when restarting the server (whether it was in normal or in debug mode).

Maybe we need to carry around the last known mode of the server (stored in the behavior's shared data?)

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -105,14 +111,36 @@ public void launch(ILaunchConfiguration configuration, String mode,
...
 		int state = pollState();
+		
+		if (!Objects.equals(oldMode, mode)) {
+			IModule[] modules = getServer().getModules();
+			for( int i = 0; i < modules.length; i++ ) {
+				((Server)getServer()).setModulePublishState(new IModule[]{modules[i]}, IServer.PUBLISH_STATE_FULL);
+			}
+			
+			// Kick publish server job
+			Job j = new Job("Publishing server " + getServer().getName()) {
```
fbricon: no need to instantiate that job unless we're sure to use it. It should go inside the if block below

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerBehaviour.java
```diff
@@ -33,12 +33,19 @@
...
+	
 	@Override
 	public void setServerStarted() {
 		super.setServerStarted();
 		launchPostStartupJobs();
 	}
 
+	public void restart(String launchMode) throws CoreException {
+		putSharedData(CURRENTLY_RESTARTING, true);
+	}
```
fbricon: @Override
	public void restart(String launchMode) throws CoreException {
		putSharedData(CURRENTLY_RESTARTING, true);
		super.restart(launchMode);
	}

fbricon:
Other than the broken restart (needs to call super), everything works as advertised.
Once that's fixed, along the minor problem of job instantiation, it'll be good to go

adietish:
works as expected, modifications are still present after a restart in debug mode.

adietish:
rebased and merged into jbosstools-4.3.x, cherry-picked into master

PR #1103, opened 2016-04-07
[JBIDE-22128] better detect EAP style deployments

fbricon:


fbricon:
@robstryker @adietish for your reviewing pleasure

robstryker:
tl;dr this patch lets you check the image name rather than the template, which will work for more usecases.  Still seems fragile, but a +1 bc it looks like it'll work.

alexeykazakov:
+1

PR #1104, opened 2016-04-08
[JBIDE-22034] Replace Pin Console toggle tool item by Show Console Ou…

jeffmaury:
…tput When Standard Out Changes toggle item

Logs are written to the console with default activation flag


fbricon:
Does this need the PR in servers to be applied, in order to compile?

jeffmaury:
The PR is mostly implemented in servers. The part in Openshift is that logs are now written to the console without the activate flag, so as to let the user decide if auto activation will take place or not. Please not that there is no compilation dependency between openshift and servers brought by this PR

fbricon:
+1 from me.

@robstryker can you please apply this one along https://github.com/jbosstools/jbosstools-server/pull/409? Even though there's no compilation requirement, I'd prefer we don't lose functionality if both PRs are not applied simultaneously.

robstryker:
merged after rebase

PR #1106, opened 2016-04-11
[JBIDE-22132] Resource name is badly generated in Deploy to Openshift…

jeffmaury:
… wizard

- Add converter and model now uses the converter
- Add unit test

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DockerImage2OpenshiftResourceConverter.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DockerImage2OpenshiftResourceConverter.java
@@ -1,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.openshift.internal.ui.wizard.deployimage;
+
+import com.openshift.restclient.images.DockerImageURI;
+
+/**
+ * 
+ * A converter for the docker image name to Openshift resource mapping.
+ * Will replace '_' by '-', skip other forbidden characters and trucate the
+ * result to 24.
+ * 
+ * @author Jeff Maury
+ *
+ */
+public class DockerImage2OpenshiftResourceConverter {
+
+    public String convert(DockerImageURI dockerImage) {
+        String name = dockerImage.getName();
+        StringBuilder builder = new StringBuilder(name.length());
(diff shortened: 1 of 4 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DockerImage2OpenshiftResourceConverter.java
```diff
@@ -0,0 +1,45 @@
...
+            char c = name.charAt(i);
+            if (((c >= '0') && (c <= '9')) ||
+                ((c >= 'a') && (c <= 'z')) ||
+                ((c >= 'A') && (c <= 'Z'))) {
+                builder.append(c);
+            } else if (c == '_') {
+                builder.append('-');
+            }
+        }
+        if (builder.length() > 24) {
```
jcantrill: I would suggest exposing the max  length from the validator so we ensure they do not vary independently

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DockerImage2OpenshiftResourceConverter.java
```diff
@@ -0,0 +1,47 @@
...
+            char c = name.charAt(i);
+            if (((c >= '0') && (c <= '9')) ||
+                ((c >= 'a') && (c <= 'z')) ||
+                ((c >= 'A') && (c <= 'Z'))) {
+                builder.append(c);
+            } else if (c == '_') {
+                builder.append('-');
+            }
+        }
+        if (builder.length() > MAXLENGTH) {
```
scabanovich: Cycle 'for' can have condition buildre.length() <= MAXLENGTH so that we stop building rather than build too long and then cut.

scabanovich: Still, there may be a thing to cut. The char at max length may happen to be '-', and it is not allowed for name to end with it. All trailing '-' symbols have to be cut.

fbricon:
There seems to be a regression, when clicking finsih after selecting the docker image in the search wizard, the text field is empty : http://screencast.com/t/iTJX7SjB

If you click next instead and select a tag, the field is properly set.

scabanovich:
@fbricon , it is possible that validator does not let empty string to model, so model remains with valid previous value. Then, when you select that same value in the dialog, model is notified, but it already has that value and it does not notify ui because it did not change. If you select in dialog another value, will you see it in the text fielld?

fbricon:
@scabanovich : looks like you're right. If I search for a different value, then the field is properly filled.

scabanovich:
@fbricon , then it is a new issue, not caused by this PR. Maybe it deserves a separate issue in jira because there can be other components with the same behaviour. I see one fix: before calling setImageName(newValue) check if it has the same value, then call setImageName(""); and only after this check and adjustment call setImageName(newValue).

fbricon:
@scabanovich: right, this problem can be reproduced without that PR, I'll open a separate JIRA

fbricon:
Squashed and merged in master. Thanks @jeffmaury !

PR #1119, opened 2016-04-20
[JBIDE-22118] Add ability to create an app using a builder image

jcantrill:
Allows a user to create apps from a builder image

cc @fbricon @adietish


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
```diff
@@ -52,20 +52,20 @@
 	 * in name or tags.
 	 * 
 	 * @param filterText
-	 * @param template
+	 * @param resource
 	 * @return
 	 */
-	public static boolean isMatching(final String filterText, IResource template) {
+	public static boolean isMatching(final String filterText, IResource resource) {
```
adietish: I guess that a guard for resource == null (like the one we already have for the filterText or maybe even assert vua org.eclipse.core.runtime.Assert that throws an assertion (runtime-) exception?) would be a great idea?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
```diff
@@ -78,6 +78,23 @@ public Boolean visit(ITags capability) {
 		}, Boolean.FALSE);
 	}
 
+	public static boolean isMatching(final String filterText, String name, Collection<String> tags) {
+		if (StringUtils.isBlank(filterText)) {
```
adietish: To me it looks as if this #isMatching (for tags) and the other #isMatching (for a resource) share at least
- entry parameter validation
- filterText sanitation (#replaceAll(",", " ")
- containsAll check.
Shouldnt we use share the code for better maintainability?

plugins/org.jboss.tools.openshift.ui/META-INF/MANIFEST.MF
```diff
@@ -72,4 +72,6 @@ Export-Package: org.jboss.tools.openshift.internal.ui;x-friends:="org.jboss.tool
  org.jboss.tools.openshift.internal.ui.wizard.deployimage;x-friends:="org.jboss.tools.openshift.test",
  org.jboss.tools.openshift.internal.ui.wizard.deployimage.search;x-friends:="org.jboss.tools.openshift.test",
  org.jboss.tools.openshift.internal.ui.wizard.importapp;x-friends:="org.jboss.tools.openshift.test",
- org.jboss.tools.openshift.internal.ui.wizard.newapp;x-friends:="org.jboss.tools.openshift.test"
+ org.jboss.tools.openshift.internal.ui.wizard.newapp;x-friends:="org.jboss.tools.openshift.test",
+ org.jboss.tools.openshift.internal.ui.wizard.newapp.fromimage,
```
adietish: missing x-friends:=

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
```diff
@@ -142,8 +146,6 @@ public StyledString getStyledText(Object element) {
 			case ResourceKind.SERVICE:
 				IService service = (IService) resource;
 				return getStyledText(service);
-			case ResourceKind.TEMPLATE:
```
adietish: we're removing support for templates since we're not displaying them any more, right? Shouldnt we remove the other types we're not displaying either?

jcantrill: This class was updated to be a common label provider anywhere in the UI where we need such capability.  I think it would be reasonable to leave the others there.  We may need to add template back in at some time if we ever wish to provide a tab or some other visibility into templates other then the way we display them in the list page.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
jcantrill: added null check

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerLabelProvider.java
adietish: mmkay, but I still strongly believe that non-used code is a liability. I dont see a need (for now) that justify this to remain in.

jcantrill:
@adietish @fbricon  updated to support changes to https://github.com/openshift/openshift-restclient-java/pull/150 and rebased on master

fbricon:
There's a regression with the project creation: http://screencast.com/t/gaN4DLj1

- the new project doesn't show up in the list of projects
- the list of templates/builders is then grouped by project (minus the new project)

It seems the list of template/builders is way too aggressive abbreviating labels (too many `...`)

Also, the deployment configuration and scalability page crops the NOTICE label. Maybe the data volumes table could be reduced to 2-3 rows.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ApplicationSourceListPage.java
```diff
@@ -615,36 +624,36 @@ private TreeViewer createServerTemplatesViewer(Composite parent, final Text temp
 		TreeViewer viewer = new TreeViewer(parent, SWT.BORDER | SWT.SINGLE | SWT.V_SCROLL | SWT.H_SCROLL);
 		IListProperty childrenProperty = new MultiListProperty(
 				new IListProperty[] {
-						BeanProperties.list(ITemplateListPageModel.PROPERTY_TEMPLATES),
+						BeanProperties.list(IApplicationSourceListPageModel.PROPERTY_APP_SOURCES),
 						BeanProperties.list(ObservableTreeItem.PROPERTY_CHILDREN) });
 		ObservableListTreeContentProvider contentProvider = new ObservableListTreeContentProvider(
 				childrenProperty.listFactory(), null);
 		viewer.setContentProvider(contentProvider);
 		viewer.setLabelProvider(new ObservableTreeItemStyledCellLabelProvider());
```
scabanovich: @jcantrill , please set to inner OpenShiftExplorerLabelProvider explorerLabelProvider label limit 100. Maybe it is better to do by adding a method, or it can be as well done just inside ObservableTreeItemStyledCellLabelProvider, since it is currently used in wizards only.

Auto-adjusting abbreviation inside available width can be not a good idea (though worth to study), because view asks texts from label provider on refresh, not on resize, and calling refresh of view on every resize event may cause performance issues. So, if we just have an idea of the optimal wizard size, we may fit label limit to it. The default 60 is of course too small.

fbricon:
Merged as https://github.com/jbosstools/jbosstools-openshift/commit/b04e5eab76dc335af0024560761724bc06807653, after fixing the abbreviation length, some missing maven config and a few test regressions

PR #1124, opened 2016-04-22
[JBIDE-22237] Fix property table update where fired event does not match method signature

jcantrill:
cc @fbricon @adietish 

Fixed event fire.  relies on merge from rest client

### DIFF
--- a/plugins/org.jboss.tools.openshift.client/.classpath
+++ b/plugins/org.jboss.tools.openshift.client/.classpath
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry exported="true" kind="lib" path="lib/openshift-restclient-java-4.0.3.Final.jar"/>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry exported="true" kind="lib" path="lib/openshift-restclient-java-4.0.4-SNAPSHOT.jar"/>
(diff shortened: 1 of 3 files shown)


plugins/org.jboss.tools.openshift.client/.classpath
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry exported="true" kind="lib" path="lib/openshift-restclient-java-4.0.3.Final.jar"/>
+	<classpathentry kind="lib" path="lib/openshift-restclient-java-4.0.3.Final.jar"/>
```
adietish: shouldnt this be removed (we have a 2nd entry for the client lib 4.0.4 further down)

adietish: I believe that we also need a change in the pom.xml where we bump the client version 4.0.3->4.0.4

jcantrill:
@adietish removed extraneous lib.

fbricon:
This PR also requires to change the rest client version in the *.client/pom.xml

Still, using the latest client 4.0.4 snapshot, I see the error described in https://issues.jboss.org/browse/JBIDE-22237

fbricon:
I no longer see the issue after merging the latest changes. Merged as https://github.com/jbosstools/jbosstools-openshift/commit/bafb33de27c84b3bcde5327cf69113e8c8a708bc

PR #1130, opened 2016-04-28
[JBIDE-22190] Application wizard: Support "URL" templates to create a new application

jeffmaury:
[JBIDE-22190] Application wizard: Support "URL" templates to create a new application

- Added a new tab to the template page to allow specify a template by URL

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ApplicationSourceListPage.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ApplicationSourceListPage.java
@@ -786,7 +786,7 @@ public class ApplicationSourceListPage  extends AbstractProjectPage<IApplication
 
     @Override
     public void handleValueChange(ValueChangeEvent<? extends Object> event) {
-        if (model.isUseLocalAppSource()) {
+        if (model.isUseLocalAppSource() && StringUtils.isNotEmpty(model.getLocalAppSourceFileName())) {
             Job job = new Job("Loading application source") {
                 
                 @Override
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ApplicationSourceListPage.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ApplicationSourceListPage.java
@@ -25,6 +25,7 @@ import org.eclipse.core.databinding.observable.value.IValueChangeListener;
 import org.eclipse.core.databinding.observable.value.ValueChangeEvent;
 import org.eclipse.core.databinding.property.list.IListProperty;
 import org.eclipse.core.databinding.property.list.MultiListProperty;
+import org.eclipse.core.databinding.validation.IValidator;
 import org.eclipse.core.databinding.validation.MultiValidator;
 import org.eclipse.core.databinding.validation.ValidationStatus;
 import org.eclipse.core.resources.IFile;
@@ -291,16 +292,19 @@ public class ApplicationSourceListPage  extends AbstractProjectPage<IApplication
 				.span(3, 1)
 				.applyTo(txtLocalTemplateFileName);
 		IObservableValue localTemplateFilename = WidgetProperties.text(SWT.Modify).observe(txtLocalTemplateFileName);
+		IValidator validator = (o -> {
+		    IStatus status = ValidationStatus.ok();
+		    if (!OpenshiftUIConstants.URL_VALIDATOR.isValid(o.toString()) &&
+		        StringUtils.isNotBlank(o.toString()) &&
+		        !isFile(o.toString())) {
+		        status = ValidationStatus.error(o +" is not a file");
+		    }
+		    return status;});
 		ValueBindingBuilder
 				.bind(localTemplateFilename )
-             .validatingBeforeSet(o -> !OpenshiftUIConstants.URL_VALIDATOR.isValid(o.toString()) && !isFile(o.toString())?
-                        ValidationStatus.error(txtLocalTemplateFileName.getText() +" is not a file"):
-                        ValidationStatus.ok()) 
+             .validatingBeforeSet(validator) 
 				.to(BeanProperties.value(
 						IApplicationSourceListPageModel.PROPERTY_LOCAL_APP_SOURCE_FILENAME).observe(model))
-				.validatingBeforeSet(o -> !OpenshiftUIConstants.URL_VALIDATOR.isValid(o.toString()) && !isFile(o.toString())?
-						ValidationStatus.error(txtLocalTemplateFileName.getText() +" is not a file"):
-                        ValidationStatus.ok()) 
 				.in(dbc);
 
 		// browse button
@@ -786,7 +790,7 @@ public class ApplicationSourceListPage  extends AbstractProjectPage<IApplication
 
     @Override
     public void handleValueChange(ValueChangeEvent<? extends Object> event) {
-        if (model.isUseLocalAppSource() && StringUtils.isNotEmpty(model.getLocalAppSourceFileName())) {
+        if (model.isUseLocalAppSource()) {
             Job job = new Job("Loading application source") {
                 
                 @Override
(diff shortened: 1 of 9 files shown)


fbricon:
Can't we simply reuse the Local Template tab, rename it Custom Template and then do everything from the text field there? i.e same field could cover URL, local or workspace path

jeffmaury:
I don't see how will you make the distinction between file and URL ? Another UI component ?

fbricon:
you simply check if the text is a well formed url, with a scheme, if not ignore it and consider it's a local path

jeffmaury:
Yes but you face errors like 'Can't open file http:'. Not speaking of Windows files C:/xx/y that will be handled as URLs

fbricon:
if c:/xx/y generates a valid url object, then check if the scheme is http(s) or file and that should be good enough

fbricon:
See if http://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/UrlValidator.html#isValid(java.lang.String) would be good enough.

fbricon:
Can you rebase from master please?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ApplicationSourceListPage.java
```diff
@@ -248,7 +249,7 @@ private IObservableValue createLocalTemplateControls(TabFolder tabContainer, IOb
 			.applyTo(parent);
 
 		Label lbl = new Label(parent, SWT.NONE);
-		lbl.setText("Select a local template:");
+		lbl.setText("Select a local template file or a full URL:");
 		GridDataFactory.fillDefaults().span(3,1).applyTo(lbl);
 		
 		// local template file name
```
fbricon: need to .span(3,1) that txtLocalTemplateFileName field

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ApplicationSourceListPage.java
```diff
@@ -277,7 +278,7 @@ private IObservableValue createLocalTemplateControls(TabFolder tabContainer, IOb
 		
 		// browse button
 		Button btnBrowseWorkspaceFiles = new Button(parent, SWT.NONE);
-		btnBrowseWorkspaceFiles.setText("Workspace...");
+		btnBrowseWorkspaceFiles.setText("Browse Workspace...");
 		GridDataFactory.fillDefaults()
 				.align(SWT.LEFT, SWT.CENTER)
```
fbricon: align right on next line

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ApplicationSourceListPage.java
```diff
@@ -261,14 +262,14 @@ private IObservableValue createLocalTemplateControls(TabFolder tabContainer, IOb
...
+						ValidationStatus.error(txtLocalTemplateFileName.getText() +" is not a file"):
+                        ValidationStatus.ok()) 
 				.in(dbc);
 
 		// browse button
 		Button btnBrowseFiles = new Button(parent, SWT.NONE);
-		btnBrowseFiles.setText("File system...");
+		btnBrowseFiles.setText("Browse File System...");
 		GridDataFactory.fillDefaults()
 				.align(SWT.LEFT, SWT.CENTER)
```
fbricon: align right on next line

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizardModel.java
```diff
@@ -215,8 +221,12 @@ public boolean isUseLocalAppSource() {
 		return useLocalAppSource;
 	}
 
-	public InputStream createInputStream(String fileName) throws FileNotFoundException {
-		return new FileInputStream(fileName);
+	public InputStream createInputStream(String fileName) throws IOException {
+	    if (!OpenshiftUIConstants.URL_VALIDATOR.isValid(fileName)) {
```
fbricon: Generally, I'd prefer checking for isValid rather than !isValid

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizardModel.java
```diff
@@ -215,8 +221,12 @@ public boolean isUseLocalAppSource() {
...
+	        return new FileInputStream(fileName);
+	    } else {
```
fbricon: 2 problems here:
- this method is called on the UI thread
- it ignores Eclipse's proxy settings. You should look at using ECF instead

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizardModel.java
```diff
@@ -117,7 +121,7 @@ private IApplicationSource getLocalAppSource(String filename) {
 		IResource resource = null;
 		filename = VariablesHelper.replaceVariables(filename);
 		try {
-			if (!Files.isRegularFile(Paths.get(filename))) {
+			if (!OpenshiftUIConstants.URL_VALIDATOR.isValid(filename) && !Files.isRegularFile(Paths.get(filename))) {
 				return null;
 			}
```
fbricon: this test can probably be put outside the try block

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizardModel.java
```diff
@@ -117,7 +121,7 @@ private IApplicationSource getLocalAppSource(String filename) {
...
 			resource = resourceFactory.create(createInputStream(filename));
```
fbricon: the inputstream is not closed.
You should use 
```java
try (InputStream is = createInputStream(filename)){
...			
}
```

fbricon: Following on last comment, we should check the inputstream is empty before trying to create the resource. That'll give is more info on debugging.
```java 
String content = IOUtils.toString(is);
if (StringUtils.isBlank(content)) {
  throw new OpenShiftException(NLS.bind("\"{0}\" is empty", filename));
}
resource = resourceFactory.create(content);
```

fbricon:
NewApplicationWizardModelTest doesn't compile, and even when fixed, some tests fail

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizardModel.java
```diff
@@ -207,29 +222,58 @@ public void loadResources() {
...
 	@Override
 	public boolean isUseLocalAppSource() {
 		return useLocalAppSource;
 	}
 
-	public InputStream createInputStream(String fileName) throws FileNotFoundException {
-		return new FileInputStream(fileName);
+	public InputStream createInputStream(String filename, IProgressMonitor monitor) throws IOException {
+	    String url = OpenshiftUIConstants.URL_VALIDATOR.isValid(filename)?filename:new File(filename).toURI().toString();
+	    try (ByteArrayOutputStream out = new ByteArrayOutputStream()) {
```
fbricon: I'd rather prefer we only call OpenshiftUIConstants.TRANSPORT_UTILITY.download if it's a valid URL, else expect a file and return a fileinputstream directly

fbricon:
I can see a couple issues in http://screencast.com/t/JVFfcjgRi:
- the load app source job is triggered even if the template name is empty (when entering the tab)
- the description of the template is loaded, shows up and then disappears

fbricon:
Still seeing some problems (including regressions) http://screencast.com/t/F1yfQRdC5x

- when loading the template, it says `is not a file `
- the details pane is cropped (it used to work fine)
- resetting the path doesn't reset the details, keeps saying `is not a file`

Also, not shown on the screencast:
- when a server side template is selected in the 1st tab, the `is not a file ` error is shown.

I'm testing with this template: https://raw.githubusercontent.com/openshift/origin/master/examples/sample-app/application-template-stibuild.json

fbricon:
So the loading on en empty field is really annoying http://screencast.com/t/kZQ1oQEluP3.
I still believe we should not try to load the resource if the field is blank. 
Also, a blank field should be converted to a null value and assigned to the model.
"is not a valid file" is definitely not the error message we want to display if the field is blank (rather "You must enter a valid file path or URL")

PR #1143, opened 2016-05-06
[JBIDE-22275] Make oc version parsing more robust

fbricon:


robstryker:
Looks good, so long as you can get a guarantee from the openshift guys that the returned string will *always* fit that format, and not change.

PR #1144, opened 2016-05-06
[JBIDE-21823] Unable to re-edit Service config

jeffmaury:
- Add refresh of Openshift explorer when a resource is refreshed
- Suppressed the use of the JSON editor (https://bugs.eclipse.org/bugs/show_bug.cgi?id=493145)

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
@@ -65,7 +65,9 @@ public class OpenShiftExplorerContentProvider extends BaseExplorerContentProvide
 				cache.flushFor((Connection) newValue);
 			}
 			refreshViewer(newValue);
-		} 
+		} else if (ConnectionProperties.PROPERTY_RESOURCE.equals(property)) {
+		    refreshViewer(newValue);
+		}
 	}
 
 	/**
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/EditResourceHandler.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/EditResourceHandler.java
@@ -16,7 +16,6 @@ import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.widgets.Shell;
-import org.eclipse.ui.IEditorDescriptor;
 import org.eclipse.ui.IEditorRegistry;
 import org.eclipse.ui.IStorageEditorInput;
 import org.eclipse.ui.IWorkbenchPage;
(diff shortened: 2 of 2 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/EditResourceHandler.java
```diff
@@ -67,25 +66,18 @@ private void openInEditor(IWorkbenchWindow window, Connection connection, IResou
...
-			   try {
-				   page.openEditor(input, editorId);
-			   } catch (PartInitException O_o) {
-				   if (defaultJsonEditorId.equals(editorId)){
-					   throw O_o;
-				   }
-				   //try to fall back on default editor
-				   page.openEditor(input, defaultJsonEditorId);
-			   }
+			   page.openEditor(input, defaultJsonEditorId);
```
fbricon: I would keep that code. Only the implementation for getDefaultJSONEditorDescriptorId needs to change right now

jeffmaury: Yes but we would be facing the switch to the JSON editor depending on the user Eclipse configuration

fbricon: fair enough

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -65,7 +65,9 @@ protected void handleConnectionChanged(IConnection connection, String property,
 				cache.flushFor((Connection) newValue);
 			}
 			refreshViewer(newValue);
-		} 
+		} else if (ConnectionProperties.PROPERTY_RESOURCE.equals(property)) {
+		    refreshViewer(newValue);
```
fbricon: this should not be necessary. It used to work when there was no JSON editor.
Now, given a *lot* or resource changes are emitted regularly, that would cause the viewer to flicker a lot

jeffmaury: For me, it is the part of the modifications that fixed the problem. If you remove that, then if you reedit from the  Openshift explorer view after you modified the resource and closed the editor will bring you to the outdated resource.
The other part of the modifications are just there to make sure when use the base TextEditor in order to listen to both editor events and connection manager events

adietish: @jeffmaury can you outline in a few words why this change to the explorer fixes the problem in the editor?

adietish:
I tested this and it works for me like a charm in the happy path with valid changes but also if I provide non-valid json. The default editor will not save. 
The pity in this change where we force the default editor (because of shortcomings in the json editor, https://bugs.eclipse.org/bugs/show_bug.cgi?id=493145) is that one gets an error but wont get told in all details what's exactly wrong (you get told of the 1st illegal character that was found). Would it be possible to put the cursor to the faulty place in the editor or is this a big change?

adietish:
Responding myself to the suggestion: jboss-dmr, which does the json parsing wont report the location of the illegal content. I thus dont think that this is even possible without further changes.

adietish:
rebased and merged into master.
Thx!

PR #1152, opened 2016-05-10
JBIDE-22325 - Move job to disable all widgets into foundation, and al…

robstryker:
…low astools to use it


fbricon:
OK to apply once https://github.com/jbosstools/jbosstools-base/pull/503 is done and a build of foundation is available in the snapshot update site

PR #1155, opened 2016-05-11
[JBIDE-22358] Add load and presist of advanced properties for a connection

jcantrill:
Add a mechanism to load and save advanced properties of a connection

cc @adietish @fbricon

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
@@ -14,7 +14,9 @@ import static org.apache.commons.lang.StringUtils.isNotBlank;
 
 import java.io.IOException;
 import java.net.MalformedURLException;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 import org.apache.commons.lang.ObjectUtils;
@@ -37,7 +39,7 @@ import org.jboss.tools.openshift.internal.common.core.security.SecureStore;
 import org.jboss.tools.openshift.internal.common.core.security.SecureStoreException;
 import org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator;
 
-import com.openshift.restclient.ClientFactory;
+import com.openshift.restclient.ClientBuilder;
 import com.openshift.restclient.IClient;
 import com.openshift.restclient.IResourceFactory;
 import com.openshift.restclient.ISSLCertificateCallback;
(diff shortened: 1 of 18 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
```diff
@@ -88,6 +90,23 @@ public Connection(IClient client, ICredentialsPrompter credentialsPrompter, ISSL
...
 	}
 	
+	@Override
+	public Properties getExtendedProperties() {
+		return extendedProperties;
+	}
+
+	@Override
+	public void setExtendedProperty(String name, String value) {
+		Properties oldExt = new Properties(extendedProperties);
```
fbricon: I always found default properties to be tricky. For instance, considering:
```java
  Properties p1 = new Properties();
  p1.put("foo", "bar");
  Properties p2 = new Properties(p1);
```
then p1.equals(p2) return false!

jcantrill: Is this a concern for me?  I am just trying to make sure 'firePropertyChange' to actually propagates the event.  Should it really be send the original as old and then create a new one?

fbricon: my point is if your properties contain foo=bar, and you call setExtendedProperty("foo","bar"), then it'll fire a change event, but it shouldn't

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
```diff
@@ -88,6 +91,23 @@ public Connection(IClient client, ICredentialsPrompter credentialsPrompter, ISSL
 		this.sslCertificateCallback = sslCertCallback;
 	}
 	
+	@Override
+	public Map<String, Object> getExtendedProperties() {
+		return new HashMap<>(extendedProperties);
+	}
+
+	@Override
+	public void setExtendedProperty(String name, String value) {
```
fbricon: Either the value is an Object or the extendedProperties map is Map<String, String>

jcantrill:
cc @fbricon  updated to use map instead of property

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/IOpenShiftConnection.java
```diff
@@ -43,4 +46,24 @@
...
+	 * @param ext
+	 */
+	void setExtendedProperties(Map<String, Object> ext);
+	
+	/**
+	 * Set a value of an extended property
+	 * @param name
+	 * @param value
+	 */
+	void setExtendedProperty(String name, String value);
```
fbricon: value type is not consistent with the rest

fbricon:
It doesn't seem we have tests covering the actual (de)serialization of extended properties on disk

jcantrill:
@fbricon updated to use set(String,Object) and added save/load test

PR #1158, opened 2016-05-12
[JBIDE-22278] Show OpenShift Explorer when a new Connection is created

fbricon:
OpenShift Explorer is shown when creating the OpenShift Connection from the CDK server adapter or the regular New OpenShift Application wizard.

http://screencast.com/t/j6EWxqS2adP


scabanovich:
I wonder if OpenShift 2 core plugin may start after OpenShift UI plugin. Then the listener will be notified that some connections are added (if there is an OpenShift 2 connection in the workspace) - and the explorer will be opened right on start. Listener may check that the connection is an instance of OpenShift 3.

fbricon:
mmm I haven't checked if a connection added event is fired on workspace startup. If it does, then it's probably not a good idea to always show the openshift explorer on startup.

Putting that issue aside, the listener should probably be added to the common ui plugin instead. The explorer should be shown regardless of the connection type.

I'll do some more testing

fbricon:
Moved the listener to the common plugin. Makes sense to have both new OS 2 & 3 connections trigger the OpenShift explorer visibility.

The explorer won't get displayed until the OS2 or 3 plugins are activated.

Also added some IT.

scabanovich:
Looks good to me.

PR #1161, opened 2016-05-13
[JBIDE-22360] Extract OAuth token from the New Connection wizard

jeffmaury:


fbricon:
I can't test it on OSX, as we have to open a browser instead, to workaround some Mac specific issue. Although I need to check if we still need the workaround in Neon.

@adietish can you please check this PR?

fbricon:
I managed to test it, LGTM

PR #1164, opened 2016-05-16
[JBIDE-22361] fixes explorer collapse on re-edit caused by multiple d…

jcantrill:
…eployment registration

[JBIDE-21823] fixes re-edit of service

cc @fbricon

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/DeploymentResourceMapper.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/DeploymentResourceMapper.java
@@ -21,6 +21,7 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.jboss.tools.common.databinding.ObservablePojo;
 import org.jboss.tools.openshift.common.core.connection.ConnectionsRegistryAdapter;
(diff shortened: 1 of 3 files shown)


fbricon:
It doesn't seem to work http://screencast.com/t/kNk79KuW

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/DeploymentResourceMapper.java
```diff
@@ -129,24 +124,12 @@ private void load(String kind) {
 
 	private void buildDeployments() {
 		if (state.compareAndSet(State.UNINITIALIZED, State.LOADING)) {
-			for (String kind : WatchManager.KINDS) {
-				load(kind);
-			}
 			WatchManager.getInstance().stopWatch(project);
+			ConnectionsRegistrySingleton.getInstance().removeListener(connectionListener);
```
scabanovich: While mapper builds deployments it does not listen to changes in the project. But while it iterates over kinds of resources, some changes to already loaded kinds can happen in the project, even if with a small probability. Could we add for the time of loading another listener that would just register the fact of a change. If a change is detected, then loading would be repeated.

jcantrill:
@fbricon removed refresh from viewer.  Please try again.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/DeploymentResourceMapper.java
jcantrill: We may not need that.  Part of a watching a resource involves retrieving the version from which to begin a watch and you use that when starting it; you then get all the changes since that version.  It would probably make more sense to try and capture the resource version from our initial load and start from there.  That would mean we might get duplicates but then at least we have a way to filter them out.  Additionally, projects are intended to be small.  We should hope users will not have large, copious amounts of resources.

scabanovich: One use case that I see is that while build is going user calls refresh. Even though there may be not much resources, the time of refresh may be not negligible because each kind is retrieved with a separate remote request. Taken that build goes for several minutes, and refresh for several seconds, we get probability that the build status changes after it is loaded but before refresh is completed about several per cents. So, some logic to prevent that should be implemented, the one you have described sounds good.

PR #1165, opened 2016-05-17
[JBIDE-22260] - Change 'Manage projects...' links to "New..." buttons

jeffmaury:
- Replace link with button


fbricon:
Works fine, but we should also transform the new docker connection link to a New button as well, as discussed in JBIDE-22260

PR #1166, opened 2016-05-17
[JBIDE-22363] open Token page in embedded browser on OSX

fbricon:


mmalina:
+1 Works for me. Token is shown in a window and the string is copied to the field automatically.

PR #1167, opened 2016-05-17
[JBIDE-22314] Allow defining an image registry URL for a connection

jcantrill:
cc @fbricon

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/URLValidator.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/URLValidator.java
@@ -1,0 +1,56 @@
+/******************************************************************************* 
+ * Copyright (c) 2016 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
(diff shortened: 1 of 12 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/validator/URLValidator.java
```diff
@@ -0,0 +1,53 @@
...
+	public URLValidator(String urlType, boolean allowEmpty) {
+		invalidURLMessage = NLS.bind("Please provide a valid {0} URL", urlType);
+		this.allowEmpty  = allowEmpty;
+	}
+	
+	
+
+	@Override
+	public IStatus validate(Object in) {
+		String value = (String) in;
```
fbricon: There is a org.apache.commons.validator.routines.UrlValidator provided by commons-validator

jcantrill: Updated

fbricon:
I'll wait for the next sprint to merge this one, since there's nothing yet consuming the new registry url

fbricon:
So whatever I type in the text field, I keep getting the "Please provide a valid image registry url" error message and the wizard's finish button stays disabled

fbricon:
Ok so I was missing the https scheme (was using http).
- are we 100% sure no docker registry can be exposed through http? (think internal company docker registry)
- if yes, then the error message should be clearer, state only https connections are supported

fbricon:
Other problem, if you have 2 connections in the wizard, or try to create a new connection, whatever you type in the advanced section is not reset when switching the connection.

jcantrill:
@fbricon  updated PR to address your comment and issue

fbricon:
+1

PR #1171, opened 2016-05-19
[JBIDE-21967] corrected impl and typo in OpenShiftServerUtils#getDeploymentConfig

adietish:


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
```diff
@@ -404,27 +420,54 @@ private static String getProjectAttribute(String name, String defaultValue, IPro
 		return ServerUtils.getProjectAttribute(name, defaultValue, SERVER_PROJECT_QUALIFIER, project);
 	}
 	
-	public static IDeploymentConfig getDeploymentConfig(IServerAttributes attributes) {
-		IService service = getService(attributes);
-		//TODO use annotations instead of labels
+	/**
```
adietish: main change is here (changes in other classes are test).
changed the method to throw exceptions so that we can distinguish failure causes properly.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -84,20 +83,20 @@ public void launch(ILaunchConfiguration configuration, String mode,
...
 		}
 		ControllableServerBehavior beh = (ControllableServerBehavior) serverBehavior;
-		
-		beh.setServerStarting();
-		
 		IServer server = beh.getServer();
-		
-		IDeploymentConfig dc = OpenShiftServerUtils.getDeploymentConfig(server);
-		if (dc == null) {
+
```
adietish: consequence of the change in OpenShiftServerUtils#getDeploymentConfiguration.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
```diff
@@ -404,27 +420,54 @@ private static String getProjectAttribute(String name, String defaultValue, IPro
...
 		}
-		String dcName = ResourceUtils.getDeploymentConfigNameForPods(service.getPods());
-		if (dcName == null) {
-			return null;
+
+		List<IPod> pods = connection.getResources(ResourceKind.POD, service.getProject().getName());
+		List<IPod> servicePods = ResourceUtils.getPodsForService(service, pods);
+		if (pods == null
+				|| pods.isEmpty()) {
+			throw new CoreException(OpenShiftCoreActivator.statusFactory().errorStatus(
```
scabanovich: Should exception be throws if specifically servicePods.isEmpty() ?

adietish: +1

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/core/server/OpenShiftServerUtilsTest.java
```diff
@@ -111,26 +159,54 @@ private void assertNotContainsEapLikeKeywords(String text) {
...
+	}
+
+	@Test
+	public void should_return_deploymentconfig() throws CoreException {
+		// given
+		// when
+		IDeploymentConfig deploymentConfig = OpenShiftServerUtils.getDeploymentConfig(server);
+		// then
+		assertThat(deploymentConfig).isEqualTo(ResourceMocks.PROJECT2_DEPLOYMENTCONFIGS[2]);
+	}
```
scabanovich: Could be tests added that check expected CoreException when in retrieving the deployment config some resources failed to be found?

adietish: +1 even though the different exception cases only differ in exception msg.Trying to avoid fragile tests that assert exception message. Can assert substrings maybe to have assert exceptions and not have it too fragile

scabanovich: To avoid relying on strings, it can be custom exception extending CoreException with int detail field that can be set to one of constants. But I am ok with how it is done now.

PR #1183, opened 2016-05-25
[JBIDE-22432] If oc binary can't be located, then Internal Error is spawned in the server adapter wizard

jeffmaury:
- Add check and unit test

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
@@ -54,16 +54,19 @@ public class OCBinaryValidator {
 	 */
 	public Version getVersion(IProgressMonitor monitor) {
         Optional<Version> version = Optional.empty();
-		try {
-            ProcessBuilder builder = new ProcessBuilder(path, "version");
-            Process process = builder.start();
-            String line;
-            try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
-                while (!monitor.isCanceled() && (!version.isPresent()) && ((line = reader.readLine()) != null)) {
-                    version = parseVersion(line);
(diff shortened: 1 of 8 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinaryValidator.java
```diff
@@ -54,16 +54,19 @@ public OCBinaryValidator(String path) {
...
+            try {
+                ProcessBuilder builder = new ProcessBuilder(path, "version");
+                Process process = builder.start();
+                String line;
+                try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
+                    while (!monitor.isCanceled() && (!version.isPresent()) && ((line = reader.readLine()) != null)) {
+                        version = parseVersion(line);
+                    }
                 }
+            } catch (IOException e) {
```
adietish: Shouldnt we at least log this error to the console instead of muting it?

adietish:
other than maybe logging an IOException PR looks fine to me in terms of code.

adietish:
when testing it I dont have the NPE in Eclipse log any more but the warning on top of the wizard:
image
The warning reports an oc that's too old. I find this misleading, I think that if there's no oc at all we should state so explicitly.

fbricon:
I disagree. The warning tells you a recent version of the binary is required, not that the one you have is too old.

adietish:
@fbricon but the point still is that I have no oc binary at all, which is not clearly stated.

fbricon:
It tells you an OpenShift client is required, if that's not clear enough :-)

adietish:
your choice, I would expect something more clear.
And then I'd still log the swallowed IOException

fbricon:
After testing the PR, I now see the problem with the message about the missing oc. I actually thought it was the message from the preferences page, where it's obvious you need to add the binary.

So I now agree the message should be explicit it about the missing binary AND the message should have an error image, not a warning

PR #1190, opened 2016-05-26
JBIDE-22431 - rename cdk plugin and feature in UI elements

robstryker:
This is an initial PR.  I may need to externalize the strings before it's suitable for push, but this seems to cover all expected bases.

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/plugin.properties
+++ b/plugins/org.jboss.tools.openshift.cdk.server/plugin.properties
@@ -7,4 +7,4 @@ openshift.cdk.name = Container Development Kit Support
 openshift.cdk.description = Integration and support for the Red Hat Container Development Kit
 enterpriseProviderName=Red Hat JBoss Middleware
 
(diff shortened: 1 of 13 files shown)


plugins/org.jboss.tools.openshift.cdk.server/plugin.properties
```diff
@@ -1,10 +1,10 @@
...
+ProviderName = JBoss by Red Hat
 
-openshift.cdk.name = CDK Server Adapter
-openshift.cdk.description = A server adapter to control Red Hat CDK installations
+openshift.cdk.name = Container Development Kit Support
+openshift.cdk.description = Integration and support for the Red Hat Container Development Kit
 enterpriseProviderName=Red Hat JBoss Middleware
 
-launch_cdk_server=Launch CDK Server
+launch_cdk_server=Launch Red Hat Container Development Kit
```
alexeykazakov: Should be "Launch Container Development Environment"

alexeykazakov:
Looks good now. Please squash and push. Thanks!

PR #1191, opened 2016-05-27
[JBIDE-21420] - Delete more OpenShift projects at once

jeffmaury:
- Delete resources in batch
- Unify resources and projects


fbricon:
Can you add tests covering the multiple deletion please?

fbricon:
So when deleting a service (deployment node in OpenShift Explorer), I see the warning about related resources being deleted as well, but once I delete it, the related resources are still visible in the properties view

jeffmaury:
This a general warning as delete project and delete resource have been unified. Cascaded deletion is another topic.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenShiftResourceUIModel.java
```diff
@@ -19,30 +21,41 @@
...
-	private final IResource resource;
-	private final String tostring;
+	private IResource resource;
+	private String tostring;
 	private final Object parent;
+	private AtomicBoolean deleting = new AtomicBoolean(false);
+
 
-	public OpenShiftResourceUIModel(IResource resource, IResourcesUIModel parent){
+	public OpenShiftResourceUIModel(IResource resource, Object parent){
```
scabanovich: This is a step back from the typed model. We have 
	ResourcesUIModel (parent can be defined with a type parameter), 
	Deployment (parent IProjectAdapter), 
	OpenShiftProjectUIModel (parent IOpenShiftConnection), 
	OpenShiftResourceUIModel (parent is IResourcesUIModel)
Is it necessary for ResourcesUIModel to extend OpenShiftResourceUIModel, or rather we can implement IResourceUIModel?
In each class and where possible, in interfaces, method getParent() may be overridden to with correct parent return type.
As a maximum, IAncestorable should declare a type parameter for the parent.

jeffmaury: The rationale for the refactoring was that every UI element owns the mapped Openshift resource thus kept as the root of the hierarchy. Deployement kept a reference to the service where the mapped resource could be used. Please not that as we want to keep a flag for storing the deleting state, then we need a common class to store this state and so the the resource

scabanovich: There can be many solutions to the problem. One of them is to add one more abstract class like AbstractOpenShiftResourceUIModel and derive  OpenShiftResourceUIModel and ResourcesUIModel from it.

jeffmaury: OpenShiftResourceUIModel is the base class for all wrappers so why do you need another class ?

fbricon:
So, back to the cascaded deletion warning. I think the message should not be confusing, and should not mention related resources deletion, until the actual feature is implemented (which is uncertain to make it to this release at the moment)

jeffmaury:
The warning was more related to project deletion as it is now unified. I suggest we change the warning with something like "If you delete a project, then all resources for this project will be deleted".

fbricon:
unified warning works for unified behavior. Since projects currently behave differently, they should get a specific warning

jeffmaury:
Well as the handler processes a bunch of resources, you may have projects and resources. Unless we go back to separate actions (possibly using the same handler)

fbricon:
We can probably do an instanceof project check, at least until we unify cascade deletion

jeffmaury:
No because the warning dialog is displayed only once for all resources

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/DeleteResourceHandler.java
```diff
@@ -10,46 +10,87 @@
...
+            @Override
+            protected boolean shouldCancel(IStatus lastCompletedJobResult, int numberOfFailedJobs,
+                    int numberOfCanceledJobs) {
+                return false;
+            }
+	        
+	    };
+        try (Stream<IResourceUIModel> stream = Arrays.stream(uiResources)) {
+            stream.forEach(uiResource -> {
+                DeleteResourceJob job = OpenShiftJobs.createDeleteResourceJob(uiResource.getResource());
```
scabanovich: This approach reuses the available job for deleting one resource. But maybe, for the sake of better user experience, it might be nice to have one job for removing several resources with the progress monitor informing user of how many resources are already removed and which resource is currently being removed.

fbricon:
Squashed/Rebased/Merged as https://github.com/jbosstools/jbosstools-openshift/commit/d240058c46370dcecdbcd51016f7e7832d846081

PR #1192, opened 2016-05-27
[JBIDE-22197] - Icon in template Details of Local Template is not loaded for some templates file

jeffmaury:
- Just adding a PNG for Java
- No code modification as strack trace is generated once for a specific file and thus caught by Aeri


fbricon:
Can you add a screenshot of how it looks like in Eclipse?
I'm pretty sure the webconsole would use another icon, have you checked?

jeffmaury:
Eclipse UI:


Web Console:

jcantrill:
All icons we use that are displayed are the same as provided by the web console.  Most are sourced from the patterfly project and others were given to me.  I will see if i can source the actual file used and we can check it in as a resource to our repo.

fbricon:
Inspecting the web console, I found an openjdk icon:

jcantrill:
https://github.com/jbosstools/jbosstools-openshift/blob/master/plugins/org.jboss.tools.openshift.ui/icons/apps/openjdk.png

fbricon:
ok, so we probably just need to copy openjdk.png to java.png (we need both)

jeffmaury:
Agree. Wil do.

PR #1196, opened 2016-05-30
JBIDE-22463 - Update CDK code after API change in Docker tooling

xcoulon:
Signed-off-by: Xavier Coulon <xcoulon@redhat.com>

### DIFF
--- a/features/org.jboss.tools.openshift.egit.integration.feature/.project
+++ b/features/org.jboss.tools.openshift.egit.integration.feature/.project
@@ -10,6 +10,11 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
+			<name>org.sonarlint.eclipse.core.sonarlintBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
 	</buildSpec>
 	<natures>
 		<nature>org.eclipse.pde.FeatureNature</nature>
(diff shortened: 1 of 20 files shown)


mickaelistria:
Is this change backward compatible or does it require the newer version of Docker Tools?

fbricon:
Yup, probably need to add a version constraint on Docker Tools

mickaelistria:
If possible, yes, a version constraint would be fine.
Please ping again when this PR is ready, so we can merge the PR about TP and this one at the same time, avoiding a build failure.

fbricon:
So yeah the current code doesn't compile in Eclipse Neon RC2. So the version constraint is required along the TP change

xcoulon:
@fbricon  @mickaelistria I added a constraint on the Docker tooling to use 2.0.0. Please let me know if this is good enough, or if a specific timestamp is required.

nickboldt:
=2.0.0 should be fine unless there's a future API break between 2.0.0.x and 2.0.0.y, in which case SPECIFIC version of 2.0.0 is needed.

You could also protect against future problems (if you expect them) using [2.0.0,2.0.1) or [2.0.0,2.1.0) as appropriate.

When do you plan to merge this?

mickaelistria:
I merged the change in TP, please merge this one ASAP.

PR #1197, opened 2016-05-31
jbide-22437: NPE when clicking on a folder containing files without extension in browsed workspace

bdshadow:


plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/utils/UIUtils.java
```diff
@@ -485,7 +485,10 @@ public static ElementTreeSelectionDialog createFileDialog(String selectedFile,
             @Override
             public boolean select(Viewer viewer, Object parentElement, Object element) {
                 return element instanceof IContainer 
-                        || (element instanceof IFile && ((IFile)element).getFileExtension().equals(extension));
+                        || (element instanceof IFile 
+                        		&& extension != null
```
scabanovich: Javadoc to the method says that extension may be null. So the check before the comparison is correct, but what should be the logic for null? Could it be that if both extension and ((IFile)element).getFileExtension() are empty (null or empty string) that the method might return true? Or even if extension is null then filter has to accept all files?

bdshadow: @scabanovich if we look at some web features, for example when you want to upload an image to some web-site, you won't see files without extensions until you select "All files". So i think it must be the same way:
if extension is null (means no restrictions) then all files selected anyway.

If you agree, i'll change the PR

scabanovich: Currently, this method is only used to select 'json' files so that we are free to set the null case logic. What you suggest is ok to me. Please add it to Javadoc comments, too.

PR #1199, opened 2016-05-31
[JBIDE-19506] added tests for Connection

adietish:


adietish:
@scabanovich @jcantrill dare to review the cleanup I did in Connection?

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/core/connection/ConnectionTest.java
```diff
@@ -111,88 +136,122 @@ public void getResourceKindShouldCallClient() {
...
+		// then
+		assertNotEquals(connection, two);
 		assertFalse(two.credentialsEqual(connection));
 	}
 
 	@Test
-	public void shouldNotCredentialsEqualIfDifferentPassword() throws Exception {
-		Connection two = (Connection)connection.clone();
-		two.setPassword("passwordTwo");
+	public void should_not_credentialsEqual_if_different_password() throws Exception {
```
jcantrill: We should either, for consistency, use camelcase or underscore.  I prefer the later because its more readable, but the former is convension

jcantrill:
LGTM other then minor nit

PR #1202, opened 2016-06-01
[JBIDE-22454] Fix env var labels on new-app pages to be more explicit

jcantrill:
cc @fbricon @mlabuda 

Modified labels per issue to be more explicit.


fbricon:
LGTM

PR #1206, opened 2016-06-02
JBIDE-22441 - hard-code the docker registry url until servicemanager …

robstryker:
…plugin has a public api


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/CDKOpenshiftUtility.java
```diff
@@ -77,6 +78,8 @@ public IConnection createOpenshiftConnection(IServer server, ServiceManagerEnvir
 		((Connection)con).setRememberPassword(true);
 		
 
+		String defaultImageRegistry = "https://hub.openshift.rhel-cdk.10.1.2.2.xip.io ";
```
adietish: nitpicking: move to constant so that it can be more easily identified as hard-coded?

adietish:
+1 beside my nitpicking

PR #1212, opened 2016-06-03
[JBIDE-22241] - OpenShift Explorer: Cascade resource deletion

jeffmaury:
- Implement oc like processing for BuildConfig, DeploymentConfig and ReplicationController

Signed-off-by: Jeff MAURY jmaury@redhat.com


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/DeleteResourceHandler.java
```diff
@@ -48,16 +49,24 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
...
+	        boolean willDeleteSubResources = stream.anyMatch(resource -> Platform.getAdapterManager().getAdapter(resource.getResource(), ResourceProcessor.class).willCascadeDeleteLinkedResources());
+	        boolean confirm;
+	        
+	        String message = (resources.length > 1)?(willDeleteSubResources)?OpenShiftUIMessages.ResourceWithLinkedResourcesDeletionConfirmationN:OpenShiftUIMessages.ResourceDeletionConfirmationN
+                    :(willDeleteSubResources)?NLS.bind(OpenShiftUIMessages.ResourceWithLinkedResourcesDeletionConfirmation, resources[0].getResource().getName(), resources[0].getResource().getKind()):
+                                  NLS.bind(OpenShiftUIMessages.ResourceDeletionConfirmation, resources[0].getResource().getName(), resources[0].getResource().getKind());
+            if (willDeleteSubResources) {
+	            int status = new CheckboxMessageDialog(HandlerUtil.getActiveShell(event), OpenShiftUIMessages.ResourceDeletionDialogTitle, message, "Cascade delete").open();
+	            confirm = ((status & 0x01) == 0);
+	            willDeleteSubResources = (status & 0x02) == 2;
```
scabanovich: A minor suggestion, please replace 0x02 with CheckboxMessageDialog.CHECKBOX_SELECTED and 0x01 with Window.CANCEL.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
```diff
@@ -0,0 +1,102 @@
...
+                    String list = overlappingRCs[1].stream().map(res -> res.getName()).collect(Collectors.joining(","));
+                    throw new OpenShiftException(
+                            "Found overlapping replication controllers for replication controller %s: %s", rc.getName(),
+                            list);
+                } else if (overlappingRCs[0].size() == 1) {
+                    rc.setDesiredReplicaCount(0);
+                    rc = connection.updateResource(rc);
+                    while (rc.getCurrentReplicaCount() > 0) {
+                        Thread.sleep(1000);
+                        rc = connection.getResource(rc);
```
scabanovich: Is there a way for user to cancel it if something goes wrong and the expected result is not received for a long time?
And reporting to the progress monitor would be very welcome here.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
```diff
@@ -0,0 +1,102 @@
...
+                    overlappingRCS[1].add(rc);
+                }
+            }
+        }
+        return overlappingRCS;
+    }
+    
+    private boolean match(IReplicationController rc, IReplicationController replicationController) {
+        return match(rc.getReplicaSelector(), replicationController.getReplicaSelector()) |
+               match(replicationController.getReplicaSelector(), rc.getReplicaSelector());
```
scabanovich: Method match(Map,Map) can be called only once if it will be defined as 'Maps match if one of them is contained in the other'.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Find if maps matches. Maps matches if map1 is contained in map2
+     * @param map1 the first map
+     * @param map2 the second map
+     * @return the match state
+     */
+    private boolean match(Map<String, String> map1, Map<String, String> map2) {
```
scabanovich: Implementing definition 'Maps match if one of them is contained in the other', this check should be added:
if(map1.size() > map2.size()) {
   //Then map1 sure cannot be contained in map2, switch them. 
   //If sizes are equal, 'match' is equivalent to 'equal', no matter which of them is the first.
  Map<String, String> map = map1;
  map1 = map2;
  map2 = map;
}

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ResourceProcessor.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+    
+    /**
+     * Process the deletion of the resource.
+     * 
+     * @param connection the Openshift connection
+     * @param resource the resource to delete
+     * @param willDeleteSubResources if cascade delete
+     */
+    default void handleDelete(ConnectionsRegistry registry, Connection connection, IResource resource, boolean willDeleteSubResources) throws OpenShiftException {
```
scabanovich: Since this method may be long running while deleting multiple resources, I suggest adding parameter IProgressMonitor and reporting to it large pieces of job, like requesting resources, deleting resources, waits for result returned by the client, etc.

By the way, may user have the rigth to cancel? If it is highly undesirable, a warning may pop up explaining all disadvantages of cancelling, but with the option to cancel anyway available. A cancel may happen in different forms anyway, like Eclipse crash, network disconnection, electric power shut-down, coffee spilled, etc. If by some reason action takes so long that user may decide that it is better to have one of the mentioned above than wait, it is better to give him the option to cancel.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
fbricon: There's already something similar in ResourceUtils.containsAll

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/DeleteResourceHandler.java
```diff
@@ -48,16 +49,24 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
 			return OpenShiftUIActivator.statusFactory().cancelStatus("No resource selected that we can delete."); //$NON-NLS-1$
 		}
 		try (Stream<IResourceUIModel> stream = Arrays.stream(resources)) {
-	        boolean hasProject = stream.anyMatch(resource -> resource instanceof IProjectAdapter);
+	        boolean willDeleteSubResources = stream.anyMatch(resource -> Platform.getAdapterManager().getAdapter(resource.getResource(), ResourceProcessor.class).willCascadeDeleteLinkedResources());
```
fbricon: Sub resources for a Deployment/Service are not found:

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/DeleteResourceHandler.java
```diff
@@ -84,7 +94,7 @@ protected boolean shouldCancel(IStatus lastCompletedJobResult, int numberOfFaile
 	    };
```
scabanovich: I suggest to create and schedule one job for the bunch of resources, if there are more than 1 of them. The job, for instance, may iterate the stream and (1) check if monitor is cancelled, (2) report to monitor begin of task, (3) set 'deleting' to true, (4) explicitly call job.doRun(monitor) (let's make it public), (5) without a job listener, check status and set 'deleting' to false if ok.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
scabanovich: That is true, but it also should be called twice then. If match means that either of maps may contain the other, the new one-pass method may be added to ResourceUtils.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/DeleteResourceHandler.java
jeffmaury: OK will do

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
jeffmaury: Yes, this idea was more or less somewhere in my head but it get lost. Will do

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
jeffmaury: Don't get it. Can you detail ?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
jeffmaury: Will check

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ResourceProcessor.java
jeffmaury: Yes, user can cancel because it's part of a job as monitor is not given, report will be delayed. Will add ProgressMonitor to the interface.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/DeleteResourceHandler.java
jeffmaury: Service has no linked resource as per the oc client. See my spec in JIRA: https://issues.jboss.org/browse/JBIDE-22241?focusedCommentId=13247145&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13247145

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
scabanovich: By the way for matching Map<String,String> the single call

map1.entrySet().containsAll(map2.entrySet())

should work as expected (with switching maps as I suggested above).  Please try it.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
```diff
@@ -565,7 +591,7 @@ public String toString() {
 		throw e;
 	}
 
-	private <T extends IResource> List<T> retryList(String message, OpenShiftException e, String kind, String namespace){
+	private <T extends IResource> List<T> retryList(String message, OpenShiftException e, String kind, String namespace, Map<String, String> labels){
```
scabanovich: The implemtation of retryList does not use labels yet.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/utils/ReplicationControllerResourceProcessor.java
scabanovich: If map1.size() > map2.size() we can tell that map1 is not contained by map2 without calling match(map1,map2). So, below for implementation of match(Map,Map) I suggested switching maps so that map1.size() <= map2.size() and then just map2.entrySet().containsAll(map1.entrySet()) should do all the job.

mickaelistria:
testPR

jeffmaury:
Closing as it has been replaced by 3d4ae6a2ad154d28b6253a90306198c82260b9c1

PR #1217, opened 2016-06-08
Jbide 21828-2 Refactored Openshift 3 UI model to allow for precise merging of updates.

tsmaeder:


plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
```diff
@@ -150,54 +142,37 @@ public boolean remove(IConnection connection) {
...
-							
-						default:
-							break;
-						}
-					}
-				}catch(Exception e) {
-					return new Status(IStatus.ERROR,OpenShiftCommonCoreActivator.PLUGIN_ID,"Exception while notifying of a connection change",e);
-				}	
-				return Status.OK_STATUS;
+
```
jcantrill: Do we really want to remove this from running in a job?  This doesn't introduce any UI freeze issues or the like?

plugins/org.jboss.tools.openshift.common.ui/plugin.xml
```diff
@@ -258,8 +258,7 @@
 						<iterate
 							ifEmpty="false"
 							operator="and">
-							<instanceof
-								value="org.jboss.tools.openshift.common.core.connection.IConnection" />
+							<adapt	type="org.jboss.tools.openshift.common.core.connection.IConnection" />
```
jcantrill: For my own edification, does this mean the class has to impl adaptable so that it can return an IConnection? Or does the platform provide some magic that determines if it can be cast to that interface?

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
```diff
@@ -182,121 +171,52 @@ protected IStatus run(IProgressMonitor monitor) {
...
 	}
 
-	@Override
-	public void dispose() {
-		if(input != null) {
-			input.removeListener(connectionListener);
-		}
-	}
-	
-	private class ConnectionsRegistryListener implements IConnectionsRegistryListener{
```
jcantrill: Remind me why the explorer no longer listens to the registry directly?

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/OpenShiftExplorerView.java
```diff
@@ -187,15 +185,21 @@ private ISelectionChangedListener onSelectionChanged() {
...
-//					if (UIUtils.isFirstElementOfType(IDomain.class, selection)) {
-//						activate(DOMAIN_CONTEXT);
-//					} else if (UIUtils.isFirstElementOfType(IApplication.class, selection)) {
-//						activate(APPLICATION_CONTEXT);
-//					} else if (UIUtils.isFirstElementOfType(ExpressConnection.class, selection)) {
-//						// must be checked after domain, application, adapter may convert
-//						// any resource to a connection
-						activate(CONNECTION_CONTEXT);
-//					}
+					// if (UIUtils.isFirstElementOfType(IDomain.class,
```
jcantrill: feel free to delete if unused.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/OpenShiftExplorerView.java
```diff
@@ -150,30 +151,27 @@ private void showConnectionsOrExplanations(Control connectionsPane, Control expl
...
 	}
-	
+
 	private static class OpenShiftExplorerContextsHandler extends Contexts {
 
 		private static final String CONNECTION_CONTEXT = "org.jboss.tools.openshift.explorer.context.connection";
-//		private static final String APPLICATION_CONTEXT = "org.jboss.tools.openshift.explorer.context.application";
-//		private static final String DOMAIN_CONTEXT = "org.jboss.tools.openshift.explorer.context.domain";
-		
+		// private static final String APPLICATION_CONTEXT =
```
jcantrill: same here...feel free to delete if unused

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/IOpenShiftConnection.java
```diff
@@ -22,7 +24,7 @@
  * @author jeff.cantrill
  *
  */
-public interface IOpenShiftConnection {
+public interface IOpenShiftConnection extends IConnection {
```
jcantrill: This interface intentionally did not extend IConnection because things like 'username' and 'password' are not necessarily relevant to v3 openshift connections; they rely specifically on a token.  Additionally, I thought there were advantages in some of our consumers that facilitated testing without having to provide an IConnection.  Maybe it is immaterial now.

plugins/org.jboss.tools.openshift.ui/META-INF/MANIFEST.MF
```diff
@@ -60,7 +60,7 @@ Export-Package: org.jboss.tools.openshift.internal.ui;x-friends:="org.jboss.tool
  org.jboss.tools.openshift.internal.ui.explorer;x-friends:="org.jboss.tools.openshift.test",
  org.jboss.tools.openshift.internal.ui.handler;x-friends:="org.jboss.tools.openshift.test",
  org.jboss.tools.openshift.internal.ui.job;x-friends:="org.jboss.tools.openshift.test",
- org.jboss.tools.openshift.internal.ui.models;x-friends:="org.jboss.tools.openshift.test",
+ org.jboss.tools.openshift.internal.ui.models,
```
jcantrill: does the test project not need to be a friend still?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -44,197 +41,157 @@
...
+		LoadingStub stub = makeStub(parentElement);
+		stub.add(e);
+		asyncExec(()-> refreshViewer(stub));
+	}
+
+	private LoadingStub makeStub(Object parentElement) {
+		LoadingStub stub = stubs.get(parentElement);
+		if (stub == null) {
+			stub= new LoadingStub();
+			stubs.put(parentElement, stub);
```
jcantrill: Any chance of a race condition here where concurrent threads try to add entries and one ends up replacing the other?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -44,197 +41,157 @@
...
-			} else if (parentElement instanceof Deployment) {
-				Deployment deployment = (Deployment) parentElement;
-				return handleDeployment(deployment).toArray();
 			}
-		
-		} catch (OpenShiftException e) {
-			addException(parentElement, e);
+		} else if (parentElement instanceof ProjectWrapper) {
+			ProjectWrapper project = (ProjectWrapper) parentElement;
+			if (project.load(e -> {
```
jcantrill: Looks like there is an opportunity to refactor some loading logic into a common method with the ConnectionWraper

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/DeployImageHandler.java
```diff
@@ -69,15 +69,9 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
...
-				IResourceUIModel resourceModel = UIUtils.getFirstElement(selection, IResourceUIModel.class);
-				while(resourceModel != null && project == null) {
-					if(resourceModel.getResource() instanceof IProject) {
-						project = (IProject)resourceModel.getResource();
-					} else if(resourceModel.getParent() instanceof IResourceUIModel) {
-						resourceModel = (IResourceUIModel)resourceModel.getParent();
-					} else {
-						resourceModel = null;
-					}
+				IResource resource = UIUtils.getFirstElement(selection, IResource.class);
```
jcantrill: We can do this because the types are adaptable to IResource?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/StartBuildHandler.java
```diff
@@ -52,29 +49,11 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
 	}
 	
 	public IResource getBuildableResource(ISelection selection, String source) {
-		IResource buildAble = null;
-		IAdaptable adaptable = UIUtils.getFirstElement(selection, IAdaptable.class);
-		if(adaptable != null) {
-			buildAble = adaptable.getAdapter(IBuildConfig.class);
-			if (buildAble == null) {
```
jcantrill: Looks like you lost the ability to initiate a build from a buildconfig which is probably our first choice.  A user would mostly start a new build then clone a previous one

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
```diff
@@ -0,0 +1,63 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: Missing header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
```diff
@@ -0,0 +1,63 @@
+package org.jboss.tools.openshift.internal.ui.models;
+
+import org.eclipse.core.runtime.IAdaptable;
+
+public abstract class AbstractOpenshiftUIElement<R, P extends IOpenshiftUIElement<?>> implements IAdaptable, IOpenshiftUIElement<P> {
```
jcantrill: Comments would be nice with what type parameters mean

jcantrill: Can we further restrict the generic type then wildcard?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
```diff
@@ -0,0 +1,63 @@
...
+		if (o == this) {
+			return true;
+		}
+		AbstractOpenshiftUIElement<?, ?> other= (AbstractOpenshiftUIElement<?, ?>) o;
+		return wrapped.equals(other.getWrapped()) && getParent().equals(other.getParent());
+	}
+	
+	@Override
+	public int hashCode() {
+		return wrapped.hashCode();
```
jcantrill: I dont profess to understand fully how equals and hashcode need to be implemented but should this also include some calculation with the parent given thats what equals also uses?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractResourceWrapper.java
```diff
@@ -0,0 +1,23 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -0,0 +1,244 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -0,0 +1,244 @@
...
+				IResource oldResource = resourceCache.getCachedVersion((IResource) oldValue);
+				if (oldResource != null) {
+					ProjectWrapper projectWrapper = findProjectWrapper(oldResource);
+					// it's a remove
+					handleRemove(projectWrapper, oldResource);
+				}
+			} else {
+				// old value == null, new value == null, ignore
+				OpenShiftUIActivator.log(IStatus.WARNING, "old and new value are null",
+						new RuntimeException("Warnign origing"));
```
jcantrill: spelling of 'warning'

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -0,0 +1,244 @@
...
+		return false;
+	}
+
+	void startLoadJob(ProjectWrapper projectWrapper, IExceptionHandler handler) {
+		new Job("load project") {
+
+			@Override
+			protected IStatus run(IProgressMonitor monitor) {
+				try {
+					WatchManager.getInstance().startWatch(projectWrapper.getResource());
```
jcantrill: I'm not certain how we used to do this, but this will start a watch while also loading resources in the following lines.  I assume we have some logic to handle duplicates?  This is done this way I suspect to avoid gaps where resources get added between load and watch start?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -0,0 +1,244 @@
...
+					Collection<IResource> resources = new HashSet<>();
+					for (String kind : RESOURCE_KINDS) {
+						resources.addAll(projectWrapper.getResource().getResources(kind));
```
jcantrill: I don't think we did this previously, but we should be retrieving resources from the connection instead of the project.  This ensures we properly handle the auth logic and token expiration.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -0,0 +1,244 @@
...
+
+	void startLoadJob(IExceptionHandler handler) {
+		new Job("load project") {
+
+			@Override
+			protected IStatus run(IProgressMonitor monitor) {
+				try {
+					IOpenShiftConnection connection = getConnection();
+					List<IResource> projects = connection.getResources(ResourceKind.PROJECT);
+					initWith(projects);
```
jcantrill: Can we have a race condition here? I see we set the state to loaded but I don't see any other locks.  Could we have multiple threads loading at the same time?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/IElementListener.java
```diff
@@ -0,0 +1,5 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/IExceptionHandler.java
```diff
@@ -0,0 +1,14 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/IOpenshiftUIElement.java
```diff
@@ -0,0 +1,36 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/LoadingState.java
```diff
@@ -0,0 +1,5 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
```diff
@@ -0,0 +1,189 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
```diff
@@ -0,0 +1,189 @@
...
+					if (connections.containsKey(connection)) {
+						return;
+					}
+					connections.put((IOpenShiftConnection) connection,
+							new ConnectionWrapper(OpenshiftUIModel.this, (IOpenShiftConnection) connection));
+				}
+				fireChanged();
+			}
+		};
+		Collection<IConnection> allConnections = registry.getAll();
```
jcantrill: registry has a getAll(Class) which will mitigate the need to run through the following sync block

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
```diff
@@ -0,0 +1,189 @@
...
+			for (IConnection connection : allConnections) {
+				if (connection instanceof IOpenShiftConnection) {
+					connections.put((IOpenShiftConnection) connection,
+							new ConnectionWrapper(OpenshiftUIModel.this, (IOpenShiftConnection) connection));
+				}
+			}
+		}
+		registry.addListener(listener);
+	}
+	public static boolean isOlder(IResource oldResource, IResource newResource) {
```
jcantrill: I don't know if this is new or left over but what if these resources are not different versions of the same resource (e.g. one is a build and the other is a route, or one is build/foo and the other is build/bar).  We should probably handle that case.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
```diff
@@ -0,0 +1,189 @@
...
+				public void run() {
+					dispatchChange(source);
+				}
+			});
+		}
+	}
+
+	protected void dispatchChange(IOpenshiftUIElement<?> source) {
+		Collection<IElementListener> copy = new ArrayList<>();
+		synchronized (listeners) {
```
jcantrill: Do we have equivalent code (or does it matter) if we use synchronized collections instead of synchronized blocks.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	@Override
+	public void refresh() {
+			Map<IOpenShiftConnection, ConnectionWrapper> updated = new HashMap<>();
+			boolean changed = false;
+			synchronized (connections) {
+				HashMap<IOpenShiftConnection, ConnectionWrapper> oldWrappers = new HashMap<>(connections);
+				connections.clear();
+				for (IConnection r : ConnectionsRegistrySingleton.getInstance().getAll()) {
```
jcantrill: registry.getAll(Class)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ProjectWrapper.java
```diff
@@ -0,0 +1,115 @@
...
+public class ProjectWrapper extends ResourceContainer<IProject, ConnectionWrapper> {
+	private AtomicReference<LoadingState> state = new AtomicReference<LoadingState>(LoadingState.INIT);
+
+	public ProjectWrapper(ConnectionWrapper parent, IProject resource) {
+		super(parent, resource);
+	}
+
+	public Collection<AbstractResourceWrapper<?, ?>> getResources() {
+		synchronized (childrenLock) {
+			return containedResources.values();
```
scabanovich: The returned collection is not synchronized and using it may result in concurrency problems.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceContainer.java
```diff
@@ -0,0 +1,30 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceWrapper.java
```diff
@@ -0,0 +1,11 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ServiceResourceMapper.java
```diff
@@ -0,0 +1,163 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
tsmaeder: This is actually an essential change. Putting the notification in a Job will potentially reorder changes (jobs run in parallel). For example, I had the case where an update notification was arriving after a delete notification (even though the delete happened last). If a client takes a long time reacting to a change notification it is up to the client to move the work to a background thread. Do you have a particular scenario in mind?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ServiceWrapper.java
```diff
@@ -0,0 +1,60 @@
+package org.jboss.tools.openshift.internal.ui.models;
```
jcantrill: header

plugins/org.jboss.tools.openshift.common.ui/plugin.xml
tsmaeder: Good question....I don't know.

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/ui/models/OpenShiftProjectCacheTest.java
```diff
@@ -21,145 +20,123 @@
...
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
+import org.mockito.Mockito;
 import org.mockito.runners.MockitoJUnitRunner;
 
 import com.openshift.restclient.ResourceKind;
 import com.openshift.restclient.model.IProject;
 
 @RunWith(MockitoJUnitRunner.class)
 public class OpenShiftProjectCacheTest {
```
jcantrill: We should rename this test to an appropriate name of the class its testing now

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/ui/models/ResourceEquality.java
```diff
@@ -0,0 +1,27 @@
+package org.jboss.tools.openshift.test.ui.models;
```
jcantrill: header

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/ui/models/ResourceEquality.java
```diff
@@ -0,0 +1,27 @@
+package org.jboss.tools.openshift.test.ui.models;
+
+import com.openshift.restclient.model.IResource;
+
+public class ResourceEquality {
```
jcantrill: utility class?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceContainer.java
```diff
@@ -0,0 +1,30 @@
...
+public abstract class ResourceContainer<R extends IResource, P extends IOpenshiftUIElement<?>>
+		extends AbstractResourceWrapper<R, P> {
+	protected Map<IResource, AbstractResourceWrapper<?, ?>> containedResources = new HashMap<>();
+	protected Object childrenLock = new Object();
+
+	public ResourceContainer(P parent, R resource) {
+		super(parent, resource);
+	}
+
+	public Collection<AbstractResourceWrapper<?, ?>> getResourcesOfKind(String kind) {
```
scabanovich: For the performance sake,
1) a private map (kind -> resources of the kind) can be kept and updated when containedResources are changed, scanning the full map for each request is costly;
2) One more method getResourcesOfKindAsStream() may return stream object for further filtering, picking the first match, etc, as in many cases the returned collection is only used to create new stream, or take the first element, or evaluate size.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/OpenShiftExplorerView.java
tsmaeder: Each great journey starts with a small step...

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/IOpenShiftConnection.java
tsmaeder: I see the convienience argument, but if there are different kinds of connection, shouldn't we remove the not-common API from the common superinterface? In other news, interfaces should tell the intention (the TRUTH!) of a structure, not a convenient lie. 
But honestly it was just a PITA to cast ;-)

jcantrill: Agreed, it should, but you have to consider the history; the v3 code base evolved from v2 which is a completely different model.  There was(is?) a desire for the explorer to support both and as such steps were taken to accommodate not disrupting the v2 code base.  This change is probably not that big of an issue

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
tsmaeder: I think you are correct. handleException() will be called from a background thread, so this is not clean. Bugger! I don't think it will have negative effects, though.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
tsmaeder: Working on it.....

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/DeployImageHandler.java
tsmaeder: Yes.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
tsmaeder: Working on it.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
tsmaeder: What do you mean?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceCache.java
```diff
@@ -0,0 +1,177 @@
...
+	 * 
+	 * @param resource
+	 * @return
+	 */
+	private String getCacheKey(IResource resource) {
+		return getCacheKey(resource.getKind(), resource.getName());
+	}
+
+	private String getCacheKey(String kind, String name) {
+		return NLS.bind("{0}/{1}", new Object[] { kind, name });
```
scabanovich: This is a private method and it is quite heavily used. It should be fast. Now, NLS.bind here is not efficient, it is ok for messages/labels in UI and reporting, but outrageously slow for the hard use. I would suggest just 
kind + "/" + name

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
tsmaeder: Since only few elements will have the same wrapped element but a different parent, this will add little to the quality of the hash function.
Note that returning 42 would be a very bad, but perfectly legal hash function.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
tsmaeder: Either the "getResources()" or the notification will arrive first. Updates to resources that are older than what we already have will be ignored. The only problem I can see is that if we get a delete notification and then get the resource form "getResources()" afterwards. Not sure how to handle that..

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
tsmaeder: If there is a "good way" and a "bad way" to get resources, why to we even have the "bad way"? If the APIs have different semantics, we should document it.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
tsmaeder: No, loading only starts when the state is "INIT". CompareAndSet ensures that only one thread will proceed to loading.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -44,197 +41,157 @@
...
+
+	private OpenshiftUIModel model;
+	private IElementListener listener;
+	private StructuredViewer viewer;
+	private Map<Object, BaseExplorerContentProvider.LoadingStub> stubs= new HashMap<Object, BaseExplorerContentProvider.LoadingStub>();
+
 	public OpenShiftExplorerContentProvider() {
-		super();
-		cache.addListener(this);
+		this(new OpenshiftUIModel());
```
scabanovich: Why OpenshiftUIModel remains hidden inside OpenShiftExplorerContentProvider. Can it be of use for other UI components/actions? Let's have a singleton instance for it.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
tsmaeder: sync is still needed, but I can get rid of the instanceof, but only if IOpenshiftConnection extends IConnectoin (otherwise the generic types don't compile)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
tsmaeder: This should be package visible. Callers ensure that it's the same resource before calling isOlder().

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
tsmaeder: If we want the collection to remain the same across multiple operations (see  removeListener()), then it makes a difference.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ProjectWrapper.java
tsmaeder: thx, good catch.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceCache.java
tsmaeder: It don't think it is heavily used. Perhaps hundreds of times each time a resource. What makes you think otherwise? 
In other news, I agree it's overkill.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceContainer.java
tsmaeder: I know this is not optimal. I posit it makes no perceivable difference to performance.

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/ui/models/ResourceEquality.java
tsmaeder: Yes it is. Would you propose a different name?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/StartBuildHandler.java
tsmaeder: Gahh....TODO still to do.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
jcantrill: I'm not super dogmatic about singletons, but I try to generally avoid them for testing reasons.  Is there a current need to expose this as a singleton?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
jcantrill: can it be restricted to something like IOpenShiftUIElement<IResource>

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
jcantrill: Hmm... so that is 'yes the calculation should include the parent' or 'no it is not required'?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
jcantrill: "cross finger" there is logic in the server to handle resources being deleted and someone trying to load a resource kind.  I would think you could get a delete notification and the resource would not then be in the collection that is returned.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
jcantrill: Its a difference in the rest client does not have logic to handle reconnect and requiring a user to refresh their token, but a connection does.  Maybe its a poor design to include the method as part of a project.  It was added their for convenience but is probably worth deprecating.  To this end, we have talked about the ability to add a caching layer to the connection which this pattern  you are using (and is used elsewhere) currently would bypass that cache

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceCache.java
jcantrill: I prefer String.format but was told to use NLS.  Even so, string concatination may be better then either NLS or string.format

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/ui/models/ResourceEquality.java
jcantrill: I do not...is it usable elsewhere?  Could be part of the util package?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
```diff
@@ -0,0 +1,189 @@
...
+					if (!(r instanceof IOpenShiftConnection)) {
+						return;
```
scabanovich: Probably 'continue', but the check will not be needed, when registry.getAll(Connection.class) is used.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
```diff
@@ -0,0 +1,189 @@
...
+
+			if (changed) {
+				fireChanged();
+			}
+
+			updated.keySet().forEach(r -> {
+				ConnectionWrapper wrapper = updated.get(r);
+				wrapper.updateWith(r);
+			});
+		for (ConnectionWrapper connection : connections.values()) {
```
scabanovich: Either should be in synchronized (connections), or just use getConnections()

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
scabanovich: @jcantrill , hashCode() must return same int value for two equal objects, but it may return same int value for two different objects. So, though usually hashcode impl uses all that is compared in equals, it does not have to. That depends on the resulting performance.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
scabanovich: Not very much yet, but there are calls Connection.getResources() in wizards, handlers, utilities and some of them might be replaced with queries to the model to avoid extra requests for resources to the client. 
For testing reasons, constructor may take not model but model factory, and tests will create content provider with a specific factory.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceCache.java
```diff
@@ -0,0 +1,177 @@
...
+	/**
+	 * 
+	 * @param resource
+	 * @return true if cached; false otherwise
+	 */
+	public boolean add(IResource resource) {
+		if (resource == null)
+			return false;
+		synchronized (cache) {
+			if (getCachedVersion(resource) != null) {
```
scabanovich: Can cached version happen to be another obsolete instance getCachedVersion(resource) != resource?
Or that is prevented with duly work of remove and update methods?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ServiceResourceMapper.java
```diff
@@ -0,0 +1,163 @@
...
+				if (isRelated(s, (IRoute)resource)) {
+					result.add(resource);
+				}
+			}
+		});
+		return result;
+	}
+
+	private static boolean isRelated(IService s, IRoute resource) {
+		return resource.getServiceName().equals(s.getName());
```
scabanovich: This method reimplements ResourceUtils.areRelated(IRoute route, IService service) without checks. If this implementation is safe and correct, should not it replace the method in ResourceUtils? Other 'isRelated' methods could go to ResourceUtils as well.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ServiceResourceMapper.java
```diff
@@ -0,0 +1,163 @@
...
+
+	private static Collection<IBuild> getRelatedBuilds(Collection<IResource> resources, Collection<String> dcImageRefs,
+			Collection<IBuildConfig> buildConfigs) {
+
+		Collection<IBuild> result = new HashSet<>();
+		Collection<String> bcNames = buildConfigs.stream().map(bc -> bc.getName()).collect(Collectors.toSet());
+		resources.forEach(r -> {
+			if (r instanceof IBuild) {
+				IBuild build = (IBuild) r;
+				if (bcNames.contains(r.getLabels().get(OpenShiftAPIAnnotations.BUILD_CONFIG_NAME)) || dcImageRefs.contains(imageRef(build))) {
```
scabanovich: @jcantrill , I found that StartBuildHandler.loadBuildConfig(build) takes the same way to get build config name from build, but in Deployment it was always received as an annotation value. Are the two ways equivalent?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceContainer.java
scabanovich: Well, it is a respected approach, to do optimization to some piece of code only after it is proven to be the cause of performance problems. As I was told, the expected number of OpenShift resources is not large, ok then. I should be excused for the habit to stop at each snippet because of the experience of working on quite a few models that required high performance on huge amount of objects.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
tsmaeder: I will use the connection to fetch resources. I have strong opinions against having a hidden ("transparent") cache, which I would be happy to discuss offline.

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/ui/models/ResourceEquality.java
tsmaeder: I just use it here to enable my handmade mocks. Unfortunately, Mockito does not allow for equals() to be mocked.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceContainer.java
tsmaeder: I meant no offense and you're right to point it out. However, my goal was to simplify as much as possible.
My approach is usually to think about performance in terms of stuff that is intiated by the user (low frequency) or by a machine (high frequency). Fetching children happens in user time (expanding a viewer, for example)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
tsmaeder: Oopsie!

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
tsmaeder: thx, good catch.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceCache.java
tsmaeder: Yes, this should be ensured outside the cache class.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ServiceResourceMapper.java
tsmaeder: thx, I was not aware of the method in ResourceUtils

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/StartBuildHandler.java
tsmaeder: Wouldn't the correct way to check be to test for IBuildTriggerable on the resource (instead of going for the resource kind?)

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
adietish: I believe that the basic observation with keeping the update sequence intact is a very valid point - using jobs wont allow this since there's no assurance on when jobs are schedules and they can also get reordered in a pool.
The objection with possible UI freezes is also valid, but then ensuring no freeze happens (aka use jobs) should be pushed to the reciever of the notification I believe

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
```diff
@@ -150,54 +142,37 @@ public boolean remove(IConnection connection) {
...
 		return recentConnection;
 	}
-	
+
 	@SuppressWarnings("unchecked")
 	public <T extends IConnection> T getRecentConnection(Class<T> clazz) {
-		if (recentConnection == null
-				|| clazz == null
-				|| !clazz.isAssignableFrom(recentConnection.getClass())) {
+		if (recentConnection == null || clazz == null || !clazz.isAssignableFrom(recentConnection.getClass())) {
```
adietish: I'd love if we could keep every condition in this "if" in its own line, it helps in reading the code and grasping every single condition.

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
```diff
@@ -209,27 +184,26 @@ public IConnection getByUrl(ConnectionURL connectionUrl) {
...
 		return connectionsByUrl.get(connectionUrl);
 	}
-	
+
 	@SuppressWarnings("unchecked")
 	public <T extends IConnection> T getByUrl(ConnectionURL connectionUrl, Class<T> clazz) {
 		IConnection connection = getByUrl(connectionUrl);
-		if (connection != null
-				&& !clazz.isAssignableFrom(connection.getClass())) {
+		if (connection != null && !clazz.isAssignableFrom(connection.getClass())) {
```
adietish: see above

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
```diff
@@ -209,27 +184,26 @@ public IConnection getByUrl(ConnectionURL connectionUrl) {
...
+	 * 
 	 * @return
 	 */
 	@SuppressWarnings("unchecked")
 	public <T extends IConnection> Collection<T> getAll(Class<T> clazz) {
 		List<T> connections = new ArrayList<>();
 		for (IConnection connection : connectionsByUrl.values()) {
-			if (connection != null
-					&& clazz.isAssignableFrom(connection.getClass())) {
+			if (connection != null && clazz.isAssignableFrom(connection.getClass())) {
```
adietish: see above

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
```diff
@@ -239,31 +213,33 @@ public IConnection getByUrl(ConnectionURL connectionUrl) {
...
 	private void addPropertyChangeListener(IConnection connection) {
 		if (!(connection instanceof ObservablePojo)) {
 			return;
 		}
-		
-		((ObservablePojo) connection).addPropertyChangeListener(connectionListener);;
-		
+
+		((ObservablePojo) connection).addPropertyChangeListener(connectionListener);
+		;
```
adietish: remove this unnecessary ";"?

plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
```diff
@@ -239,31 +213,33 @@ public IConnection getByUrl(ConnectionURL connectionUrl) {
...
 	private void removePropertyChangeListener(IConnection connection) {
 		if (!(connection instanceof ObservablePojo)) {
 			return;
 		}
-		
-		((ObservablePojo) connection).removePropertyChangeListener(connectionListener);;
-		
+
+		((ObservablePojo) connection).removePropertyChangeListener(connectionListener);
+		;
```
adietish: remove this unnecessary ";"?

plugins/org.jboss.tools.openshift.common.ui/plugin.xml
adietish: There are 2 ways for this adaptation to happen: either the object itself implements IAdaptable or the platform has an IAdapterFactory registered that's able to produce an IConnection instance for a given object. Both conditions are not met yet, but could/should be added. Using this currently wont hurt since <adapt /> also checks for type identity (instanceof - see here: https://wiki.eclipse.org/Platform_Expression_Framework#adapt).
@tsmaeder could you please verify this?

plugins/org.jboss.tools.openshift.common.ui/plugin.xml
```diff
@@ -279,8 +278,7 @@
 					<with variable="selection">
 						<and>
 							<iterate ifEmpty="false" operator="and">
-								<instanceof
-									value="org.jboss.tools.openshift.common.core.connection.IConnection" />
+								<adapt type="org.jboss.tools.openshift.common.core.connection.IConnection" />
```
adietish: see above

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
```diff
@@ -26,63 +24,52 @@
...
 	 * @param parentElement
 	 * @return
 	 */
 	protected abstract Object[] getChildrenFor(Object parentElement);
 
-	
 	@Override
 	public void inputChanged(Viewer viewer, Object oldInput, Object newInput) {
 		this.viewer = (TreeViewer) viewer;
-		if(input != null && connectionListener != null) {
```
adietish: This wont happen? Is this code really not used? We should at least test if wont happen in v2 nor in v3.

plugins/org.jboss.tools.openshift.common.ui/plugin.xml
tsmaeder: Quote from your linked page:

> Checks if the evaluated object is either an instance of, or adapts to the given class. 
so yes, if the object is an instance of X, <adapt type=x> will evaluate to true

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
```diff
@@ -182,121 +171,52 @@ protected IStatus run(IProgressMonitor monitor) {
...
-					if(object != null) {
-						viewer.refresh(object);
-					}else {
-						viewer.refresh();
+		Control control = viewer.getControl();
+		if (!control.isDisposed()) {
+			control.getDisplay().asyncExec(new Runnable() {
+				@Override
+				public void run() {
+					if (!viewer.getControl().isDisposed()) {
```
adietish: use control instead of viewer.getControl()?

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
adietish: yes, please explain, I miss this insight, too, would love to understand this design choice.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
```diff
@@ -182,121 +171,52 @@ protected IStatus run(IProgressMonitor monitor) {
...
 		public Object[] getChildren() {
 			return children.toArray();
 		}
-		
+
 		public void addChildren(Object[] children) {
 			this.children.addAll(Arrays.asList(children));
 		}
 
+		public boolean hasChildren() {
```
adietish: oh, good catch.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
```diff
@@ -182,121 +171,52 @@ protected IStatus run(IProgressMonitor monitor) {
...
+			return !children.isEmpty();
```
adietish: oh, good catch.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
tsmaeder: No, this change is crap. reverted....

plugins/org.jboss.tools.openshift.express.ui/src/org/jboss/tools/openshift/express/internal/ui/explorer/ExpressExplorerContentProvider.java
```diff
@@ -25,7 +30,34 @@
  * Content Provider for OpenShift 2 content
  */
 public class ExpressExplorerContentProvider extends BaseExplorerContentProvider {
-
+	private ConnectionsRegistry input;
+	private IConnectionsRegistryListener connectionListener = new ConnectionsRegistryListener();
+	private static final Collection<String> PROPERTY_BLACKLIST = Collections.unmodifiableList(Arrays.asList("token"));
```
adietish: could we use some constant (I see it living as private constant in Connection, maybe we should have it public and moved to IOpenShiftConnction?) since we here want to black list some notification that's being sent out by someone and we might get out-of-sync if the notifier gets refactored?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
bdshadow: agree with @scabanovich  - [javadoc for object hashCode](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()) - "However, the programmer should be aware that producing distinct integer results for unequal objects may improve the performance of hash tables."
So @jcantrill - 'no it is not required', but it's better to do this

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
bdshadow: should't we leave synchronized on viewer here?

tsmaeder: As a general rule, UI elements should not be manipulated outside the UI thread. The viewer variable is only ever changed on the UI thread. Which leaves stuff like "viewer.getControl()", etc. But again, I believe that these variables are constant after initialization, so no problem there.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceCache.java
scabanovich: Then, inside the case (getCachedVersion(resource) != null) I would add
if (getCachedVersion(resource) != resource) throw new IllegalStateException();
That may save some debugging time if something is wrong outside the cache class.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/explorer/BaseExplorerContentProvider.java
adietish: I see that this gets overriden by the v2 (https://github.com/jbosstools/jbosstools-openshift/pull/1217/files#diff-fec63e7aa9d5165cb82c461b22a3ae5bR75) and v3 (https://github.com/jbosstools/jbosstools-openshift/pull/1217/files#diff-fe25f5ebed5fcd00f5f86c6e04e1aba0R166) content provider subclasses. This portion of the code matching what's now in v2, while v3

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -44,197 +41,157 @@
  * 
  * @author jeff.cantrill
  */
-public class OpenShiftExplorerContentProvider extends BaseExplorerContentProvider 
-	implements PropertyChangeListener, IProjectCache.IProjectCacheListener {
-	
+public class OpenShiftExplorerContentProvider implements ITreeContentProvider {
```
adietish: ah, here's one of the main design changes: v3 content provider does not extend the base content provider any longer.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/StartBuildHandler.java
jcantrill: +1

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ResourceContainer.java
scabanovich: Yes, the new model looks very good, the goal is certainly achieved. 
In determining if the method is called with low or high frequency, the model listens to changes to the remote project, they may be multiple on a single user action,model will update on each remote change, then refreshing a branch in explorer on each change event, UI will request for children for all nodes in the branch (maybe I miss something here in my reasoning). All this may be harmless for small models, and maybe optimization would only add unneeded complexity here. However, to be sure of it, I would advise to add a performance test for measuring time with models of, say, 10, 100, and 1000 resources, and computing the dependency of time on size. If it turns to be cubic growth or faster, the test should always fail, if it is square growth, the guess of real world model size should be taken into account to decide whether to fail or not.

tsmaeder: The model will not refresh anything, unless the the user explicitly ask to do so. Otherwise, we are getting change notifications for single resources. Since we show resources as contained inside Services, a change of a label, for example might move a number of resources to another service. Yes, recalculating the containement hierarchy on each resource change may be expensive, but it will only touch stuff that is in-memory and should be reasonably fast. Also, the changes are merged in the background and only notifications about changes in UI model are notified on the UI thread.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ProjectWrapper.java
fbricon: we should prolly return an immutable collection

tsmaeder:
I have incorporated a bunch of your comments, and cleaned up.

plugins/org.jboss.tools.openshift.ui/META-INF/MANIFEST.MF
```diff
@@ -60,7 +60,7 @@ Export-Package: org.jboss.tools.openshift.internal.ui;x-friends:="org.jboss.tool
...
  org.jboss.tools.openshift.internal.ui.portforwading;x-friends:="org.jboss.tools.openshift.test",
  org.jboss.tools.openshift.internal.ui.property;x-friends:="org.jboss.tools.openshift.test",
  org.jboss.tools.openshift.internal.ui.property.build;x-friends:="org.jboss.tools.openshift.test",
```
fbricon: Package 'org.jboss.tools.openshift.internal.ui.propertytester' does not exist in this plug-in

fbricon:
So after opening the openshift explorer, I see 2 immediate issues:
- pods are missing (webconsole says they're running)
- service shows labels, instead of default route URL

tsmaeder:
Thx Fred, fixed.

fbricon:
Seems we can no longer Edit connections (right click on OS3 connection > Edit... is missing)

fbricon:
If I edit the Route on a service, in the Properties view, say replacing aloha-hi.rhel-cdk.10.1.2.2.xip.io by aloha.10.1.2.2.xip.io, once I save the route configuration (ctrl+S), the OpenShift explorer is not updated with the new route url

tsmaeder:
Edit connections-> fixed.

tsmaeder:
Bug is due to the fact that a child element (route) is used to render the parent element (service). Notification is correct in that the service has not changed itself.
Special-cased this in OpenshiftExplorerContentProvider

PR #1218, opened 2016-06-09
JBIDE-22575 - Read docker registry from service-manager

robstryker:
Tested with service manager 1.0.1,  1.0.2,  and 1.1.0.

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ServiceManagerEnvironment.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ServiceManagerEnvironment.java
@@ -51,6 +51,10 @@ public class ServiceManagerEnvironment {
 		return openshiftPort;
 	}
 	
+	public String get(String key) {
+		return env == null ? null : env.get(key);
+	}
+	
 	public static ServiceManagerEnvironment getOrLoadServiceManagerEnvironment(IServer server, boolean save) {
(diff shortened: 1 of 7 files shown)


alexeykazakov:
Any tests?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ServiceManagerEnvironment.java
```diff
@@ -93,7 +97,9 @@ public static ServiceManagerEnvironment loadServiceManagerEnvironment(IServer se
 	 */
 	public static ServiceManagerEnvironment loadServiceManagerEnvironment(IServer server) {
 		
-		String[] args = new String[]{CDKConstants.VAGRANT_CMD_SERVICE_MANAGER, CDKConstants.VAGRANT_CMD_SERVICE_MANAGER_ARG_ENV, "docker"};
```
alexeykazakov: So you removed "docker" from the argument? But what about service manager 1.0.1 (CDK 2.0.0) support?

fbricon: @robstryker I'm pretty sure you told me old versions of the service-manager plugin wouldn't support just `vagrant service-manager env`

alexeykazakov:
OK. If you are sure it also works fine with CDK 2.0. (service manager 1.0.1) then +1 for merging to 4.4.0.x.

fbricon:
+1

PR #1222, opened 2016-06-10
JBIDE-22585 - username not persisted properly in new server wizard for cdk

robstryker:


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/ui/internal/CDKServerWizardFragment.java
```diff
@@ -78,6 +80,7 @@ public Composite createComposite(Composite parent, IWizardHandle handle) {
 		credentials.addCredentialListener(new ICredentialCompositeListener() {
 			@Override
 			public void credentialsChanged() {
+				selectedUser = credentials.getUser();
```
adietish: why store the value in the validator if it seems to get overriden by line123?

bdshadow: @adietish after the current line we get selectedUser updated after every credentials update which is then used in performFinish (line 210). And in line 123 it's just a one time assignment of selectedUser

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/ui/internal/CDKServerWizardFragment.java
```diff
@@ -67,6 +68,7 @@ public ImageDescriptor getImageDescriptor() {
 	@Override
 	public Composite createComposite(Composite parent, IWizardHandle handle) {
 		this.handle = handle;
+		selectedUser = null;
```
bdshadow: i think adding this line assumes no sense

robstryker: I'm not sure what you mean by "assumes no sense", but if you mean "doesn't make sense", then you'd be wrong.  The WTP environment only creates one instance of the wizard fragment, and so any time it's used, its member variables basically need to be reset.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/ui/internal/CDKServerWizardFragment.java
robstryker: Line 123 is the initial value after creating the credential composite.   Line 83 is updating any time the credential composite fires an event. 

This is all necessary because performFinish is done in a non-UI thread, so cannot get the value of a widget due to invalid thread access. So the value must be cached in order to be used during performFinish.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/ui/internal/CDKServerWizardFragment.java
bdshadow: @robstryker  ok. thank you for explanation!

PR #1223, opened 2016-06-13
[JBIDE-22596] disable domain validation to cope with *.cdk urls

fbricon:


alexeykazakov:
+1 for respin-b

PR #1224, opened 2016-06-15
[JBIDE-22442] Port Forwarding wizard: When managing forwarding ports, UI is unresponsive when playing with find free local ports

jeffmaury:
Suppress TCP connection establishment to checked port

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
@@ -53,14 +53,10 @@ public class PortForwardingUtils {
 			return false;
 		}
 		try (ServerSocket socket = new ServerSocket(port)){
-			//not in use apparently
+			//not in use
+		    return false;
 		} catch (IOException e) {
 			return true;//in use
-		}
-		try (Socket socket = new Socket("localhost", port)) {
-			return true; // in use
-		} catch (IOException e) {
-			return false;
 		}
 	}
 	


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
```diff
@@ -53,15 +53,11 @@ public static boolean isPortInUse(int port) {
 			return false;
 		}
 		try (ServerSocket socket = new ServerSocket(port)){
-			//not in use apparently
+			//not in use
+		    return false;
```
adietish: obvious, +1

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
```diff
@@ -53,15 +53,11 @@ public static boolean isPortInUse(int port) {
...
 		} catch (IOException e) {
 			return true;//in use
 		}
-		try (Socket socket = new Socket("localhost", port)) {
-			return true; // in use
-		} catch (IOException e) {
-			return false;
-		}
```
adietish: this code looks obscure, trying to allocate a port twice, once with host, once without. Any idea why this was done? Is this maybe something that's required on specific platforms (Mac?). 
I'd suggest that we test this on all 3 platforms before merging.

jeffmaury: Yes that's what Xavier told me (mentioning Mac) so I tested it on Mac and Win10

adietish:
I tested this on Linux and it works flawlessly.
@fbricon can you test this on Mac please?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
adietish: @jeffmaury ok, great. Merging then.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
scabanovich: @adietish , the previous logic was implemented when doing JBIDE-21847. It seemed then that the check with ServerSocket was not always sufficient. @fbricon , could you please try to remember why we kept both checks?

PR #1228, opened 2016-06-16
[JBIDE-22612] Force OpenShift 3 core plugin startup along the UI plugin

fbricon:
This feels like a fugly hack. But it works. 
However I'd be willing to accept any other solution that would be more elegant than this.
cc @adietish @jcantrill @jeffmaury @scabanovich

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/OpenShiftCommonUIActivator.java
+++ b/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/OpenShiftCommonUIActivator.java
@@ -43,7 +43,14 @@ public class OpenShiftCommonUIActivator extends BaseUIPlugin {
 		connectionsRegistryListener = new ConnectionsRegistryAdapter() {
 			@Override
 			public void connectionAdded(IConnection connection) {
-				OpenShiftUIUtils.showOpenShiftExplorer();
+				try {
+					OpenShiftUIUtils.showOpenShiftExplorer();
+				} catch (Exception e) {
+					// Can happen during workbench startup, while the core plugins are starting.
+					// Since mutiple connections would cause multiple errors, 
+					// it's probably better to swallow the exception, else a user would see multiple 
+					// errors in the log, every time the workbench starts.
+				}
 			}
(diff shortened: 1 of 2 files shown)


scabanovich:
I think the fix is good provided that connections are loaded in start(BundleContext) methods of plugins.

Another  solution may be to load ConnectionsRegistry fully in ConnectionsRegistrySingleton.getInstance() inside synchronized block like

	private static boolean loaded = false;
	public static ConnectionsRegistry getInstance() {
		if(loaded) return model;
		synchronized (lock) {
			if(model != null) return model;
			model = new ConnectionsRegistry();
			//Do loading here or in ConnectionsRegistry constructor
			//Get loader participants from an extension point, and call them.
			loaded = true;
		}	
		return model;
	}
Loaders should be registered with an extension point.
Then acquiring ConnectionsRegistry should never be done from start() method of plugin activators.
In this way, bundles will start without a hindrance and any client of ConnectionsRegistry will receive it fully loaded. Listeners that are now added in start() methods also can be added with the extension point, after the registry is loaded.

jeffmaury:
Why are we not using the startup extension point ?

jeffmaury:
Why are we forcing the user to see the Openshift explorer ? What if he decides to remove this view ?

adietish:
@jeffmaury that's exactly why the view should be the trigger for loading the connections. If there's no view no loading happens and no view is shown.
I really feel like we should find a way so that the explorer (view) is triggering the core plugin activation.
@fbricon can you please add steps to https://issues.jboss.org/browse/JBIDE-22612 so that I can reproduce?

adietish:
I cant reproduce the problem with v3 but with v2: my v2 plugins are not activated, I thus cant add v2 connections (connection wizard wont offer the v2 connection type)

fbricon:
@adietish can you check if the patch fixes your v2 issue?

@jeffmaury @scabanovich I also believe connections should be loaded from an extension point, but I wanted to get a simple fix working first

fbricon:
@jeffmaury not having the view opened when the working on openshift doesn't make sense.
The view will only open when the openshift plugins are started, i.e. either because the view is already visible, or the user accesses some openshift feature

adietish:
superseeded by #1230

PR #1235, opened 2016-06-22
[JBIDE-22079] - Import wizard: Error when import a project already existing in workspace

jeffmaury:
- Add warning dialog for Maven projects
- Cleanup workspace in case of projects already exists

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/application/importoperation/GeneralProjectImportOperation.java
+++ b/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/application/importoperation/GeneralProjectImportOperation.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011 Red Hat, Inc.
+ * Copyright (c) 2011-2016 Red Hat, Inc.
  * Distributed under license by Red Hat, Inc. All rights reserved.
  * This program is made available under the terms of the
  * Eclipse Public License v1.0 which accompanies this distribution,
@@ -22,13 +22,12 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.Path;
-import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.osgi.util.NLS;
-import org.eclipse.swt.widgets.Display;
-import org.eclipse.ui.PlatformUI;
+import org.jboss.tools.openshift.internal.common.ui.OpenShiftCommonUIMessages;
 
 /**
  * @author Andre Dietisheim <adietish@redhat.com>
+ * @author Jeff Maury
  * 
  */
 public class GeneralProjectImportOperation extends AbstractProjectImportOperation {
@@ -42,8 +41,9 @@ public class GeneralProjectImportOperation extends AbstractProjectImportOperatio
 
 		IWorkspace workspace = ResourcesPlugin.getWorkspace();
 		IProject project = workspace.getRoot().getProject(getProjectDirectory().getName());
-		overwriteExistingProject(project, workspace, monitor);
-		importToWorkspace(getProjectDirectory(), workspace, monitor);
+		if (overwriteExistingProject(project, workspace, monitor)) {
+	        importToWorkspace(getProjectDirectory(), workspace, monitor);
+		}
 		return Collections.singletonList(project);
 	}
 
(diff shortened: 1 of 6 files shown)


scabanovich:
A project may be previously deleted from workspace without deleting it from disk. Will import operation work successfully in this case? Even so, it may be worth to warn user of the fact that resources on disk are going to be overwritten.

jeffmaury:
This is done in the Import Wizard, the user has to check the reuse repo flag if files/resources already exist

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/application/importoperation/GeneralProjectImportOperation.java
```diff
@@ -42,8 +41,9 @@ public GeneralProjectImportOperation(File projectDirectory) {
 
 		IWorkspace workspace = ResourcesPlugin.getWorkspace();
 		IProject project = workspace.getRoot().getProject(getProjectDirectory().getName());
-		overwriteExistingProject(project, workspace, monitor);
-		importToWorkspace(getProjectDirectory(), workspace, monitor);
+		if (overwriteExistingProject(project, workspace, monitor)) {
```
scabanovich: Do we need to ask user at this moment? Method importToWorkspace() called right after only creates description and causes refresh.  Overwriting does not affect files on disk, so that if user says no, all the same as soon as he refreshes the project, the effect to it will be the same as delete/create done here.
Maybe, if .project, .classpath etc files are not shared and are not replaced, then it may make sense to recreate them (I would then call it other than 'overwrite'), but in this case, if user cancels, we may need to call the refresh of the project.

jeffmaury: This is a 2 step process

- first files are downloaded and user has control through the reuse repo flag
- then project description may be updated and this is controlled through this dialog

scabanovich: Then, if at the first step new files are downloaded and at the second step user declines updating the project description, may we have to explicitly refresh the project in 'else' block? Or it will be refreshed outside of this method?

jeffmaury: No, refresh was not done. Will add.

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/application/importoperation/MavenProjectImportOperation.java
```diff
@@ -53,9 +59,38 @@ public void setFilters(Collection<String> filters) {
...
+	        final String message = (s1.size() == 1) ? NLS.bind(OpenShiftCommonUIMessages.MavenProjectWarningMessage,
+                                                               s1.iterator().next().getName())
+	                                                : OpenShiftCommonUIMessages.MavenProjectsWarningMessage; 
+                        
+	        overwrite = displayOverwriteDialog(OpenShiftCommonUIMessages.OverwriteProjectsDialogTitle, message);
+            for(IProject project : s1) {
+                if (overwrite) {
+                    project.delete(false, true, monitor);
+                } else {
+                    project.refreshLocal(IResource.DEPTH_INFINITE, monitor);
```
adietish: if we do not overwrite, why do we have to refresh?

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/application/importoperation/MavenProjectImportOperation.java
```diff
@@ -53,9 +59,38 @@ public void setFilters(Collection<String> filters) {
...
 		ProjectImportConfiguration projectImportConfiguration = new ProjectImportConfiguration();
-		List<IMavenProjectImportResult> importResults =
-				configurationManager.importProjects(projectInfos, projectImportConfiguration, monitor);
-		return validate(toProjects(importResults));
+		if (overwriteMavenProjects(projectInfos, projectImportConfiguration, monitor)) {
+	        List<IMavenProjectImportResult> importResults =
+	                configurationManager.importProjects(projectInfos, projectImportConfiguration, monitor);
+	        return validate(toProjects(importResults));
+		} else {
+		    return Collections.EMPTY_LIST;
```
adietish: replace by ```Collections.emptyList()``` to avoid a compiler warning about unchecked conversion to List<IProject>

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/application/importoperation/MavenProjectImportOperation.java
scabanovich: Resources on disk are already changed, there was another question dialog for that, now user has been asked if he wants to overwrite the project but that did not mean resources on disk, but exactly the project resource object itself. if user declines, he remains with old project resource referencing new resources on disk, the workspace has to be made aware of it by calling refresh.

adietish:
I still have the error described in jira when I try to create and import twice (in sequence) jboss-eap-quickstarts cloned to 2 different locations. The warning wasnt displayed, the error came up as screenshotted in jira. Am I missing something?

fbricon:
@adietish everything works fine for me. No errors, even after importing the same source to 2 different locations

adietish:
I retested with branch on latest state and I still have the error:

# ASSERT: have a connection with 2 empty projects (p1 and p2)
# EXEC: create eap-basic-s2i in 1st project (p1) and have it imported to file location 1 (ex. /home/<user>/git)
# ASSERT: jboss-kitchensink gets imported to the Eclipes workspace
# EXEC: create eap-basic-s2i in 2nd project (p2) and have it imported to 2nd file location (ex. /tmp/). 
# ASSERT: cloning happens, but workspace import fails with the following error:

Result: 
image

fbricon:
@adietish: nope, still working for me, following your procedure

adietish:
weird, after updating TP, refreshing/rebuilding the dialog appears. +1 LGTM

PR #1236, opened 2016-06-23
[JBIDE-22595] dont allow pw-prompt while in connection wizard

adietish:
# WORK IN PROGRESS, DO NOT MERGE

PR #1241, opened 2016-06-27
jbide-22594: persist credentials between "Deploy to OS" wizard pages

bdshadow:
DON'T MERGE UNTIL "[JBIDE-22595 dont allow pw-prompt while in connection wizard](https://github.com/jbosstools/jbosstools-openshift/pull/1236)" IS MERGED. Some small conflicts will be needed to be resolved

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
+++ b/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
@@ -348,9 +348,12 @@ public class ConnectionWizardPageModel extends ObservableUIPojo {
 		IStatus status = Status.OK_STATUS;
 		listener.secureStoreException = null;
 		try {
-			IConnection connection = createConnection(connectionFactory, connectionAuthenticationProvider);
+			IConnection connection = 
+					isNewConnection()
+					? createConnection(connectionFactory, connectionAuthenticationProvider)
+					: this.selectedConnection;
 			if(connection != null) {
-				addConnectionListener(connection);
+				addConnectionListener(connection);	
 				if (connection.connect()) {
 					connection.enablePromptCredentials(true);
 					this.connection = connection;
--- a/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
+++ b/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
@@ -348,12 +348,9 @@ public class ConnectionWizardPageModel extends ObservableUIPojo {
 		IStatus status = Status.OK_STATUS;
 		listener.secureStoreException = null;
 		try {
-			IConnection connection = 
-					isNewConnection()
-					? createConnection(connectionFactory, connectionAuthenticationProvider)
-					: this.selectedConnection;
+			IConnection connection = createConnection(connectionFactory, connectionAuthenticationProvider);
 			if(connection != null) {
-				addConnectionListener(connection);	
+				addConnectionListener(connection);
 				if (connection.connect()) {
 					connection.enablePromptCredentials(true);
 					this.connection = connection;
--- a/tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/core/connection/ConnectionTest.java
+++ b/tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/core/connection/ConnectionTest.java
@@ -570,7 +570,7 @@ public class ConnectionTest {
 	}	
 	
 	@Test
-	public void should_clear_token_when_connecting_successfully_with_password() throws Exception {
+	public void should_not_clear_token_when_connecting_successfully_with_password() throws Exception {
 		// given
 		testableConnection.setAuthScheme(IAuthorizationContext.AUTHSCHEME_BASIC);
 		testableConnection.setPassword("007"); // non-used pw, should be cleared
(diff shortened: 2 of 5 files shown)


bdshadow:
rebased and resolved all issues

fbricon:
This does not solve the problem where, when connected to openshift, you deploy a docker image, the connection token is null

```java
	public void initModelRegistryWithConnection(Connection connection) {
		if(connection != null) {
			setTargetRegistryLocation(
				(String) connection.getExtendedProperties().get(ICommonAttributes.IMAGE_REGISTRY_URL_KEY));
			setTargetRegistryUsername(connection.getUsername());
			setTargetRegistryPassword(connection.getToken());
		}
	}
```

I suspect this regression might have been caused by something along  https://github.com/jbosstools/jbosstools-openshift/commit/67bcf6d3d2c70d02744e45e6875bc54a4d9bba1d#diff-63f0fe66ae00113b07727df6263af68cR247, although I haven't investigated in more details. @adietish?

bdshadow:
@fbricon i've just reverted to the previous commit of the one you provided link to above.  However, when i 
1) don't have any OpenShift Connection selected in the OS Explorer; 
2) try to Deploy Docker Image;
3) choose connection in the Connection Page
4) Press next
username and password are not filled in.

Also i tested for this commit and for the current master state the situation when connection is selected in OS Explorer. Both username and password are prefilled when the Deploy page is opened.

bdshadow:
So my commit fixes only the case when we need to pass username\password from connection page to deploy page, which didn't work

fbricon:
@bdshadow was the OpenShift connection created manually? or through the CDK Server Adapter?
Does it have the registry url in the advanced section of the connection wizard?

bdshadow:
@fbricon , itried with 2 connections: that has registry url and that doesn't. Both were created through CDK.
So currently in master:
If any of these two OS connections selected in OS Explorer and we press "Deploy to OS..." in docker -> Deploy page appears and all data is filled according to the connection. For both username is prefilled + 
{For the on WITH URL} -> only URL is filled in 
{For the one WITHOUT url} -> only password is filled in

But if no OS connection is selected -> Connection page opens and after choosing one it isn't passed to the Deploy page. 
This PR fixes it. But i noticed one problem. After merge with @adietish 's [this PR](https://github.com/jbosstools/jbosstools-openshift/pull/1236), URL isn't passed. This happened because of the [following change](https://github.com/jbosstools/jbosstools-openshift/pull/1236/files#diff-01c4863bc31948052342aedc6a09d290R220):
[before](https://github.com/jbosstools/jbosstools-openshift/pull/1236/files#diff-550344f773e2ea9e5dbaae8818136c3bL79):

```java
model.setTargetRegistryLocation((String)connection.getExtendedProperties().get(ICommonAttributes.IMAGE_REGISTRY_URL_KEY));
model.setTargetRegistryUsername(connection.getUsername());
model.setTargetRegistryPassword(connection.getToken());
```

[after](https://github.com/jbosstools/jbosstools-openshift/pull/1236/files#diff-01c4863bc31948052342aedc6a09d290R220):

```java
setTargetRegistryLocation((String)connection.getExtendedProperties().get(ICommonAttributes.IMAGE_REGISTRY_URL_KEY));
setTargetRegistryUsername(connection.getUsername());
setTargetRegistryPassword(connection.getToken());
```

which works now through `firePropertyChange`

fbricon:
the connection token is still null, which is a pretty big regression from last release, since deploying a docker image to the CDK registry  (https://www.youtube.com/watch?v=BnUz2TuL25M) is completely broken

bdshadow:
Tested it once again commit by commit. @fbricon, you were right about the commit which broke. Seems to me i missed something when tested it first time.

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/core/connection/ConnectionTest.java
```diff
@@ -581,7 +581,8 @@ public void should_clear_token_when_connecting_successfully_with_password() thro
 		assertThat(isConnected).isTrue();
 		// then
 		verify(store).remove(Connection.SECURE_STORAGE_TOKEN_KEY);
-		assertThat(testableConnection.getToken()).isNullOrEmpty();
+		//normally, even with basic auth, the connection should get a token
+		assertThat(testableConnection.getToken()).isNotEmpty();
```
fbricon: assertThat(testableConnection.getToken()).isEqualTo("42");//token was reset

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
```diff
@@ -348,9 +348,12 @@ public IStatus connect() {
 		IStatus status = Status.OK_STATUS;
 		listener.secureStoreException = null;
 		try {
-			IConnection connection = createConnection(connectionFactory, connectionAuthenticationProvider);
+			IConnection connection =
```
fbricon: this change actually breaks authentication. If you have an existing connection without stored password, it won't let you authenticate. Reverting the changes works

plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/connection/ConnectionWizardPageModel.java
```diff
@@ -348,9 +348,12 @@ public IStatus connect() {
 		IStatus status = Status.OK_STATUS;
 		listener.secureStoreException = null;
 		try {
-			IConnection connection = createConnection(connectionFactory, connectionAuthenticationProvider);
+			IConnection connection = 
+					isNewConnection()
+					? createConnection(connectionFactory, connectionAuthenticationProvider)
```
adietish: we should never edit/use the (edited) connection directly since you can cancel the wizard. You expect the edited connection to be untouched if you cancel the wizard.

PR #1243, opened 2016-06-29
[JBIDE-22639] display proper err & keep editor disabled if no connection exists

adietish:
@fbricon @jeffmaury @bdshadow plz review

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
@@ -20,8 +20,6 @@ import org.eclipse.core.databinding.DataBindingContext;
 import org.eclipse.core.databinding.beans.BeanProperties;
 import org.eclipse.core.databinding.conversion.Converter;
 import org.eclipse.core.databinding.observable.value.IObservableValue;
-import org.eclipse.core.databinding.observable.value.IValueChangeListener;
-import org.eclipse.core.databinding.observable.value.ValueChangeEvent;
 import org.eclipse.core.databinding.validation.IValidator;
 import org.eclipse.core.databinding.validation.ValidationStatus;
 import org.eclipse.core.resources.IProject;
@@ -37,6 +35,7 @@ import org.eclipse.jface.databinding.swt.WidgetProperties;
 import org.eclipse.jface.databinding.viewers.ObservableListContentProvider;
 import org.eclipse.jface.databinding.viewers.ViewerProperties;
 import org.eclipse.jface.dialogs.Dialog;
+import org.eclipse.jface.dialogs.ErrorDialog;
 import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.jface.layout.GridLayoutFactory;
 import org.eclipse.jface.viewers.ColumnLabelProvider;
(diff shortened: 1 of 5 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
```diff
@@ -545,26 +543,35 @@ protected IStatus run(IProgressMonitor monitor) {
 								return Status.OK_STATUS;
 							}
 						})
-				.runWhenDone(
-						new Job("Setting connection, deploy project...") {
+				.runWhenDoneIf(job -> { 
+							if (connection == null) {
+								container.getDisplay().syncExec(() ->
```
scabanovich: It should be tested when Eclipse is closed/restarted with the editor selected.

adietish: +1, that's what I did. Upon startup with an opened editor the error message is displayed as expected.
image

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
```diff
@@ -530,48 +529,71 @@ public boolean isCanceled() {
...
 							}
 						})
 				.runWhenDone(new DisableAllWidgetsJob(false, container, false, busyCursor))
 				.schedule();
 	}
+	
+	private String getConnectionErrorMessage(String url, IServerAttributes server) {
+		ConnectionURL connectionUrl = ConnectionURL.safeForURL(OpenShiftServerUtils.getConnectionURL(server));
+		if (connectionUrl == null) {
+			return "Could not find OpenShift connection for server \"{0}\"";
```
bdshadow: Why don't we use property files for all such strings? I've seen it almost everywhere in openshift UI plugins. There is even an existing property file: [OpenShiftUIMessages.properties](https://github.com/jbosstools/jbosstools-openshift/blob/master/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/OpenShiftUIMessages.properties). In opinio it's more comfortable to have things like that in one place. However it's a matter of taste.

adietish: It is very unlikely that we ever translate it to any other language than English. In the sake of efficiency I personally never extract the messages, I only (mostly) make sure that I bind messages (NLS.bind) so that extracting would be as simple as launching the internationalization wizard in Eclipse. Unfortunately there are peope  with different opinions and we were not strict enough to impose either way so that there are some messages and then there are lots on texts that are not extracted.

adietish:
+1 after testing this anyone, please?

bdshadow:
+ 1 tested

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
scabanovich: Great. I have a couple of questions/suggestions that may be put to a new issue or ignored.

Will editor reload if user goes and creates the missing connection,or user has to reopen it manually? The other way, how the successfully loaded editor will react on deleting the connection? In this case it might be good not to show error dialog, but add an error section in the upper part of the editor. That section may be shown when dialog is shown too. Also, it may have active one-click solutions to the problem (e.g. buttons):
- Create the connection (wizard prefilled with host and username)
- Delete server adapter

adietish:
thx!, merging

PR #1250, opened 2016-07-12
[JBIDE-22592] Application wizard: Failed even though provided required value to parameters

jeffmaury:
- Change parameters list if values are different

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/fromtemplate/ApplicationSourceFromTemplateModel.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/fromtemplate/ApplicationSourceFromTemplateModel.java
@@ -50,6 +50,7 @@ import com.openshift.restclient.model.template.ITemplate;
  * based on a template
  * 
  * @author jeff.cantrill
+ * @author Jeff Maury
  *
  */
 public class ApplicationSourceFromTemplateModel 
@@ -187,7 +188,15 @@ public class ApplicationSourceFromTemplateModel
 		  parameters.forEach(p -> paramsMap.put(p.getName(), p.getValue()));
 		}
 		originalValueMap = paramsMap;
-		firePropertyChange(PROPERTY_PARAMETERS, this.parameters, this.parameters = injectProjectParameters(this.eclipseProject, parameters));
+		/*
+		 * firePropertyChange will send an event if the array are differents so we need to check
+		 * otherwise we will risk the ObservableList to be based on the wrong list as it is assigned
+		 * in the call.
+		 */
+		List<IParameter> newParameters = injectProjectParameters(this.eclipseProject, parameters);
+		if (!Objects.equals(this.parameters, newParameters)) {
+	        firePropertyChange(PROPERTY_PARAMETERS, this.parameters, this.parameters = newParameters);
+		}
 	}
 
 	private static List<IParameter> injectProjectParameters(org.eclipse.core.resources.IProject project, List<IParameter> originalParameters) {


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/fromtemplate/ApplicationSourceFromTemplateModel.java
```diff
@@ -187,7 +188,15 @@ public void setParameters(List<IParameter> parameters) {
...
 		originalValueMap = paramsMap;
-		firePropertyChange(PROPERTY_PARAMETERS, this.parameters, this.parameters = injectProjectParameters(this.eclipseProject, parameters));
+		/*
+		 * firePropertyChange will send an event if the array are differents so we need to check
+		 * otherwise we will risk the ObservableList to be based on the wrong list as it is assigned
+		 * in the call.
+		 */
+		List<IParameter> newParameters = injectProjectParameters(this.eclipseProject, parameters);
+		if (!Objects.equals(this.parameters, newParameters)) {
+	        firePropertyChange(PROPERTY_PARAMETERS, this.parameters, this.parameters = newParameters);
```
scabanovich: An example of another solution that is applied several times in openshift ui:

		List<org.eclipse.core.resources.IProject> oldProjects = new ArrayList<>(this.projects);
		List<org.eclipse.core.resources.IProject> newProjects = new ArrayList<>();
		if (projects != null) {
			newProjects.addAll(projects);
		}
		this.projects.clear();
		this.projects.addAll(newProjects);
		firePropertyChange(PROPERTY_PROJECTS, oldProjects, this.projects);

It is verbose, but it has an advantage of always providing for clients the same instance of list. If this solution works, it might be applied in all similar cases. It can be moved to one utility method to reduce its usage by a single call.

jeffmaury: Why do you fire a property change if the list property is not modified ?

scabanovich: It will not be fired, PropertyChangeSupport does the necessary check, no need to duplicate it. But whether the list changed and event is fired, or it is not changed and pcs will not fire an event, the model and clients share the same list, which ensures that objects are same.

jeffmaury: As the projects property is not changed, do we need to send an event ? I'm confused because the old value of the event does not match a value taken by the projects property

scabanovich: PropertyChangeSupport calls oldValue.equals(newValue) which for lists compares elements so that it does not matter if it is another instance passed for old value. Sure, if we know that lists contain same elements, we do not need to send the event, but since exactly that is done by pcs, we are spared from doing the check.

jeffmaury: Yeh I got this but my feeling is that is we never change the projects property we should never send a change event. Do we ?

scabanovich: If we want to be strict and not call the fire, then in each similar case we should wrap the code with 

if (!Objects.equals(this.list, newList)) {
   //
}

We will not gain much, because when the list is changed, the comparison of lists will be performed twice - here and in pcs. My point is that we should treat firing lists consistently in cases that have the same pattern. If calling equals() seems good, it should be inserted everywhere we fire lists. Also, since it is good to ensure the same list instance in the model and clients, it may be done in this case too, no matter if equality check is done or not. Otherwise if we decide that ensuring the same list is not needed, then it should be removed from all other cases, for the sake of consistency.

adietish: @scabanovich @jeffmaury the pattern that you spotted in several places (actually in most places where the binding is done to a list of items) is required because of a bug in the databinding where it wont work unless you're always using the very same list. I'd fully agree in moving this to a common class. I believe that there are more elegant solution as a util class though.

adietish:
I looked into this and as far as I can understand the problem I see the following problem in the model: the instance of the selected parameter gets out of sync with the instances contained within the list of all parameters. ApplicationSourceFromTemplateModel#injectProjectParameters clones the parameters (most likely so that instances in the original values, that are used to restore the original the values, dont get affected by changes). But even though new instances are created, the selected parameter is not updated. Editing on the other hand relies on the selected parameter. Since it does not get updated, a dangling instance, that's not contained in the list of all parameters is edited. Therefore when the templates is applied with the parameters, the changes in the parameters wont get applied.
I dont understand yet why the above change fixes the problem (I tested it). Any idea?

adietish:
I also noticed a 2nd problem where each time you choose a template, the paramters are set twice to the model (and most likely loaded twice): NewApplicationWizardModel#updateSelectedAppSource is called twice.

adietish:
I suggest the following PR instead, that does NOT clone the parameters: #1261

adietish:
replaced this by #1261

PR #1255, opened 2016-07-13
jbide-22684: show up login prompt only for appropriate OS connection

bdshadow:
when "Show in -> OS Exlporer" on CDK in Servers view

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/ui/internal/view/CDKActionProvider.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/ui/internal/view/CDKActionProvider.java
@@ -18,7 +18,6 @@ import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.ISelectionProvider;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.jface.viewers.StructuredSelection;
-import org.eclipse.jface.viewers.StructuredViewer;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.IWorkbenchPart;
@@ -27,7 +26,6 @@ import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.navigator.CommonActionProvider;
 import org.eclipse.ui.navigator.CommonNavigator;
-import org.eclipse.ui.navigator.CommonViewer;
 import org.eclipse.ui.navigator.ICommonActionExtensionSite;
 import org.eclipse.ui.navigator.ICommonViewerSite;
 import org.eclipse.ui.navigator.ICommonViewerWorkbenchSite;
@@ -44,6 +42,7 @@ import org.jboss.tools.openshift.cdk.server.core.internal.adapter.CDKServer;
 import org.jboss.tools.openshift.cdk.server.core.internal.listeners.CDKDockerUtility;
 import org.jboss.tools.openshift.cdk.server.core.internal.listeners.CDKOpenshiftUtility;
 import org.jboss.tools.openshift.cdk.server.core.internal.listeners.ServiceManagerEnvironment;
+import org.jboss.tools.openshift.internal.ui.models.OpenshiftUIModel;
 
 public class CDKActionProvider extends CommonActionProvider {
 	private ICommonActionExtensionSite actionSite;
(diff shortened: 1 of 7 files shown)


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/ui/internal/view/CDKActionProvider.java
```diff
@@ -89,8 +84,11 @@ protected Object adaptToViewItem(IServer server) {
 			if( beh != null ) {
 				adb = (ServiceManagerEnvironment)beh.getSharedData(ServiceManagerEnvironment.SHARED_INFO_KEY);
 			}
-			if( adb != null )
-				return new CDKOpenshiftUtility().findExistingOpenshiftConnection(server, adb);
+			
+			if( adb != null ) {
+				OpenshiftUIModel model = new OpenshiftUIModel();
+				return model.getConnectionWrapperForConnection(new CDKOpenshiftUtility().findExistingOpenshiftConnection(server, adb));
```
scabanovich: When OpenshiftUIModel instance is created it is added to ConnectionsRegistrySingleton as a listener. When we do not need that instance it should be disposed by model.dispose(). Otherwise it is a memory leak. Moreover in this case it is 'performance leak' as the created connection wrapper will listen to connection changes and build its content. Now, this code may be improved by calling model.dispose(), since it does nothing but removes the listener from the registry. But nobody can predict what may be added to dispose() later.

Guys! I again appeal with creating a singleton instance of OpenshiftUIModel for every client to use. If tests need its own instance let them create it, but lets give the singleton to runtime.

bdshadow: @jcantrill, @adietish, @fbricon, I agree with Slava. By now i can just add `model.dispose();`, but such code doesn't look good, as i think, and if dispose change, it will lead to even more confusion. Do we need to create one more JIRA for refactoring OpenshiftUIModel to singleton or may i do it in this PR ?

adietish: yes I see the problem that you're pointing at and went unnoticed so far :(
Having singletons would be most likely the more robust option than the manual disposal I think  even though singletons have their own problems

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/AbstractOpenshiftUIElement.java
```diff
@@ -59,7 +59,8 @@ public boolean equals(Object o) {
 			return true;
 		}
 		AbstractOpenshiftUIElement<?, ?> other= (AbstractOpenshiftUIElement<?, ?>) o;
-		return wrapped.equals(other.getWrapped()) && getParent().equals(other.getParent());
+		return wrapped.equals(other.getWrapped())
+				&& (getParent() == null && other.getParent() == null || getParent().equals(other.getParent()));
```
fbricon: if getParent() == null but other.getParent() != null, then you'll get an NPE calling getParent().equals(...)

bdshadow:
@fbricon, @adietish, @scabanovich, please, review. Refactored model to a singleton + fixed equals in AbstractOpenshiftUIElement

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/OpenshiftUIModel.java
```diff
@@ -27,16 +27,25 @@
...
+	private static OpenshiftUIModel instance;
+	
 	private Map<IOpenShiftConnection, ConnectionWrapper> connections = new HashMap<>();
 	private List<IElementListener> listeners = new ArrayList<IElementListener>();
 
 	private IConnectionsRegistryListener listener;
 	
-	public OpenshiftUIModel() {
-		this(ConnectionsRegistrySingleton.getInstance());
+	public static OpenshiftUIModel getInstance() {
```
fbricon: unless you instantiate the instance L31, you'll have to synchronize that method to be thread safe. 
I quite like the [Initialization On Demand Holder pattern](http://blog.crazybob.org/2007/01/lazy-loading-singletons.html) as an alternative

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -45,17 +45,12 @@
...
+	private OpenshiftUIModel model = OpenshiftUIModel.getInstance();
 	private IElementListener listener;
 	private StructuredViewer viewer;
 	private Map<Object, BaseExplorerContentProvider.LoadingStub> stubs = new HashMap<Object, BaseExplorerContentProvider.LoadingStub>();
 
 	public OpenShiftExplorerContentProvider() {
-		this(new OpenshiftUIModel());
-	}
-
-	public OpenShiftExplorerContentProvider(OpenshiftUIModel model) {
```
fbricon: keeping that constructor makes is more testable as you could inject a mock OpenshiftUIModel

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
```diff
@@ -45,17 +45,12 @@
 public class OpenShiftExplorerContentProvider implements ITreeContentProvider {
 	private static final List<String> TERMINATED_STATUS = Arrays.asList("Complete", "Failed", "Error", "Cancelled");
 
-	private OpenshiftUIModel model;
+	private OpenshiftUIModel model = OpenshiftUIModel.getInstance();
```
fbricon: better initialize that model in the default constructor, for testability purposes

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/explorer/OpenShiftExplorerContentProvider.java
bdshadow: mockito doesn't call the default constructor of the mocked class, so it can be mocked anyway

fbricon:
All seems to work fine now. Will merge once all tests have finished running locally

PR #1257, opened 2016-07-18
[JBIDE-22649] - When importing an OS2 app, the Git import does not use the default import folder used while importing OS3

jeffmaury:
- Use common dialog settings to store the Git repo value

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/OpenShiftCommonUIConstants.java
+++ b/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/internal/common/ui/OpenShiftCommonUIConstants.java
@@ -1,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Red Hat Inc..
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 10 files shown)


fbricon:
LGTM but I think we could add some junit tests here to prevent future regressions

adietish:
looks ok to me. The only question I have is why not use preferences vs dialog settings to store the default repo path. Typically you'd store UI settings in DialogSettings while here we have a more core-alike setting that would equally well fit into prefs (and could therefore be exposed in the preferences).

adietish:
testPR

jeffmaury:
Dialog settings is implemented on top of preferences but I remember you got added value using it (dialog size) but not sure it is relevant for wizards

adietish:
@jeffmaury are you sure about dialog settings being stored to preferences? http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Fdialogs_settings.htm talks of xml files (memento as far as I can remember - which is distinct from preferences, but I could be wrong).
Anyway, I'm not against using dialog settings but the value store here feels more like a candidate for preferences to me.

jeffmaury:
You are right, dialog setting uses it's own serialization format (probably for compatibility reasons) and is saved in the workspace.

fbricon:
It doesn't work for OpenShift 2 apps. Even though the stored path is loaded in the express wizard constructor, the default value is still displayed in the UI: http://screencast.com/t/SG4CFbotk

PR #1260, opened 2016-07-19
[JBIDE-22763] use explorer selection if no image is selected

adietish:


adietish:
This was about adding logic back in that went lost in the last refactoring. This now should match the prior state. 
@fbricon @jeffmaury @bdshadow @scabanovich please review

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/DeployImageHandler.java
```diff
@@ -49,10 +49,14 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
 		IProject project = null;
 		ISelection selection = HandlerUtil.getCurrentSelection(event);
 		final IDockerImage image = UIUtils.getFirstElement(selection, IDockerImage.class);
-		if (image != null) {
-			selection = OpenShiftUIUtils.getOpenShiftExplorerSelection();
+		if (image == null
+				|| OpenShiftUIUtils.hasOpenShiftExplorerSelection()) {
```
scabanovich: May it be && rather than || ?

adietish: +1

fbricon:
+1

PR #1262, opened 2016-07-20
jbide-22641: remove secure entries when OS connection deleted

bdshadow:


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
```diff
@@ -311,6 +311,18 @@ private void clearToken() {
...
 		saveOrClear(SECURE_STORAGE_TOKEN_KEY, null, false);
 	}
+	
+	public void removeSecureStoreData() {
+		SecureStore store = getSecureStore(getHost(), getUsername());
+		if (store != null) {
+			try {
+				store.removeNode();
+			} catch (SecureStoreException e) {
+				firePropertyChange(SecureStoreException.ID, null, e);
```
fbricon: why throw an event here? how are the consumers?

bdshadow: @fbricon, the consumers are the same as [here](https://github.com/bdshadow/jbosstools-openshift/blob/fa4164c8bed4ea3bee37a4e6057fb0c8022c0536/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java#L221) =)
Should i delete that firePropertyChange too?

fbricon: No, that's fine for now

adietish:
well have to review and reorganize the code for storing the connection and it's secure storage entries. They^re currently in ConnectionPresistency and in Connection and I believe that we should move all to a single location, the persistency. 
Tested this patch, works for me.
+1

PR #1264, opened 2016-07-21
[JBIDE-22773] - Openshift server adapter can't start with eap64-basics2i template on Windows: can't copy MANIFEST.MF reported

jeffmaury:
- Suppress perms mapping when performing oc rsync


adietish:
publishing eap in Linux still works properly after the change. Launching my virtualbox to verify that the problem is fixed with your change.

fbricon:
LGTM

adietish:
+1

PR #1266, opened 2016-07-22
jbide-22738: persist image registry URL in "Deploy to OS" wizard

bdshadow:


bdshadow:
testPR

fbricon:
if you go back to the connection page to change the registry url value, when going to the deploy page next, the old url value is shown

PR #1269, opened 2016-07-28
[JBIDE-22704] - Application Wizard: Application based on builder images do not have prefilled configuration

jeffmaury:


adietish:
@mlabuda do we have this covered with a bot-test?

mlabuda:
Nope. There are tests only when wizard is opened from context menu of a project. Test for opening from shell menu and central are missing, but current tests for handling builder image wizard are easily extensible for those use cases too.

adietish:
+1, tested the PR, works for me.

adietish:
@mlabuda ok, we should then try to do such a bot test. @jeffmaury agree?

mlabuda:
Yeah, we should. The tests are located at https://github.com/jbosstools/jbosstools-integration-tests/blob/master/tests/org.jboss.tools.openshift.ui.bot.test/src/org/jboss/tools/openshift/ui/bot/test/application/v3/basic/BuilderImageApplicationWizardHandlingTest.java

fbricon:
Merged as a bot test will need to live in a separate repo anyway

PR #1244, opened 2016-06-30
[JBIDE-22603] - Sometimes multiple OpenShift watch managers are periodically created and finished

jeffmaury:
- Use a single watcher job per Openshift project
- Job got respawned every 5mn because of web socket read timeout
- Watchers are now keyed per connection and project instead of project

Signed-off-by: Jeff MAURY <jmaury@redhat.com>

PR #1271, opened 2016-07-29
[JBIDE-22603] - Sometimes multiple OpenShift watch managers are periodically created and finished

jcantrill:


jcantrill:
@fbricon This partially updates and should supercede #1244 .  It reverts @jeffmaury changes that reverted a change I introduced early where there is 1 listener per kind per project.  We went to this structure because we saw issues where using a single listener per project would receive multiple disconnect notices and we were improperly handling restarting of the watch manager. I would prefer a single listener per project but I don't have the bandwidth to resolve that at the moment.  Additionally, I don't know if this corrects the original issue @jeffmaury was trying to fix, but it does resolve an issue I saw where the same project for 2 separate connections was not being updated properly.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/IOpenShiftConnection.java
```diff
@@ -38,6 +38,8 @@
 	
 	<T extends IResource> List<T> getResources(String kind, String namespace);
 
+	<T extends IResource> T getResource(IResource resource);
```
fbricon: That feels pretty confusing from an API perspective. Get a resource from a resource. At the very least this requires some Javadoc

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -271,9 +331,9 @@ public void connectionRemoved(IConnection connection) {
 			Connection conn = (Connection)connection;
 			synchronized (watches) {
 				watches.keySet().stream()
-					.filter(p->conn.ownsResource(p))
+					.filter(k->k.connection.equals(conn))
```
fbricon: will that work if the connection token changed?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/IOpenShiftConnection.java
fbricon: getResource(...) doesn't seem to be used in this PR

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
jeffmaury: connection.equals deals with URL and basic authentication info (so likely will not work with OAuth). Wondering if == should be used

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/IOpenShiftConnection.java
jeffmaury: The rationale is that the connection is now used as IOpenShiftConnection and getResource(IResource) was used (in WatchManager) so it was added in the interface definition but I agree I should have copied the Javadoc at least

jcantrill: I dont remember this methods purpose?  Is it to refresh a resource?  or to load a resource from basically a predicate?  should we call it something else?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
jcantrill: @fbricon connection should really not consider token during evaluation.  I thought @adietish made changes a while back to address that.  I thought equality was determined to me server + username

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/IOpenShiftConnection.java
jeffmaury: It's a refresh

jcantrill: I think we should rename it to make it more obvious

fbricon: +1 to rename it to refresh(...)

jeffmaury:
testPR

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
adietish: @fbricon @jcantrill @jeffmaury this was corrected quite some time ago. Connection#equals does NOT consider authentication info, it just compares username and server-urlhttps://github.com/adietish/jbosstools-openshift/blob/master/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java#L626 You can even see this in the tests: https://github.com/adietish/jbosstools-openshift/blob/master/tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/core/connection/ConnectionTest.java#L203

fbricon:
Merged as https://github.com/jbosstools/jbosstools-openshift/commit/d4565ac20d6eeea695890110dc4e3f8ac85a55ba

PR #1272, opened 2016-07-31
jbide-22433: add Storage tab to Properties

bdshadow:


bdshadow:
@adietish @jcantrill @jeffmaury @fbricon please, have a look

plugins/org.jboss.tools.openshift.ui/plugin.xml
```diff
@@ -978,6 +978,34 @@
...
   </menuContribution>
+  <menuContribution
+        allPopups="false"
+        locationURI="popup:org.jboss.tools.openshift.ui.properties.tab.StorageTab">
+     <command
+           commandId="org.jboss.tools.openshift.ui.command.resource.refresh"
+           icon="icons/refresh.gif"
+           id="org.jboss.tools.openshift.ui.command.resource.refresh"
+           style="push"
+           tooltip="Refresh the resource">
```
jcantrill: @fbricon Since we have to add one of these per tab, is there any reason for it not to say 'Refresh the storage'; otherwise I would suggest changing them all to just 'Refresh'

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/OpenInWebConsoleHandler.java
```diff
@@ -90,6 +91,8 @@ private String getWebConsoleUrl(Connection connection, IResource resource) {
 				url.append("/services/").append(resource.getName());
 			} else if (resource instanceof IImageStream) {
 				url.append("/images/").append(resource.getName());
+			} else if (resource instanceof IPersistentVolumeClaim) {
```
jcantrill: I think this surrounding 'if' block should be refactored because of all the mostly same logic; maybe introduce a map

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -53,7 +53,8 @@
 			ResourceKind.BUILD, 
 			ResourceKind.IMAGE_STREAM, 
 			ResourceKind.ROUTE,
-			ResourceKind.TEMPLATE
+			ResourceKind.TEMPLATE,
```
jcantrill: Not necessarily your refactor but why do we have this list twice; here and IConnectionWrapper?

jcantrill:
Overall minor nits IMO.  There might be some room also for tests.

plugins/org.jboss.tools.openshift.ui/plugin.xml
fbricon: Agreed that having a simple "Refresh" command is self-describing, for all resource objects. The user when-right clicking knows what he's refreshing.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
jeffmaury: See https://issues.jboss.org/browse/JBIDE-22727

jeffmaury:
There is a missing information for the persistent volume claim: the status that states if the claim is bound or not and if it is give the name of the volume the claim is bound to:

Web console:
image


JBossTools:
image

bdshadow:
@jeffmaury this information is actually in the Status column. However, yes, there is no info about the volume it is bounded to

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
bdshadow: just refactored this thing and noticed that there is a little difference in the arrays: there is `ResourceKind.IMAGE_STREAM_TAG` in ConnectionWrapper, but there is no such element in WatchManager's array. 
After i made them use the same array (including `ResourceKind.IMAGE_STREAM_TAG`) i started to get the following message in the console when expanding the service: 

> WARNING: A connection to https://10.1.2.2:8443/ was leaked. Did you forget to close a response body?

When i deleted `ResourceKind.IMAGE_STREAM_TAG` it became ok.
@jeffmaury @jcantrill so do we need this ResourceKind for both classes? Was it just missed in WatchManager before?

adietish: @jcantrill does WatchManager ommit IMAGE_STREAM_TAG on purpose? Is WatchManager not observing these on purpose?

adietish: @jeffmaury I'm not sure that refactoring the 2 arrays makes sense:
I understand the KINDS in WatchManager as a declaration what resource the WatchManager wants to watch. But this is only valid if not observing the image stream tags (see @bdshadow 's comment above) in WatchManager was a thoughtful decision. If that's not the case and the WatchManager shall observe all Resources, then yeah, you're right, WatchManager should use the list in ConnectionWrapper (should we move the declaration to the interface?) I believe.

jcantrill: We should investigate, but I believe IMAGE_STREAM_TAG is not a watchable resource.  Trying to watch on such a resource should probably throw; we should consider adding an issue to ensure we do that.

bdshadow:
@jeffmaury, i added the name of the volume the claim is bound to. 
@jcantrill, @adietish, I believe that refactoring of duplicated ResourceKind-s should be done in https://issues.jboss.org/browse/JBIDE-22727, because it requires additional investigation (see the comments [above](https://github.com/jbosstools/jbosstools-openshift/pull/1272#discussion_r73242201))

jeffmaury:
LGTM except that I expected the name of the volume to be displayed as well in the misc properties. WDYT ?

bdshadow:
@jeffmaury yes, i also think about it, but finally decided to do it the same way as it done in openshift console, see [here](https://github.com/openshift/origin-web-console/blob/563e73a7269dd51f90e201f2b5092d498cd0d420/app/views/browse/persistent-volume-claim.html#L65)

jeffmaury:
Yes but in the Openshift console, the volume is displayed in the storage list (equivalent of our table) and in the detail (equivalent of our properties). @fbricon @adietish WDYT ?

fbricon:
indeed, it'd make more sense if the bound volume was also displayed in the details pane.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/OpenInWebConsoleHandler.java
```diff
@@ -67,31 +70,55 @@ public Object execute(final ExecutionEvent event) throws ExecutionException {
...
-				url.append("/services/").append(resource.getName());
-			} else if (resource instanceof IImageStream) {
-				url.append("/images/").append(resource.getName());
-			}
+			url.append(getResourceURL(resource));
 		}
 		return url.toString();
 	}
+	
+	protected String getResourceURL(IResource resource) {
```
fbricon: actually @jcantrill's refactoring is much simpler https://github.com/jbosstools/jbosstools-openshift/pull/1278/files#diff-6567f7e23a24fc10bdeda424c7c7e76dR75

bdshadow: @fbricon i think my version is more universal and doesn't have `if`s that we were trying to get rid of

fbricon: @jcantrill's version is definitely more universal as it rather treats exceptions rather than extensively listing all supported urls

bdshadow: @fbricon that's why my version is universal. It doesn't make exceptions and treats all ReasourceKinds equally

PR #1273, opened 2016-08-02
JBIDE-22446 with enforcer plugin in parent...

nickboldt:
JBIDE-22446 with enforcer plugin in parent pom, only need these 4 lines to enable it for openshift.client plugin


nickboldt:
This PR is not needed at the moment since parent pom is only throwing warnings for AM3 and failures for Final|GA. https://github.com/jbosstools/jbosstools-build/blob/master/parent/pom.xml#L425

PR #1275, opened 2016-08-03
JBIDE-22446 since...

nickboldt:
JBIDE-22446 since maven-enforcer-plugin:enforce-no-snapshots is now enabled for all projects, this will allow you to SKIP the failure using excludePattern with openshift-restclient-java.version; could also skip by setting buildAliasSearch = Final|GA or fail = false


nickboldt:
This PR is not needed at the moment since parent pom is only throwing warnings for AM3 and failures for Final|GA. https://github.com/jbosstools/jbosstools-build/blob/master/parent/pom.xml#L425

PR #1278, opened 2016-08-08
[JBIDE-22817] Add events tab for projects

jcantrill:
This adds an events tabs for kubenetes events.

@fbricon Did we loose a tab for ImageStreams on the 'service' level with this PR?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
@@ -45,14 +45,15 @@ public class WatchManager {
 	private static final long BACKOFF_MILLIS = 5000;
 	private static final long BACKOFF_RESET = FIBONACCI[FIBONACCI.length - 1] * BACKOFF_MILLIS * 2;
 	public static final String [] KINDS = new String[] {
+			ResourceKind.BUILD, 
 			ResourceKind.BUILD_CONFIG, 
 			ResourceKind.DEPLOYMENT_CONFIG, 
-			ResourceKind.SERVICE, 
+			ResourceKind.EVENT, 
+			ResourceKind.IMAGE_STREAM, 
 			ResourceKind.POD,
 			ResourceKind.REPLICATION_CONTROLLER, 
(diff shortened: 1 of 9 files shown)


fbricon:
@jcantrill ImageStreams and Templates are only available at the Project level. At least that's how it has been implemented until then.

fbricon:
It seems to mostly work as expected. I found a minor bug though:
- after scaling up pods, when you right click on an event to show it in the web console, it opens the pods view, regardless of the event kind

However, I have a concern about the volume of events that we receive/display. Should we limit the total events queue? To what value?

Web Console has a way to filter by keywords, that we could also implement in a future iteration

jcantrill:
In regards to open in console, maybe we should just remove that context menu item.  We provide all the details  of the event in the JBT view and probably is not much gained by allowing someone to navigate to the web console.  Side-note, we need to be careful regarding this navigations given the console UI has been redesigned for origin;  do the same url's still apply?

In regards to the volume of events, wont the table just allow you to scroll and you will see only the events that are visible?  Is it a memory concern? If so, maybe a cache that dumps entries after a given count is appropriate.  I presume we could limit that number to 50 or 100.  I would think beyond that they may not be of much use.  We should also maybe have some navigation context menu that takes you to the web console to where you can see all the events.

fbricon:
I haven't been able to check the urls against v1.3 yet, so not sure if that still works. The show in webconsole link would open the events/ url, so we wouldn't see the details of on particular event (we can't AFAIK), but all of them.

As for the high volume of events issue, yes, it's both a memory and a usability issue. Even if we displayed 10 events, but the list contained 1000 of them, scrolling would be slow and unusable. So +1 to limit to 50. Maybe implement /use a list limited to a given size http://stackoverflow.com/questions/5498865/size-limited-queue-that-holds-last-n-elements-in-java

jcantrill:
Added https://issues.jboss.org/browse/JBIDE-22972 to add queue implementation

fbricon:
I'm currently rebasing the PR against master

fbricon:
Merged as https://github.com/jbosstools/jbosstools-openshift/commit/b5a77964c3ecf3c4d4bca55246c924a4df0eaf2e

PR #1282, opened 2016-08-10
Revert "[JBIDE-22773] - Openshift server adapter can't start with eap…

alexeykazakov:
…64-basic-s2i template on Windows: can't copy MANIFEST.MF reported (#1264)"

This reverts commit 7a2c6cf9bbb0740f5867ab1a61314bef77f4c4ba.


alexeykazakov:
Check failed because windows slave doesn't like some symbols in the commit message. All tests pass locally.

dgolovin:
Replace can't to cannot in first comment and remove \" from commit message. That might help.

alexeykazakov:
We need to find a way to fix this problem on jenkins side for WIndows. So we don't have to dance with commit messages to pass it through WIndows slaves.

alexeykazakov:
Replaced by https://github.com/jbosstools/jbosstools-openshift/pull/1284
Rejecting.

PR #1286, opened 2016-08-11
jbide-22643: connection stays loading after cancelling the first attempt to provide the token

bdshadow:


bdshadow:
testPR

bdshadow:
Strangely, the build is successful locally

robstryker:
Code review:
  1) refreshing the parent is correct, since loading node shouldn't exist anymore:  +1
  2) isolating removeStub is fine
  3) breaking three cases out to separate methods is good for clarity +1
  4) significant change as part of (3) is adding the cases for LOAD_STOPPED, which probably fixes the root cause of the bug +1

+1 on the patch if you promise me it fixes the issue, and the errors in build are unrelated to the code you changed.

PR #1291, opened 2016-08-22
[JBIDE-22621] - Server adapter wizard: should be able to import projects to my workspace

jeffmaury:


jeffmaury:
@scabanovich can you check this as I found a strange comment in the page.dispose() method

scabanovich:
It was a reminder, that model = null would cause problems, because dispose() may be called in UI thread while the thread that uses if to perform finish is still in work.

adietish:
Thx!
+1, works for me, merging.

PR #1292, opened 2016-08-22
[p2diff-12] example of how to set snapshot...

nickboldt:
[p2diff-12] example of how to set snapshot and sha256sum variables inside content.xml for a given IU; DO NOT MERGE


nickboldt:
Next step would be to have maven GENERATE these properties when it encounters a SNAPSHOT dependency inside a jar.

PR #1293, opened 2016-08-23
[JBIDE-22916] - Deploy Docker Wizard: Deploying a docker image from OpenShift explorer does not fetch metadata

jeffmaury:
- Change decision between Config and ContainerConfig


jeffmaury:
@jcantrill can you check this, please

jcantrill:
I don't recall if there is an associated test but could there be one to verify this?  Other then that LGTM

jeffmaury:
test is document in the JIRA

adietish:
This unfortunately still doesnt work for me if I paste the image name (ex. "docker.io/openshift/hello-openshift". It also doesnt work when I use "Browse" to select the image. It only works if I use "Search".

adietish:
discussing this shows that pasting doesnt work because my image is in the local docker daemon. "Browse" and "Search" are expected not to work. Thus +1 assuming we do jiras for the remaining issues.

PR #1305, opened 2016-09-07
[JBIDE-22915] - Deploy Docker Image: Replication controller is not created for some images

jeffmaury:
- ImageStream is created with tag to mimic oc behavior

Signed-off-by: Jeff MAURY jmaury@redhat.com


jcantrill:
Is there an associated test to validate this logic?  Also, we should confirm it works against both origin and enterprise releases as 'oc new-app' changed in this regards quite a lot over time and I have concerns it supporting this behavior over the range of versions we need to suport

jeffmaury:
It's documented in the JIRA. You need to deploy the  docker.io/openshift/hello-openshift:v1.2.1 image

adietish:
@jeffmaury @jcantrill This works for me using the steps in jira. I'd love to merge this and maybe file a jira   to make sure it works in enterprise and 1.2?

adietish:
retested it using the steps in jira. replication controller "hello-openshift-1" was created. Works for me.

PR #1309, opened 2016-09-12
[JBIDE-23014] - Allow connections to configure the cluster namespace from which to get clusterwide templates

jeffmaury:

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/AdvancedConnectionEditor.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/AdvancedConnectionEditor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * coright (c) 2016 Red Hat, Inc.
+ * Copyright (c) 2016 Red Hat, Inc.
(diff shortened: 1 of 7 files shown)


adietish:
testPR

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/AdvancedConnectionEditor.java
```diff
@@ -182,6 +202,25 @@ public String getRegistryURL() {
...
+                firePropertyChange(PROP_CLUSTER_NAMESPACE, old, value);
+            }           
+        }
+        
+        public String getClusterNamespace() {
+            Connection connection = getConnection();
+            if(connection != null) {
+                return connection.getClusterNamespace(); 
+            }
+            return "";
```
adietish: Shouldnt we have this defaulting to the common namespace (ICommonAttributes#COMMON_NAMESPACE) so that we can tell the user what this defaults to?

jeffmaury: You are theorically right although I practically that it will never happen because we are editing a connection but will add for code safety

adietish:
+1, LGTM

PR #1310, opened 2016-09-12
[JBIDE-23015] - Creating a route should have a default port

jeffmaury:


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
```diff
@@ -250,11 +258,17 @@ public void handleEvent(){
...
+			final boolean previousRoutePort = model.getRoutingPort() == port;
+			ServicePortDialog dialog = new ServicePortDialog(target, message, model.getServicePorts(), previousRoutePort);
 			if(Window.OK == dialog.open()) {
 				target.setName(NLS.bind("{0}-tcp", target.getPort()));
 				model.updateServicePort(port, target);
 				model.setSelectedServicePort(target);
+				boolean routePort = dialog.isRoutePort();
+				if (routePort != previousRoutePort) {
+				    model.setRoutingPort(routePort?target:null);
+				    portsViewer.refresh();
```
adietish: we're apparently binding the list of ports, do we really have to refresh? if so I'd think that refreshing is more of a workaround for a bug than required code.

jeffmaury: if you only change the port flag, then the port models are not modified and the change will not be reflected to the UI

adietish: @jeffmaury can you explain plz?

jeffmaury: When you edit a port, you can change name, source port, target port and the flag that tells if this port is used by the route. If you edit only this last one, the port model is not changed and the UI won't be updated. Thus the refresh is we detect the flag has changed

adietish: but in this case nothing changes in the table with the ports, thus no need to refresh it, wrong?

jeffmaury: No because the table now displays both the ports model and the global flag that represents the port selected to the route (last column of the table).

adietish:
testPR

plugins/org.jboss.tools.openshift.client/META-INF/MANIFEST.MF
```diff
@@ -6,7 +6,7 @@ Bundle-Version: 3.3.0.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-Vendor: %Bundle-Vendor
 Bundle-Localization: plugin
-Bundle-ClassPath: lib/openshift-restclient-java-5.1.0.Final.jar,
+Bundle-ClassPath: lib/openshift-restclient-java-5.2.0-SNAPSHOT.jar,
```
adietish: I guess that this should be obsolete with JBIDE-23204 having been merged

plugins/org.jboss.tools.openshift.client/pom.xml
```diff
@@ -16,7 +16,7 @@
 			https://repository.jboss.org/nexus/content/repositories/snapshots/com/openshift/openshift-restclient-java/5.0.0-SNAPSHOT/
 			https://repository.jboss.org/nexus/content/repositories/snapshots/com/openshift/openshift-restclient-java/5.0.0-SNAPSHOT/openshift-restclient-java-5.0.0-20160809.161416-8.jar
 		-->
-		<openshift-restclient-java.version>5.1.0.Final</openshift-restclient-java.version>
+		<openshift-restclient-java.version>5.2.0-SNAPSHOT</openshift-restclient-java.version>
```
adietish: I guess that this should be obsolete with JBIDE-23204 having been merged

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/OpenShiftUIActivator.java
```diff
@@ -106,4 +108,11 @@ public static void log(int status, String message, Throwable e) {
 	private String getId() {
 		return getBundle().getSymbolicName();
 	}
+
+    @Override
+    protected void initializeImageRegistry(ImageRegistry reg) {
+        super.initializeImageRegistry(reg);
+        reg.put(OpenshiftUIConstants.CHECKED_IMAGE_KEY, ImageDescriptor.createFromURL(this.getClass().getResource("/icons/checked.png")));
+        reg.put(OpenshiftUIConstants.UNCHECKED_IMAGE_KEY, ImageDescriptor.createFromURL(this.getClass().getResource("/icons/unchecked.png")));
+    }
```
adietish: we have OpenShiftImages. Shouldnt we have an uniformous approach (where all images are images are in this class) for images?

adietish:
sorry for the multiple review-results. I somehow didnt fully get how github reviews are supposed to get used.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
```diff
@@ -116,15 +122,37 @@ private void createExposedPortsControl(Composite parent, DataBindingContext dbc)
 			.applyTo(label);
 		Composite tableContainer = new Composite(container, SWT.NONE);
 		
+        IObservableList portsObservable = BeanProperties.list(
+                IServiceAndRoutingPageModel.PROPERTY_SERVICE_PORTS).observe(model);
 		portsViewer = createTable(tableContainer);
+		ObservableListContentProvider contentProvider = new ObservableListContentProvider();
+		portsViewer.setContentProvider(contentProvider);
+		ObservableMapLabelProvider labelProvider = new ObservableMapLabelProvider(Properties.observeEach(contentProvider.getKnownElements(), BeanProperties.values(ServicePortAdapter.NAME, ServicePortAdapter.PORT, ServicePortAdapter.TARGET_PORT, ServicePortAdapter.ROUTE_PORT))) {
```
bdshadow: Can you, please, break this line. It's length (280) is too big and uncomfortable to read

jeffmaury:
@bdshadow I don't agree with the default port as I think it's error prone and CLI does not do that by default

adietish:
testPR

bdshadow:
+1

adietish:
@jeffmaury would be great if you could rebase this PR, it does not cleanly but only with conflicts that I am not sure how to fix. Thanks!

adietish:
I think that we need to explain to the user what happens when there's no port checked (the route will round-robin). This isnt obvious to the non-poweruser.

adietish:
works for me, merging.

PR #1311, opened 2016-09-13
[JBIDE-23029] allow integration tests to use core classes

adietish:


plugins/org.jboss.tools.openshift.common.core/META-INF/MANIFEST.MF
```diff
@@ -27,69 +27,78 @@ Export-Package: org.jboss.tools.openshift.common.core;
    org.jboss.tools.openshift.express.test,
    org.jboss.tools.openshift.express.ui,
    org.jboss.tools.openshift.test,
-   org.jboss.tools.openshift.ui",
+   org.jboss.tools.openshift.ui,
+   org.jboss.tools.openshift.reddeer",
```
jeffmaury: Why are you exporting reddeer packages in core ?

plugins/org.jboss.tools.openshift.client/META-INF/MANIFEST.MF
```diff
@@ -20,6 +20,7 @@ Bundle-ClassPath: lib/openshift-restclient-java-5.1.0.Final.jar,
  lib/commons-io-2.1.jar,
  .
 Export-Package: com.openshift.internal.restclient;x-friends:="org.jboss.tools.openshift.test",
+ com.openshift.internal.restclient.authorization,
```
jeffmaury: internal package should be exported to friends only, no ?

adietish: +1

adietish: btw. for great reusability the jobs within openshift.ui should be living in core. Their design (dependencies) prevent that unfortunately. IMHO not the best design. That's not the moment to address this though.

adietish:
@jeffmaury I'm not reexporting reddeer packages. I make sure common.core packages are only exported to friends, among them reddeer (which is questionable given they're not internal classes):
> @@ -27,69 +27,78 @@ Export-Package: org.jboss.tools.openshift.common.core;
>     org.jboss.tools.openshift.express.test,
>     org.jboss.tools.openshift.express.ui,
>     org.jboss.tools.openshift.test,
> -   org.jboss.tools.openshift.ui",
> +   org.jboss.tools.openshift.ui,
> +   org.jboss.tools.openshift.reddeer",

adietish:
@jeffmaury removed the friends from public (non-internal) packages

PR #1314, opened 2016-09-17
JBIDE-23178 - vagrant utility not reading process or closing streams …

robstryker:
…correctly


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/VagrantLaunchUtility.java
```diff
@@ -182,39 +181,89 @@ private static void ensureCommandOnPath(String rootCommand, Map<String, String>
...
+		} else {
+			exitCode = p.exitValue();
+		}
+		
+		inGob.cancel();
+		errGob.cancel();
 		
 		List<String> inLines = null;
 		if( exitCode == null ) {
+			inGob.cancel();
```
fbricon: cancel is called 5 lines up

PR #1316, opened 2016-09-19
[JBIDE-23014] - Allow connections to configure the cluster namespace from which to get clusterwide templates

jeffmaury:
- Update for templates


jeffmaury:
@adietish please review

adietish:
testPR

adietish:
@jeffmaury how can I test this? I am guessing that I should add templates to a namespace that's different from openshift. Do you know how to do this?

jeffmaury:
@adietish one simple way to test is to change the common namespace to a non existing namespace and try to deploy an app. You should see no templates and builders

adietish:
@nickboldt @mickaelistria somehow the testPR trigger isnt working any more. See current state of the jenkins check (executed 1d 20h ago, while I commented "testPR" 2h ago and no new build/check was triggered since then)
CC @jeffmaury

jeffmaury:
I did want to implement such a check because it will be costly. Unless data bind supports the notion of delayed check (not on every keystroke)

PR #1320, opened 2016-09-20
Change cdk test feature to say test

mmalina:


features/org.jboss.tools.openshift.cdk.test.feature/feature.properties
```diff
@@ -15,7 +15,7 @@
 # This file should be translated.
 
 # "featureName" property - name of the feature
-featureName=Red Hat Container Development Kit Support
+featureName=Red Hat Container Development Kit Support Test
```
robstryker: I like the idea behind the patch, but I think we need to work on the language a bit.

robstryker: just kidding.   +1   ;)

PR #1321, opened 2016-09-21
 JBIDE-23177 - Server Editor: Combo items have wrong size

snjeza:
https://issues.jboss.org/browse/JBIDE-23177


scabanovich:
This seems a GTK3 issue that is not reproduced on all OS, at least not on my Ubuntu. Then, maybe a fork is desirable that would apply the solution to only GTK3 case.

Do we have already similar solution in our code? I could not find one. For the lack of experience, to me this solution looks like it may cause a cycle of updates and paint events.

scabanovich:
projectcombolayout

snjeza:
You have to use a GTK >= 3.20

scabanovich:
There was a reasonable amount of calls, but the picture above shows what happened to the combo. I tried to fix it with 

		projectsViewer.getControl().addPaintListener(new PaintListener() {
			public void paintControl(PaintEvent e) {
				projectsViewer.getControl().getParent().update();
				projectsViewer.getControl().pack();
				projectsViewer.getControl().getParent().layout(true);
				System.out.println("Called!!!!");
			}
		});

it helped, but that is when I got the eternal loop!

snjeza:
You also need to add a few projects with a long name to OpenShift and select something in the project's combo. If the controls are repainted too often, that can cause flickering on GTK 3.20.
The code you have shown causes it.

PR #1324, opened 2016-09-21
JBIDE-23208 - fix ssl certificate validity parser

robstryker:
This PR has not been appropriately tested. Please test it before merging.

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/SSLCertificatesPreference.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/SSLCertificatesPreference.java
@@ -16,6 +16,7 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.Locale;
 import java.util.Objects;
 
 import org.eclipse.jface.preference.IPreferenceStore;
@@ -103,19 +104,7 @@ public class SSLCertificatesPreference {
 	}
 
 	private boolean isValid(String validity) {
-		int expiresOnIndex = validity.indexOf(X509CertificateParser.EXPIRES_ON_PREFIX);
-		if(expiresOnIndex >= 0) {
-			SimpleDateFormat dateFormat = new SimpleDateFormat(X509CertificateParser.DATE_FORMAT);
-			String expiresOn = validity.substring(expiresOnIndex + X509CertificateParser.EXPIRES_ON_PREFIX.length()).trim();
-			try {
-				Date date = dateFormat.parse(expiresOn);
-				return date.getTime() > System.currentTimeMillis();
-			} catch (ParseException e) {
-				OpenShiftUIActivator.getDefault().getLogger().logError("SSLCertificatesPreference.isValid(String): Could not parse '" + expiresOn + "' in format " + X509CertificateParser.DATE_FORMAT);
-			}
-		}
-		//In case of any failure, just assume that certificate is valid.
-		return true;
+		return X509CertificateParser.isValid(validity);
 	}
 
 	private Item findItem(List<Item> items, Item item) {
--- a/plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/utils/X509CertificateParser.java
+++ b/plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/utils/X509CertificateParser.java
@@ -14,13 +14,17 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.X509Certificate;
+import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
 
 import javax.naming.InvalidNameException;
 import javax.naming.ldap.LdapName;
 import javax.naming.ldap.Rdn;
 
 import org.eclipse.core.runtime.Assert;
+import org.jboss.tools.openshift.internal.common.core.OpenShiftCommonCoreActivator;
 
 /**
  * Parses a given X509 certificate and offers human readable portions of it.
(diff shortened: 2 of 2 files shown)


scabanovich:
@mmalina , that must be because in your Eclipse the default locale is changed. So, the default locale did not work for the old preference, and adding English cannot help because the date is obviously not in English.

The change is to provide future consistency (because all new dates will be saved in English), and backword compatibility (because _working_ dates saved in a default locale will stil be tried in the default locale).

mmalina:
@scabanovich hmm, I'm not sure I understand entirely what's happening here. It seems that the date was saved in the wrong format somewhere already, so this patch didn't help here. Now I started with a new workspace and a new cdk instance (after vagrant destroy) and it works properly - no more SSL parse error.

scabanovich:
@mmalina , sure, because what is broken, is broken :). If the date was saved at some time ago in a locale that is not now default, nor it is English, than the only way to fix this concrete case would be to iterate over all existing locales but it would be hardly reasonable.

scabanovich:
It seems we do not have yet tests for X509CertificateParser/SSLCertificatesPreference.
For a unit test I would suggest 
- making method SSLCertificatesPreference.isValid(String validity) to be public static.
- changing X509CertificateParser.getValidity(X509Certificate) like this

		//Test may override it to getValidity with another locale
		protected String getValidity(X509Certificate certificate) {
			return getValidity(certificate, Locale.ENGLISH);
		}
	
		protected final String getValidity(X509Certificate certificate, Locale locale) {
			SimpleDateFormat dateFormat = new SimpleDateFormat(DATE_FORMAT, locale);
			return new StringBuilder()
				.append(ISSUED_ON_PREFIX).append(dateFormat.format(certificate.getNotBefore())).append('\n')
				.append(EXPIRES_ON_PREFIX).append(dateFormat.format(certificate.getNotAfter()))
				.toString();
		}

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/SSLCertificatesPreference.java
```diff
@@ -105,13 +106,19 @@ public void setAllowedByCertificate(X509Certificate certificate, boolean result)
 	private boolean isValid(String validity) {
 		int expiresOnIndex = validity.indexOf(X509CertificateParser.EXPIRES_ON_PREFIX);
 		if(expiresOnIndex >= 0) {
-			SimpleDateFormat dateFormat = new SimpleDateFormat(X509CertificateParser.DATE_FORMAT);
+			SimpleDateFormat dateFormat = new SimpleDateFormat(X509CertificateParser.DATE_FORMAT,  Locale.ENGLISH);
```
adietish: to me the parsing should go into a single class, the X509CertificateParser, not spread across several ones. Imho this method should be moved into X509CertificateParser. SSLCertificatesPreference being in ui while the parser is in core hints to this too imho. In terms of separation of concerns we have here a class that handles preferences matters, but also parses certificate validity. You also see this in the fact that this PR has to change things in 2 locations, because parsing the certificate is spread across 2 classes

adietish: Yes, we ideally also should have unit tests for this.

adietish: @robstryker @scabanovich Could you please update this so that we can include this change in the upcoming 4.4.2.AM2?

adietish: @robstryker @scabanovich will we be able to merge this this week?

jeffmaury:
I think what is really broken is that we use string to store dates but we should use numeric value of the UTC date (which is what is probably in the certificate). So I would suggest that we change our preference format accordingly

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/SSLCertificatesPreference.java
jeffmaury: In order to test it, just add -nl de to your Eclipse run

scabanovich:
@jeffmaury , that is quite reasonable, but may we need backward compatibility? If not, no problem, if yes, both formats need be supported for a version or two.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/SSLCertificatesPreference.java
scabanovich: Maybe test like this will help

package org.jboss.tools.openshift.test.common.ui.utils;

import java.security.cert.X509Certificate;
import java.util.Date;
import java.util.Locale;

import org.jboss.tools.openshift.common.core.utils.X509CertificateParser;
import org.junit.Assert;
import org.junit.Test;
import org.mockito.Mockito;

public class X509CertificateParserTest {
	
	@Test
	public void testValidity() {
		X509Certificate certificate = Mockito.mock(X509Certificate.class);
		Mockito.when(certificate.getNotBefore()).thenReturn(new Date(new Date().getTime() - 1000000));
		Mockito.when(certificate.getNotAfter()).thenReturn(new Date(new Date().getTime() - 1000000));
		String validity = X509CertificateParser.getValidity(certificate, Locale.ENGLISH);
		Assert.assertFalse("Date is not valid, true means that validation failed with parser error.", X509CertificateParser.isValid(validity));
		
		//Set unlikely locale 
		validity = X509CertificateParser.getValidity(certificate, Locale.TAIWAN);
		Assert.assertTrue("Date is not valid, but locale is wrong, parser error is expected.", X509CertificateParser.isValid(validity));

		validity = X509CertificateParser.getValidity(certificate, Locale.getDefault());
		System.out.println("---------------->" + Locale.getDefault());
		Assert.assertFalse("Date is not valid, true means that validation failed with parser error.", X509CertificateParser.isValid(validity));
	}

}

jeffmaury:
@scabanovich  yes we need to support old format but this is easily feasible

robstryker:
I don't actually know the steps to replicate this. I just had some time and thought I'd implement the quick fix that was listed here or on the jira.  So I can't drive this without further information.

adietish:
This PR was filed almost a month ago. I think that we should at least have this 1st improvement and then have further PRs for the valid suggestions that were suggested.

adietish:
merging this for now. Filed the outstanding issue to new jiras

PR #1327, opened 2016-09-24
jbide-22626: Updating deployed docker image on OpenShift does not work

bdshadow:
@jeffmaury @jcantrill @adietish , please, have a look
Possibly it is better to do update in a separate method, but then we need to get DC from connection again. Also i believe, that some better naming for the method can be created

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/DeployImageJob.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/DeployImageJob.java
@@ -100,7 +100,7 @@ public class DeployImageJob extends AbstractDelegatingMonitorJob
 		try {
 			final Connection connection = parameters.getConnection();
 			final String name = parameters.getResourceName();
-			if(isUpdate(connection, parameters.getProject().getName(), name)) {
+			if(isUpdateThenDo(connection, parameters.getProject().getName(), name)) {
 				return Status.OK_STATUS;
 			}
 			Map<String, IResource> resources = generateResources(connection, name);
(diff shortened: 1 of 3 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/DeployImageJob.java
```diff
@@ -124,18 +124,26 @@ protected IStatus doRun(IProgressMonitor monitor) {
 		return Status.OK_STATUS;
 	}
 
-	private boolean isUpdate(Connection connection, String project, String name) {
+	private boolean isUpdateThenDo(Connection connection, String project, String name) {
```
adietish: isUpdate then do what? I agree that isUpdate() isnt a good choice for a method name, but I still dont get (by reading the method signature) what this method does.
Is this something like "updateTriggerIfIsUpdate" or similar?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/DeployImageJob.java
```diff
@@ -124,18 +124,26 @@ protected IStatus doRun(IProgressMonitor monitor) {
...
-			return trigger != null && 
-				ResourceKind.IMAGE_STREAM_TAG.equals(trigger.getKind()) && 
-				StringUtils.isNotBlank(trigger.getNamespace()) &&  
-				connection.getResource(ResourceKind.IMAGE_STREAM, trigger.getNamespace(), trigger.getFrom().getName()) != null;
-		}catch(NotFoundException e) {
+			if (trigger == null || 
+				!ResourceKind.IMAGE_STREAM_TAG.equals(trigger.getKind()) || 
+				StringUtils.isBlank(trigger.getNamespace()) ||  
+				connection.getResource(ResourceKind.IMAGE_STREAM, trigger.getNamespace(), trigger.getFrom().getName()) == null) {
+				return false;
```
adietish: could we please have unit-tests to assert these rather complex alternative conditions?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImageWizard.java
```diff
@@ -72,11 +71,8 @@ public DeployImageWizard(IDockerImage image, Connection connection, IProject pro
 	private String getImageNameWithTag(IDockerImage image) {
 		String imageName = image.repo();
 		if(!image.tags().isEmpty()) {
-			List<String> tags = new ArrayList<>(image.tags());
-			if(tags.size() > 1) {
-				Collections.sort(tags); //same as in docker explorer
-			}
-			imageName += ":" + tags.get(0);
+			TreeSet<String> tags = new TreeSet<>(image.tags());
+			imageName += ":" + tags.first();
```
adietish: good catch!

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/DeployImageJob.java
bdshadow: is Update? then do update =) That's what i meant. It was isUpdate before and added update to the same method for not to get resources from connection 2 times.
I agree, that  `updateTriggerIfUpdate` is better

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/DeployImageJob.java
bdshadow: ok, i'll work on it

bdshadow:
I had some mess in branches locally, now fixed it. Added test on checking tag updates.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/job/DeployImageJob.java
bdshadow: added

adietish:
testPR

adietish:
+1, works for me.

PR #1330, opened 2016-09-29
[JBIDE-23147] - application wizard: Support optional host when route is created

jeffmaury:

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
@@ -20,11 +20,13 @@ import org.apache.commons.validator.routines.DomainValidator;
 import org.eclipse.core.databinding.DataBindingContext;
 import org.eclipse.core.databinding.beans.BeanProperties;
 import org.eclipse.core.databinding.observable.list.IObservableList;
+import org.eclipse.core.databinding.observable.value.IObservableValue;
 import org.eclipse.core.databinding.validation.IValidator;
 import org.eclipse.core.databinding.validation.MultiValidator;
 import org.eclipse.core.databinding.validation.ValidationStatus;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.jface.databinding.fieldassist.ControlDecorationSupport;
 import org.eclipse.jface.databinding.swt.WidgetProperties;
 import org.eclipse.jface.databinding.viewers.ObservableListContentProvider;
 import org.eclipse.jface.databinding.viewers.ViewerProperties;
@@ -83,10 +85,10 @@ public class ServicesAndRoutingPage extends AbstractOpenShiftWizardPage  {
 		createExposedPortsControl(parent, dbc);
 		
 	    GridDataFactory
-          .fillDefaults()
-          .align(SWT.FILL, SWT.BEGINNING)
-          .grab(true, false)
-          .applyTo(new Label(parent, SWT.SEPARATOR | SWT.HORIZONTAL));
+	        .fillDefaults()
+	        .align(SWT.FILL, SWT.BEGINNING)
+	        .grab(true, false)
+	        .applyTo(new Label(parent, SWT.SEPARATOR | SWT.HORIZONTAL));
 
 		//routing
 		Composite routingContainer = new Composite(parent, SWT.NONE);
@@ -104,15 +106,15 @@ public class ServicesAndRoutingPage extends AbstractOpenShiftWizardPage  {
 		btnAddRoute.setToolTipText("Adding a route to the service will make the image accessible\noutside of the OpenShift cluster on all the available service ports. \nYou can target a specific port by editing the route later.");
 		GridDataFactory.fillDefaults()
 			.align(SWT.FILL, SWT.FILL).grab(false, false).span(2, 1).applyTo(btnAddRoute);
+		final IObservableValue<Boolean> addRouteModelObservable = BeanProperties.value(IServiceAndRoutingPageModel.PROPERTY_ADD_ROUTE).observe(model);
 		ValueBindingBuilder.bind(WidgetProperties.selection().observe(btnAddRoute))
-			.to(BeanProperties.value(IServiceAndRoutingPageModel.PROPERTY_ADD_ROUTE)
-			.observe(model))
+			.to(addRouteModelObservable)
 			.in(dbc);
 		
 		Label labelRouteHostname = new Label(routingContainer, SWT.NONE);
 		labelRouteHostname.setText("Hostname:");
 	    GridDataFactory.fillDefaults()
-          .align(SWT.FILL, SWT.FILL)
+          .align(SWT.FILL, SWT.CENTER)
           .applyTo(labelRouteHostname);
 
 		Text textRouteHostname = new Text(routingContainer, SWT.BORDER);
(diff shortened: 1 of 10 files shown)


bdshadow:
Tested, everything works fine.
Two questions. 
1) I created 2 different applications one after another and set the same host name for both. It finished without any problems, but in web console we can see the warning: 
> Requested host name123 was rejected by the router. Reason: route jboss-webserver30-tomcat already exposes name123 and is older.
Shouldn't we have any similar warning about it?
2) i set "%@#$&^*" for the host name. Then the window with results of creating resources popped up: http://imgur.com/a/cA8Ry. And no host was generated, even default one.
Shouldn't we check the validity of the host name?

adietish:
@bdshadow @jboss-jira-hook I agree, I think that we should validate upfront like the web-ui does.

jeffmaury:
testPR

adietish:
testPR

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
```diff
@@ -92,11 +103,40 @@ protected void doCreateControls(Composite parent, DataBindingContext dbc) {
...
+		
+		Label labelRouteHostname = new Label(routingContainer, SWT.NONE);
+		labelRouteHostname.setText("Hostname:");
+	    GridDataFactory.fillDefaults()
+          .align(SWT.FILL, SWT.FILL)
+          .applyTo(labelRouteHostname);
+
+		Text textRouteHostname = new Text(routingContainer, SWT.BORDER);
+        GridDataFactory.fillDefaults()
+        .align(SWT.FILL, SWT.CENTER)
```
adietish: "Hostname" label is not centered vertically when compared to the text field.
image

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
```diff
@@ -92,11 +103,40 @@ protected void doCreateControls(Composite parent, DataBindingContext dbc) {
...
+        .in(dbc);
+        ValueBindingBuilder.bind(WidgetProperties.text(SWT.Modify).observe(textRouteHostname))
+        .converting(new TrimmingStringConverter())
+        .validatingAfterConvert(new IValidator() {
+            
+            @Override
+            public IStatus validate(Object value) {
+                boolean valid = DomainValidator.getInstance(true).isValid((String) value);               
+                return valid?ValidationStatus.ok():ValidationStatus.error(NLS.bind(OpenShiftUIMessages.InvalidHostNameErrorMessage, value));
+            }
```
adietish: Validation status for a missing value should be CANCEL, not ERROR
image

adietish: Would be great if we could decorate the text field with an appropriate error marker
image

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
```diff
@@ -92,11 +103,40 @@ protected void doCreateControls(Composite parent, DataBindingContext dbc) {
...
+        })
+        .to(BeanProperties.value(IServiceAndRoutingPageModel.PROPERTY_ROUTE_HOSTNAME)
+        .observe(model))
+        .in(dbc);
```
adietish: Validation wont only happen if "Add Route" is checked. It is always executed. 
image
Use a MultiValidator that tracks the checkbox and the textfield for this sake.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
```diff
@@ -124,19 +126,28 @@ protected void doCreateControls(Composite parent, DataBindingContext dbc) {
...
             
             @Override
-            public IStatus validate(Object value) {
-                boolean valid = DomainValidator.getInstance(true).isValid((String) value);               
-                return valid?ValidationStatus.ok():ValidationStatus.error(NLS.bind(OpenShiftUIMessages.InvalidHostNameErrorMessage, value));
+            protected IStatus validate() {
+                IStatus status = ValidationStatus.ok();
+                boolean isAddRoute = addRouteModelObservable.getValue();
+                String hostName = routeHostnameObservable.getValue();
+                if (isAddRoute && StringUtils.isNotBlank(hostName) && !DomainValidator.getInstance(true).isValid(hostName)) {
```
adietish: the validation still doesnt fully work as expected. To work so that an empty Hostname would be marked as required value and an invalid one as error, the validator has to look as follows:
```
				if (isAddRoute) {
					if (StringUtils.isBlank(hostName)) {
						status = ValidationStatus.cancel("Please provide a valid hostname.");
					} else if (!DomainValidator.getInstance(true).isValid(hostName)) {
						status = ValidationStatus
								.error(NLS.bind(OpenShiftUIMessages.InvalidHostNameErrorMessage, hostName));
					}
				}
```

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
```diff
@@ -124,19 +126,28 @@ protected void doCreateControls(Composite parent, DataBindingContext dbc) {
...
+                }
+                return status;
             }
-        })
-        .to(BeanProperties.value(IServiceAndRoutingPageModel.PROPERTY_ROUTE_HOSTNAME)
-        .observe(model))
-        .in(dbc);
+        };
+        dbc.addValidationStatusProvider(validator);
+        ControlDecorationSupport.create(validator, SWT.LEFT | SWT.TOP);
```
adietish: For an empty Hostname to be decorated as required value, the ControlDecorationSupport would have to be implemented as follows:
```
        ControlDecorationSupport.create(validator, SWT.LEFT | SWT.TOP, null, new RequiredControlDecorationUpdater());
```

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicesAndRoutingPage.java
```diff
@@ -124,19 +126,28 @@ protected void doCreateControls(Composite parent, DataBindingContext dbc) {
...
 
 	}
```
adietish: and to mark CANCEL and ERROR as error states, the wizardPageSupport needs to be initialized as follows (otherwise CANCEL wont disable "Next >"):

```
	@Override
	protected void setupWizardPageSupport(DataBindingContext dbc) {
		ParametrizableWizardPageSupport.create(IStatus.ERROR | IStatus.CANCEL, this, dbc);
	}
```

jeffmaury:
@adietish I don't agree with you, don't forget that hostname is optional so that empty value means Openshift will generate the hostname value (using xip.io and others)

adietish:
@jeffmaury ok, I missed this. In this case the validator most likely can stay as is. I'd add this explanation in a explanatory label (below the Hostname?) since this not obvious to the dummy user (as me) I think.

jeffmaury:
OK. Let's do that.

PR #1331, opened 2016-09-29
[JBIDE-23210] made sure deploying docker image is using selected proj

adietish:


scabanovich:
Yes, that is good and seems quite natural. The resource loading in model was necessary long time ago, when deploy image page had the choice for connection. As soon as a separate page for choosing connection was added, and the selector in deploy image page was replaced with info controls, there was no need any more for loading resource immediately on connection setting, it can wait until page activation.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DeployImagePage.java
```diff
@@ -172,6 +173,11 @@ protected IStatus run(IProgressMonitor monitor) {
 			OpenShiftUIActivator.getDefault().getLogger().logError(message, e);
 		}
 	}
+	
+	@Override
+	protected void onPageActivated(DataBindingContext dbc) {
+		loadResources(dbc);
```
scabanovich: Can it be an issue, that when deploy image page is not the first one (for instance, wizard is started from the docker explorer, and the connection page is before it), going back and forward will always cause resource loading, even if connection was not changed (though, this use case is very minor).

adietish: good point, fixing.

PR #1332, opened 2016-10-03
jbide-23152: NPE when deploying image with ENV with no value

bdshadow:


adietish:
+1, worked in the 2nd scenario for me.

PR #1340, opened 2016-10-11
JBIDE-22225: Node.js Debugging support for OpenShift applications

ibuziuk:


ibuziuk:
testPR

ibuziuk:
@nickboldt the build error is the following:

> [ERROR] Cannot resolve project dependencies:
> [ERROR]   Software being installed: org.jboss.tools.openshift.js 1.0.0.qualifier
> [ERROR]   Missing requirement: org.jboss.tools.openshift.js 1.0.0.qualifier requires 'bundle org.eclipse.wst.jsdt.chromium.debug.core 0.5.0' but it could not be found

Basically, building locally via vanilla `mvn clean verify` results in the same error, so I have to explicitly define Neon.1 TP:

`mvn clean verify -Dtpc.version=4.61.0.AM1-SNAPSHOT`

I a bit baffled with all this Neon.1 TP switch because CentralCI is currently using Neon bits. Am I missing something ?

nickboldt:
As we have been doing for years, the jobs are configured to;
* build with the minimum target platform (usually Eclipse 4.x.0)
* run tests with the maximum target platform (usually Eclipse 4.x.z, where z is the latest .z release of Luna, Mars, Neon, Oxygen. etc.)

BUT... if you need things from Neon.z in the minimum TP with which we BUILD, rather than just the maximum TP with which we RUN TESTS, we can add things to the 4.60.x branch as well as the 4.61.x branch, so that the Neon.0 TP is actually the Neon.0+ TP.

This requires @alexeykazakov 's approval, since the usual approach, as noted above, is to BUILD with the MINIMUM TP, not a mix of min/max.

@alexeykazakov please advise. 

@ibuziuk please clarify what you need in the 4.60.1 (Neon.0) TP that is currently only in the 4.61 (Neon.1) TP.

alexeykazakov:
I think this update is worth Neon.1 dependency. So the new openshift.js plugin will require org.eclipse.wst.jsdt.chromium.debug.core;bundle-version="0.5.0" (Neon.1) and won't support older jsdt from Neon.0
@jeffmaury WDYT?

ibuziuk:
@nickboldt thanks for clarifications (I always thought that both building and running tests happens against maximum TP :/ ). Basically, having updated version of plugins defined in "org.eclipse.wst.jsdt.chromium.debug.feature" feature is a requirement for OpenShift Node.js debugging facilities.

nickboldt:
Building and testing with the same "maximum" TP is what we have done until simrel.0 is released (when master moves to Oxygen, we'll have both min and max = Oxygen.M4, then M5, then M6, etc.). 

Once a simrel.0 is available (eg., neon.0, oxygen.0) that becomes the minimum, and maximum moves ahead with maintenance releases (neon.1, neon.2, neon.3).

Further question asked in the JIRA: https://issues.jboss.org/browse/JBIDE-22225?focusedCommentId=13306102&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13306102

ibuziuk:
testPR

ibuziuk:
testPR

alexeykazakov:
@mlabuda please take a look at this PR too.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -188,7 +192,9 @@ public void onDebugChange(DebuggingContext debuggingContext, IProgressMonitor mo
 					if( debuggerLaunch != null ) {
 						overrideHotcodeReplace(server, debuggerLaunch);
 					}
-				}					
+				} else if (isNodeJsProject(server)) {
```
alexeykazakov: I think this check should be done in NodeDebugSessionTracker.
OpenShift launch controller just notifies all registered trackers and it's up to tracker to ignore the project or take any action.

site/category.xml
```diff
@@ -26,10 +26,12 @@
...
 		<category name="JBoss Tools OpenShift Nightly Build Update Site" />
 	</feature>
-
-        <feature url="features/org.jboss.tools.openshift.cdk.feature_0.0.0.jar" id="org.jboss.tools.openshift.cdk.feature" version="0.0.0">
-                <category name="JBoss Tools OpenShift Nightly Build Update Site" />
-        </feature>
+	<feature url="features/org.jboss.tools.openshift.cdk.feature_0.0.0.jar" id="org.jboss.tools.openshift.cdk.feature" version="0.0.0">
+		<category name="JBoss Tools OpenShift Nightly Build Update Site" />
+	</feature>
+	<feature url="features/org.jboss.tools.openshift.js.feature_0.0.0.jar" id="org.jboss.tools.openshift.js.feature" version="0.0.0">
```
alexeykazakov: We also need to include this feature to devstudio installer and in the same categories we have openshift in JBossTools. cc: @nickboldt

nickboldt:
Well, when this is merged and built, you or Ilya can submit a PR against these two files:

https://github.com/jbosstools/jbosstools-build-sites/blob/master/aggregate/site/category.xml
https://github.com/jbdevstudio/jbdevstudio-product/blob/master/site/category.xml

ibuziuk:
@alexeykazakov Thank you for review - all remarks are very relevant
@nickboldt once merged I will provide additional PRs for including new feature to site, product etc.

ibuziuk:
@alexeykazakov PR has been updated

site/category.xml
ibuziuk: will provide separate PR for feature including

alexeykazakov:
I approved this change but I didn't test it. We still need to get some input at least from @mlabuda.
And btw @ibuziuk you don't have to get this PR merged to create PRs for product and sites ;)

ibuziuk:
@alexeykazakov I don't but I would rather wait until applied - maybe someone simply reject it  ;-)

plugins/org.jboss.tools.openshift.js/src/org/jboss/tools/openshift/internal/js/storage/SessionStorage.java
```diff
@@ -0,0 +1,41 @@
...
+ * and debug state will be lost
+ * 
+ * @author "Ilya Buziuk (ibuziuk)"
+ */
+public final class SessionStorage {
+
+	private SessionStorage() {
+	}
+
+	private static final Set<IServer> INSTANCE = Collections.synchronizedSet(new HashSet<>());
```
adietish: ConcurrentSkipListSet or Collections.newSetFromMap(new ConcurrentHashMap()) might be better performance since they allow concurrent read/writes

plugins/org.jboss.tools.openshift.js/src/org/jboss/tools/openshift/js/debug/NodeDebugSessionTracker.java
```diff
@@ -0,0 +1,37 @@
...
+import org.jboss.tools.openshift.js.launcher.NodeDebugLauncher;
+
+/**
+ * @author "Ilya Buziuk (ibuziuk)"
+ */
+public class NodeDebugSessionTracker implements DebugSessionTracker {
+
+	@Override
+	public void startDebugSession(IServer server, int port) throws CoreException {
+		if (NodeDebuggerUtil.isNodeJsProject(server)) {
```
adietish: Shouldnt this throw if the given server is not a nodejs server? It currently does nothing in this case. So how would the calling code or the user know?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -63,7 +65,9 @@
 public class OpenShiftLaunchController extends AbstractSubsystemController
 		implements ISubsystemController, ILaunchServerController {
 
-	private static final String DEBUG_MODE = "debug";
+	private static final String DEBUG_MODE = "debug"; //$NON-NLS-1$
+	public static final String PACKAGE_JSON = "package.json"; //$NON-NLS-1$
```
adietish: This look duplicate/redundant to NodeDebuggerUtil.PACKAGE_JSON

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
```diff
@@ -49,8 +50,18 @@
...
 			throws CoreException {
+		
 		syncDownFailed = false;
-		final IProject deployProject = OpenShiftServerUtils.getDeployProject(getServer());
+		IServer server = getServer();
+		
+		
+		// Noode.js Debug session is running - do *not* perform rsync!
+		if (DebugTrackerContributionEvaluation.isDebugSessionAlive(server)) {
+			return;
```
ibuziuk: @alexeykazakov agree, let's keep it simple for now and not put the cart before the horse ;-)

plugins/org.jboss.tools.openshift.js/src/org/jboss/tools/openshift/js/debug/NodeDebugSessionTracker.java
ibuziuk: this should not throw any exception - if it is not Node.js project tracker simply skip this part and does not start a debug session

plugins/org.jboss.tools.openshift.js/src/org/jboss/tools/openshift/internal/js/storage/SessionStorage.java
ibuziuk: @adietish I do not believe it will have any impact on performance. IMO making the HashSet synchronized is just enough for this case (the number of server adapters can not be sufficiently big and using suggested complex data structures will not boost performance but rather make the code less understandable IMO). So I would rather suggest keep it simple for now, unless there will be any issues coupled with this piece of code (which is very unlikely)

ibuziuk:
@adietish PR has been updated

ibuziuk:
@mlabuda one of the test cases is described in the demo - https://youtu.be/1cKPPQrlC4k

rhopp:
When stopping server adapter (which was running in debug mode) it does not stop the debugging session (V8 engine).
Further more when user stops the debugging session (terminating it in console view) the server adapter starts itself.
I've made a screencast of this behavior: https://youtu.be/aykn8-yK_Nc
In the beginning it's just testing of debugging (works)
~0:45 - stopping the server
~0:40-2:00 - waiting and trying if the debugging session would end
~2:05 - terminating the debugging session (V8 engine)
2:05-end - server adapter is starting again (the termination of debugging session trigerred that)

Other than that I'm not aware of any problems preventing this PR from being pushed.

ibuziuk:
@rhopp thanks for review - I can reproduce
@adietish weird but it looks like server adapter is not terminated. Can it be coupled with  replication controller issue reported yesterday - https://issues.jboss.org/browse/JBIDE-23412 ?
The debug session is running after termination but connection should be immediately lost once the server stopped

ibuziuk:
testPR

ibuziuk:
@rhopp PR was updated.

> Further more when user stops the debugging session (terminating it in console view) the server adapter starts itself.

This issue^ should be fixed now. 
However, stopping server adapter will not stop the debug session - one should still terminate it manually. This might be coupled with the replication controller issue -  https://issues.jboss.org/browse/JBIDE-23412 due to the fact that after stopping server adapter app is still running in the `DEV_MODE ` (which is unexpected behaviour IMO) and debug session is still alive. Moreover, If you try to restart the adapter the debug session will be terminated (expected behaviour). Assuming that manual debug session termination is not coupled with this PR and might be just a side effect of the other issue I would propose to merge this PR in the current state. 

@adietish @alexeykazakov WDYT ?

alexeykazakov:
+1. We can address the remaining issue with debug session separately.

ibuziuk:
@alexeykazakov sounds great
@rhopp could you please verify that terminating debug session does not start server adapter and +1 the PR ?

rhopp:
I can confirm, that terminating debug sesion does not start server adapter.
But I have to give -1 to this PR... We've found, that when one is debugging (the code is stopeed at breakpoint, stepping through code, inspecting variables, ...) longer than ~1 minute, browser displays error code 502 and the debug session gets terminated. This makes this feature not very useful, because all debugging must be quicker than that timeout.
The behavior is captured in this screencast: https://youtu.be/BJf7wcPqNmM (note how at the time 0:42 the page is finally loaded (502 error) and the debug session is terminated (in debug view)).
We were able to reproduce this issue on F24 and Win10 (using CDK and console.engint.openshift.com)

ibuziuk:
I have created two separate issues that have been reported during review
- [Terminating Server Adapter running in debug mode does not terminate Node.js debug session](https://issues.jboss.org/browse/JBIDE-23421)
- [Node.js debug session is terminated after ~1 minute and browser shows 502 error](https://issues.jboss.org/browse/JBIDE-23422)  

I was not able to reproduce JBIDE-23422 on Ubuntu 16.04 -  https://youtu.be/GmQzp_jU754
As you can see debug session is running for more than 4 minutes without 502 error (there is only 504  error which does not terminate debug session and can be treated as expected timeout IMO)

Those two issues are going to be addressed in the next sprints and I still suggest to merge this PR in the current state due to the fact that the JBIDE-23421 is minor one can be treated as side effect of another issue and  JBIDE-23422 is not reproducible on all OS and also not coupled with debugging process itself but rather with some OpenShift configurations.

alexeykazakov:
I agree with @ibuziuk 
+1 to merge the PR

alexeykazakov:
@rhopp @mlabuda please approve this PR. If we want to have this feature done for 4.4.2 (and we do want) then we need to merge it before the codefreeze this week.

jrichter1:
We've just run this through the same hoops on @rawagner 's Ubuntu 16.04 and unlike @ibuziuk we came to the exact same result as we had on Fedora and Windows. That is, debug session is terminated after the application is suspended for a minute or so. 

At this state I'd say the debugger is very much unusable. It might be better to postpone the feature and rather have it working properly in the next release.

ibuziuk:
@jcantrill please, take a look at @jeffmaury last comment

> Reproducted the timeout problem but seems rather related to router as it can also be reproduced debugging a JavaEE based depmoyement

Can you reproduce the same behaviour for Java EE? If yes, the issue is not Node.js debugger related and should be addressed separately  - https://issues.jboss.org/browse/JBIDE-23422
Feel free to add more details ^

jrichter1:
Tried with an eap-basic project (that should be Java EE). Debug works, unlike with a node.js project.

adietish:
I face the timeout problem too, when debugging. But I also faced this without any debugging involved (nodejs). So I really dont think that this is related to the debugging that Ilya implemented.
I thus advocate us merging this.

adietish:
Here's how the problem is reproducible for me without any debugging being involved:
1. EXEC: create a new application via the nodejs-example. 
2. ASSERT: pod is running
3. EXEC: pick Show In > Web Browser 
4. ASSERT: Browser showing the webpage is opened
5. EXEC: refresh a few times

Result:
File save dialog is opened, prompting me to save a response to my local harddisk. The response is empty. Afaics the file dialog is opened because the mime-type of the response cannot be determined (the response is empty, lenght 0). 
This is exactly how things behave for me when I have the debug session timeouting. Instead of a new page I get the file dialog that prompts me to save the response.

6. EXEC: edit the route (ex. add some labels)
7. EXEC: refresh
Result:
Browser shows the web page

[nodejs page wont load](https://www.youtube.com/watch?v=RQ6GEzEIaII)

PR #1346, opened 2016-10-14
[JBIDE-23362] - Failed test because of JBIDE-22916

jeffmaury:


adietish:
@jeffmaury thx for the fix! Works for me.
Took me quite some time to get it, but I finally understood it I think. 
#select potentially returns null when config is non-null but empty and containerConfig is null. To avoid this, #select needs to know what collection it should return, a list or a set.

PR #1348, opened 2016-10-17
[JBIDE-23372] corrected layout for new app/res wizard

adietish:


adietish:
@jeffmaury plz review

PR #1355, opened 2016-10-25
[JBIDE-23354] - Deploy Docker Wizard: Reset of port mapping whille at least 1 default is present causes NPE

jeffmaury:


adietish:
+1, works for me.

PR #1356, opened 2016-10-25
[JBIDE-23030] fix scaling pods starting from a service

adietish:
# Pull Request Checklist
## General
- Is this a blocking issue or new feature? If yes, QE needs to +1 this PR
## Code
- Are method-/class-/variable-names meaningful?
- Are methods concise, not too long?
- Are catch blocks catching precise Exceptions only (no catch all)?
## Testing
- Are there unit-tests?
- Are there integration tests (or at least a jira to tackle these)?
- Is the non-happy path working, too?
- Are other parts that use the same component still working fine?
## Function
- Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
@@ -361,6 +361,54 @@ public class ResourceUtils {
 				.orElse(null);
 	}
 
+	public static IReplicationController selectByDeploymentConfigVersion(List<IReplicationController> rcs) {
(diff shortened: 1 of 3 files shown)


adietish:
this works for me +1. 
@mlabuda @rhopp can you please verify this critical fix?

rhopp:
You got my +1 too.

adietish:
@jeffmaury I dont follow, which 2 jobs are you talking about (I only see 1 job that is doing the scaling). And then I see a check for negative replica counts at https://github.com/jbosstools/jbosstools-openshift/pull/1356/files#diff-a6aab387e56a343ca4f258b20cc3af57R102 . I must be missing something?
I now also spotted that this is loading resources in the UI thread where it tries to get the replication controller. We should do this in a job and then, in sequence, do the scaling.

PR #1359, opened 2016-10-27
JBIDE-23184: Add Capability to Watch Projects

bdshadow:
# Pull Request Checklist
## General
- Is this a blocking issue or new feature? yes
## Code
- Are method-/class-/variable-names meaningful? 
- Are methods concise, not too long?
- Are catch blocks catching precise Exceptions only (no catch all)?
## Testing
- Are there unit-tests?
- Are there integration tests (or at least a jira to tackle these)?
- Is the non-happy path working, too?
- Are other parts that use the same component still working fine?
## Function
- Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
@@ -11,6 +11,7 @@
 package org.jboss.tools.openshift.internal.core;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -18,6 +19,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
+import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
@@ -54,7 +56,8 @@ public class WatchManager {
 			ResourceKind.ROUTE,
 			ResourceKind.PVC,
 			ResourceKind.SERVICE, 
-			ResourceKind.TEMPLATE
+			ResourceKind.TEMPLATE,
+			ResourceKind.PROJECT
 	};
 	
 	/**
@@ -77,7 +80,21 @@ public class WatchManager {
 		ConnectionsRegistrySingleton.getInstance().addListener(new DeletedConnectionListener());
 	}
 	
+	private void checkResourceKindsExist(String... kinds) {
+		Assert.isTrue(Arrays.asList(KINDS).containsAll(Arrays.asList(kinds)), 
+				"Some of the resources are unwatchable");
+	}
+	
 	public void stopWatch(IProject project, IOpenShiftConnection connection) {
+		stopWatchSafe(project, connection, KINDS);
+	}
+	
+	public void stopWatch(IProject project, IOpenShiftConnection connection, String... kinds) {
+		checkResourceKindsExist(kinds);
+		stopWatchSafe(project, connection, kinds);
+	}
+	
+	private void stopWatchSafe(IProject project, IOpenShiftConnection connection, String... kinds) {
 		for (String kind : KINDS) {
 			AtomicReference<IWatcher> watcherRef = watches.remove(new WatchKey(connection, project, kind));
 			if((watcherRef != null) && (watcherRef.get() != null)) {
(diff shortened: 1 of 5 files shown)


bdshadow:
This PR comes together with: https://github.com/openshift/openshift-restclient-java/pull/234 and should be merged only after it.
This is a new a feature, that works on OS starting from [1.3](https://github.com/openshift/origin/releases/tag/v1.3.0)
@mlabuda please, need your QEs +1 for this PR
@adietish @jeffmaury @jcantrill please, review it and test it properly (delete projects, delete other resources, recreate-and-delete with the same names etc. for both 1.3 and 1.2 versions). 1.2 should work as it was working before (so after project deletion, when user gets an error while operating with resources, user should refresh connection), and 1.3 should work with this new feature.
It seems to me that i tested all the situations, however it's better to go through all this stuff once again.
Looking forward to your comments

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -192,7 +196,13 @@ void connectionChanged(String property, Object oldValue, Object newValue) {
 
 	private ProjectWrapper findProjectWrapper(IResource resource) {
 		synchronized (projects) {
-			return projects.get(resource.getProject());
+			// `projects.get(resource.getProject())` was replaced, because project
```
jcantrill: This was a convenience method that made an additional call to get the project that owns a resource.  It probably should be deprecated and what you have here is a better implementation.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -204,9 +214,11 @@ private void updateWithResources(List<IProject> newValue) {
 			projects.clear();
 			for (IProject r : newValue) {
 				ProjectWrapper existingWrapper = oldWrappers.remove(r);
-
 				if (existingWrapper == null) {
 					ProjectWrapper newWrapper = new ProjectWrapper(this, r);
+					resourceCache.add(r);
```
jcantrill: This seems like the wrong place to begin the watch.  My thought is we would start watch on connection expansion, maybe this is the only place to insert it.  Also, I would not expect to need to provide the project name for a watch; watches provide callbacks based on resource kind and you receive callbacks for all resource of that kind for which you have access.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -77,7 +80,21 @@ private WatchManager() {
...
+		stopWatchSafe(project, connection, KINDS);
+	}
+	
+	public void stopWatch(IProject project, IOpenShiftConnection connection, String... kinds) {
+		checkResourceKindsExist(kinds);
+		stopWatchSafe(project, connection, kinds);
+	}
+	
+	private void stopWatchSafe(IProject project, IOpenShiftConnection connection, String... kinds) {
 		for (String kind : KINDS) {
```
bdshadow: @scabanovich +1, definitely it's my miss

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -46,7 +46,8 @@
 			ResourceKind.REPLICATION_CONTROLLER, 
 			ResourceKind.SERVICE, 
 			ResourceKind.TEMPLATE,
-			ResourceKind.PVC
+			ResourceKind.PVC,
+			ResourceKind.PROJECT
 		};
```
bdshadow: +1, Slava thank you!

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
bdshadow: @jcantrill i understand that it's a rare case, but if we start watching on connection expansion, then if the project is deleted on the server just after creation (when we haven't expanded the connection yet), it would not be deleted in eclipse.
Jeff, i didn't understand about the project names. I provide it nowhere. Or what have i missed?

bdshadow:
testPR

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -237,14 +243,23 @@ protected void handleAdd(ProjectWrapper projectWrapper, IResource newResource) {
...
 	protected void handleRemove(ProjectWrapper projectWrapper, IResource oldResource) {
 		resourceCache.remove(oldResource);
-		Collection<IResource> resources = resourceCache.getResources(oldResource.getProject().getNamespace());
-		projectWrapper.updateWithResources(resources);
+		WatchManager.getInstance().stopWatch(projectWrapper.getWrapped(), projectWrapper.getParent().getWrapped(), oldResource.getKind());
+		if (oldResource instanceof IProject) {
+			synchronized(projects) {
+				projects.remove(oldResource.getName());
+				resourceCache.flush(oldResource.getName());
+				fireChanged();
```
adietish: this could be a substantial bottleneck since we block other threads from getting processed even when we roundtrip to the UI. Is this really necessary?

adietish:
testPR

bdshadow:
testPR

adietish:
+1, works for me. All tests pass.

adietish:
we basically agreed on me merging it without QE ACK, given the exceptional situation and marian will test this tomorrow together with the smoketests

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
bdshadow: here it is: https://issues.jboss.org/browse/JBIDE-23466

adietish: @bdshadow great, thanks

PR #1369, opened 2016-11-17
[JBIDE-23352] - Deploy Docker Wizard: Default routing port selection …

jeffmaury:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

…need more info for users

- Added message to user that round robin will be user if no port used
- Removed Add and Delete buttons

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/OpenShiftUIMessages.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/OpenShiftUIMessages.java
@@ -45,4 +45,5 @@ public class OpenShiftUIMessages extends NLS {
 	
 	public static String InvalidHostNameErrorMessage;
 	public static String EmptyHostNameErrorMessage;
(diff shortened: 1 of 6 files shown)


adietish:
@jeffmaury when editing the ports, it seems like there's an upper limit for the port but you wont get told so in the "Edit..." dialog.
# EXEC: launch app wizard and pick builder image
# EXEC: get to the "Services & Routing Settings" page
# EXEC: select a port mapping and hit "Edit"
# EXEC: in "Service Port" provide "111111" (6 digits) and hit "OK"
Result:
The port being stored is "11111" (5 digits)

adietish:
Works for me. It is not the most elegant solution in terms of usablity though. The cursor is moved to the beginning of the spinners text-field once you pass the maximum value (ex. have 6888 -> add '8'. 68888 exceeds the maximum valid value of 65535. -> the value gets automatically resetted to 65535 and the cursor gets moved to the start of the text field).
Most likely the best solution would be to use a text field for both ports and use validators for both, so that the UI gets consistent. We currently have a spinner for the service port and a text-field for the pod port.

adietish:
merging.

PR #1367, opened 2016-11-14
jbide-23470: Server Adapter: stuck restarting in debug mode

bdshadow:
# Pull Request Checklist
## General
* Is this a blocking issue or new feature? no

## Code
* Are method-/class-/variable-names meaningful?
* Are methods concise, not too long?
* Are catch blocks catching precise Exceptions only (no catch all)?

## Testing
* Are there unit-tests?
* Are there integration tests (or at least a jira to tackle these)?
* Is the non-happy path working, too?
* Are other parts that use the same component still working fine?

## Function
* Does it work?

PR #1375, opened 2016-11-23
jbide-23740: Server Adapter: stuck restarting in debug mode

jeffmaury:
This is an update of PR #1367

 # Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/RSync.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/RSync.java
@@ -68,7 +68,7 @@ public class RSync {
 				if (shouldSync) {
 					for (IPod pod : service.getPods()) {
 						try {
-						    if (pod.getStatus() == "Running") {
+						    if ("Running".equals(pod.getStatus())) {
 	                            syncPodToDirectory(pod, podPath, deployFolder, consoleWriter);
 						    }
 						} catch (IOException | OpenShiftException e) {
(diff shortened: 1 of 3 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/RSync.java
```diff
@@ -68,7 +68,9 @@ protected void runOCBinary(MultiStatus multiStatus) {
 				if (shouldSync) {
 					for (IPod pod : service.getPods()) {
 						try {
-							syncPodToDirectory(pod, podPath, deployFolder, consoleWriter);
+						    if (pod.getStatus() == "Running") {
```
adietish: we should make osjc provide enums for the pod states. I'll file an issue.

jeffmaury:
@adietish I was not able to reproduce your bug and I think this is not related to this PR but rather to the JS integration. However, I found an important problem leading to no sync being performed anymore. So please retest

adietish:
Eap worked for me flawlessly. 
I have new symptoms appear when debugging a nodejs app: the V8 debugger cant connect, complaining that it cannot read the version. I suspect some issue with the pods being replaced (old pod terminated, new pod started). I'll file the issue for nodejs as soon as I got the way to replicate it consistently.

adietish:
I corrected the JBIDE-23740 to JBIDE-23470 in the commit message.

PR #1380, opened 2016-12-20
Clean code during investigation on JBDS-4227

apupier:
- remove useless check for non-nullity as variable already used
previously
- use constants instead of hardcoded and duplicated literals
- leverage genericity
- remove useless assignment
- invert equals to avoid potential NPEs
- use method reference instead of anonymous inner class

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/VagrantPoller.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/VagrantPoller.java
@@ -70,12 +70,7 @@ public class VagrantPoller implements IServerStatePoller2 {
 		launchThread();
 	}
 	protected void launchThread() {
-		Thread t = new Thread(new Runnable(){
-			@Override
-			public void run() {
-				pollerRun();
-			}
-		}, "CDK Poller"); //$NON-NLS-1$
+		Thread t = new Thread(this::pollerRun, "CDK Poller"); //$NON-NLS-1$
 		t.start();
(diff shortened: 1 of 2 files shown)


adietish:
testPR

adietish:
looks good to me, @robstryker, dare to review?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/VagrantPoller.java
```diff
@@ -70,12 +70,7 @@ public void beginPolling(IServer server, boolean expectedState) throws PollingEx
 		launchThread();
 	}
 	protected void launchThread() {
-		Thread t = new Thread(new Runnable(){
-			@Override
-			public void run() {
-				pollerRun();
-			}
-		}, "CDK Poller"); //$NON-NLS-1$
+		Thread t = new Thread(this::pollerRun, "CDK Poller"); //$NON-NLS-1$
```
adietish: why not use a job?

apupier:
> But given the very long methods and all the for's with nested if/else etc. I bet that there's plenty of room for another simple cleanup: split stuff up into smaller more concise methods.

Agree. I didn't take time to understood the code well enough to split into smaller methods. Unless I need to go deeper to resolve my issue JBDS-4227 I won't spend time for that.

adietish:
@apupier ok, thanks! I'll add further changes according to the issues reported in Sonar.

adietish:
added my changes on top #1384

PR #1383, opened 2016-12-26
JBIDE-23622 - Build Log: impossible to get logs once the build pod is completed

jeffmaury:
- Added Completed as supported pod state

### DIFF
--- a/tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/handler/PodLogsHandlerTest.java
+++ b/tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/handler/PodLogsHandlerTest.java
@@ -1,0 +1,132 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.openshift.test.handler;
+
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.core.commands.ExecutionEvent;
+import org.eclipse.osgi.util.NLS;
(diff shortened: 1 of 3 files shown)


jcantrill:
Lgtm. Is there a test we can write to go with the change

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/handler/PodLogsHandlerTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public static class TestPodLogsHandler extends PodLogsHandler {
+
+        @Override
+        public <T> T getSelectedElement(ExecutionEvent event, Class<T> klass) {
+            return super.getSelectedElement(event, klass);
+        }
+
+        @Override
+        protected void showLogs(IPod pod, ExecutionEvent event) {
```
adietish: @jeffmaury looks like sonar is right and we should be able to simply remove this method?

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/handler/PodLogsHandlerTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    public static class TestPodLogsHandler extends PodLogsHandler {
+        @Override
+        public Object execute(ExecutionEvent event) throws ExecutionException {
+            handleEvent(event);
+            return null;
+        }
+
+        @Override
+        public <T> T getSelectedElement(ExecutionEvent event, Class<T> klass) {
```
adietish: @jeffmaury remove this method?

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/handler/PodLogsHandlerTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import com.openshift.restclient.model.IBuild;
+import com.openshift.restclient.model.IContainer;
+import com.openshift.restclient.model.IPod;
+
+@RunWith(MockitoJUnitRunner.class)
+public class PodLogsHandlerTest {
+
```
adietish: @jeffmaury added tests, awesomeness, thx!

adietish:
testPR

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/handler/PodLogsHandlerTest.java
jeffmaury: No because then the test class is not allowed to call this method because its protected

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/handler/PodLogsHandlerTest.java
jeffmaury: No because then the test class is not allowed to call this method because its protected

PR #1385, opened 2017-01-10
[JBIDE-23707] clarified loading job messages in explorer

adietish:
# Pull Request Checklist
## General
* Is this a blocking issue or new feature? If yes, QE needs to +1 this PR

## Code
* Are method-/class-/variable-names meaningful?
* Are methods concise, not too long?
* Are catch blocks catching precise Exceptions only (no catch all)?

## Testing
* Are there unit-tests?
* Are there integration tests (or at least a jira to tackle these)?
* Is the non-happy path working, too?
* Are other parts that use the same component still working fine?

## Function
* Does it work?

PR #1386, opened 2017-01-11
[JBIDE-23707] clarified loading job messages in explorer

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
corrected the messages including OpenShift to provide a context in job messages as suggested in #1385

PR #1387, opened 2017-01-12
[JBIDE-23514] - Server adapter: adapters for non-java app should not have a JMX child (double clicking it causes error)

jeffmaury:


adietish:
+1 as first step. But on the long run, we need means to identify if JMX is enabled for the particular OpenShift application.

PR #1392, opened 2017-01-17
JBIDE-23675 - Move openshift reddeer wrapper classes from integration-tests repo to jbosstoos-openshift repo.

psrna:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/features/org.jboss.tools.openshift.test.feature/feature.xml
+++ b/features/org.jboss.tools.openshift.test.feature/feature.xml
@@ -20,6 +20,12 @@
    </license>
 
    <plugin
+         id="org.jboss.tools.openshift.reddeer"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+   <plugin
          id="org.jboss.tools.openshift.test"
          download-size="0"
          install-size="0"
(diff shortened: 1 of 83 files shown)


psrna:
depends on https://github.com/jbosstools/jbosstools-central/pull/392

adietish:
testPR

psrna:
testPR

adietish:
FYI: dependency to central was removed, tests rewritten to test via File > New > Other > OpenShift which is equivalent to opening via central. Thus the dependency to central could be dropped.

psrna:
@adietish, dependency to central was removed from openshift.reddeer test-framework. The only dependency was CentralIsLoaded condition, which is now here: 

https://github.com/jbosstools/jbosstools-openshift/blob/master/test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/condition/CentralIsLoaded.java

The tests have not been rewritten as it turned out that the wizard launched from central behaves not the same as from File -> New -> Other -> Openshift. Here is regression bug: https://issues.jboss.org/browse/JBIDE-23704

So there obviously is a need for having tests for this scenario. And central is usually the first thing that is seen by a user.

PR #1397, opened 2017-01-26
(itests, test-framework) Move Project creation logic from requirement to separate class.

rhopp:
So this logic can be consumed by tests.

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/requirement/OpenShiftProjectRequirement.java
+++ b/test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/requirement/OpenShiftProjectRequirement.java
@@ -11,7 +11,6 @@
 package org.jboss.tools.openshift.reddeer.requirement;
 
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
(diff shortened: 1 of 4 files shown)


adietish:
testPR

test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/utils/ProjectNativeUtils.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jboss.reddeer.common.wait.WaitUntil;
+import org.jboss.tools.openshift.core.connection.Connection;
+import org.jboss.tools.openshift.reddeer.condition.OpenShiftProjectExists;
+import org.jboss.tools.openshift.reddeer.requirement.OpenShiftResourceUtils;
+
+import com.openshift.restclient.ResourceKind;
+import com.openshift.restclient.model.IProject;
+import com.openshift.restclient.model.project.IProjectRequest;
+
+public class ProjectNativeUtils {
```
adietish: I see the added value of extracting this away 
I find this class name confusing. Out of the blue I would have thought that these are utils for Eclipse "native" projects, aka workspace projects. Dont you think so?
My pick would be to either move these 2 methods into the existing OpenShiftResourceUtils or have the class called OpenShiftProjectUtils (which would also match the other naming scheme of the general purpose resource utils class OpenShiftResourceUtils). Dont you think so?

adietish:
weird, jenkins fails to run, misses the activator in egit tests.

test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/utils/ProjectNativeUtils.java
rhopp: I'll address that.

adietish:
testPR

rhopp:
I've created package o.j.t.openshift.reddeer.utils.v3, where I moved OpenShiftResourceUtils  and ProjectNativeUtils. And I also renamed both of them to OpenShift3Native<Resource/Project>Utils. I wanted to be able to distinguish whether class uses UI, or core api from the class name and I didn't think of anything better than "Native" for the core-api ones.

PR #1402, opened 2017-02-02
[JBIDE-23819] fixing oc binary version parsing

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@rhopp can you please review this critical fix?

adietish:
@jeffmaury +1 please?

rhopp:
So we are still expecting the binary to be named "oc", right? Nothing else than "oc" will not be accepted?

rhopp:
If the premise from my last comment is valid, I'm OK with this PR (I've tried to detect 3.4.2 and v1.4.1+3f9807a successfully).

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/internal/core/preferences/OCBinaryVersionValidatorTest.java
```diff
@@ -17,39 +17,54 @@
...
 		assertSameVersion("1.2.3", "oc v1.2.3");
 		assertSameVersion("1.2.0.foo", "oc v1.2-foo");
 		assertSameVersion("1.0.0", "oc v1");
+		assertSameVersion("1.4.0.rc1", "oc v1.4.0-rc1+b4e0954");
 	}
-	
+
+	@Test
+	public void testParseOCEnterpriseVersion() {
+		assertSameVersion("3.4.1.2", "oc v3.4.1.2");
```
adietish: false positive. this is a test.

adietish:
@rhopp thanks for the quick review.
Your comment is correct, we're still only allowing binaries which are called "oc". Having this fully flexible would be only possible if we changed further occurrences than just this validator (I had my binary called "oc 2" and using it failed in other places than the validator). I'd be willing to pursue this goal but we'd have to do it in a separate jira/PR

rhopp:
I would wait for the result of https://github.com/openshift/origin/issues/12768 and then we can decide what action we will take.

adietish:
@rhopp sure, +1. Even more since this would be a non-critical enhancement for an edge-case, since most users would have it named oc and even if not (ex. to have several oc versions at once), they could symlink.

adietish:
testPR

adietish:
testPR

PR #1410, opened 2017-02-09
Test-framework: wait for project creation based on displayed name

jrichter1:
The tree item lookup uses displayed name, but the method would use project name and error every time these names differ.

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


jrichter1:
@adietish @jeffmaury could you take a look at this? I'd love to use projects with different name and displayed name in a requirement :)

PR #1411, opened 2017-02-09
Test-framework: Move acceptSSLCertificate method to a public place

jrichter1:
I've found some uses for this method outside the secure storage handler. TestUtils seems like a good place to make it public.

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


jrichter1:
@adietish @jeffmaury looking for somebody to check this PR. It's just a method moved and made public.
The exception was swallowed intentionally but if don't like to be yelled at by sonar, I can log it or something.

PR #1412, opened 2017-02-10
[JBIDE-22362] - Server Adapter: Static changes done to nodejs application

jeffmaury:


adietish:
works for me.
I could do changes in index.html, but also changes in javascript.
ex. in nodejs example:
* copy index.html to index2.html
* change index2.html to have different content so that you know what html page is being rendered.
* change server.js to render index2.html: 
at line 75:
```
res.render('index.html', { pageCountMessage : null});
```
to 
```
res.render('index2.html', { pageCountMessage : null});
```
works!
+1

PR #1415, opened 2017-02-14
JBIDE-23603: do a clean publish for eap

bdshadow:
I'm really not sure about the correctness of the fix, but it works for me.

The thing is that, despite the fact that the publish kind is called `IServer.PUBLISH_CLEAN` it doesn't clean all the deployed files, but just a changed one: here is rsync log from Console View after the changes in Member.java:
> sent 46 bytes  received 1,375 bytes  2,842.00 bytes/sec
total size is 8,535,536  speedup is 6,006.71
sending incremental file list
ROOT.war.dodeploy
ROOT.war/WEB-INF/classes/org/jboss/as/quickstarts/kitchensink/model/Member.class
sent 2,147 bytes  received 190 bytes  1,558.00 bytes/sec
total size is 8,535,531  speedup is 3,652.35

i'm mostly unsure about the condition `isEapProfile()`. Is it ok to check if it is a EAP project like this?

@adietish @robstryker @jeffmaury please, take a look and test


bdshadow:
testPR

PR #1416, opened 2017-02-14
JBIDE-23896 - allow user to choose hypervisor as first class option

robstryker:
# Pull Request Checklist
## General
* Is this a blocking issue or new feature? If yes, QE needs to +1 this PR  - Doesn't need +1 

## Code
* Are method-/class-/variable-names meaningful? yes
* Are methods concise, not too long? yes
* Are catch blocks catching precise Exceptions only (no catch all)? yes

## Testing
* Are there unit-tests? no
* Are there integration tests (or at least a jira to tackle these)? no
* Is the non-happy path working, too? yes
* Are other parts that use the same component still working fine?yes

## Function
* Does it work?

yes it works


jeffmaury:
I think something weird: if the hypervisor attribute is not set, then the default value on Windows is HyperV. However, if the user run the setup-cdk command with Virtualbox then starting the CDKv3 server will fail

PR #1417, opened 2017-02-15
JBIDE-23812 - fix loading and undo situation for server editors

robstryker:
# Pull Request Checklist
## General
* Is this a blocking issue or new feature? If yes, QE needs to +1 this PR - very annoying UI stuff

## Code
* Are method-/class-/variable-names meaningful? yes
* Are methods concise, not too long? yes
* Are catch blocks catching precise Exceptions only (no catch all)? yes

## Testing
* Are there unit-tests? no, it's UI
* Are there integration tests (or at least a jira to tackle these)? idk
* Is the non-happy path working, too? kinda... it's better than it was?
* Are other parts that use the same component still working fine? yes

## Function
* Does it work? mostly. 

First, the code was needlessly complex. It was chaining jobs when it didn't need to. It could do it all in one job and accumulate the results.  Also, anonymous inner functions are ok if they're short, but not when it starts to clutter the code. 

Eventually I figured out that the model was still in "initializing" mode in the error situation.  This was the bulk of the fix. The rest is reorganization and cleanup because I couldn't understand what the hell was going on.  It's also possible that these cleanup changes, though, actually fixed something. 

This patch also includes a fix for some layout stuff, specifically in the connection portion of the widget. I would often get the connection combo box on top of the connection label. The code there, also, was needlessly complex, and has been cleaned up. It no longer sits on top of each other. Part of the fix here was adding a wrapper composite for the combo (ugly I know) and setting an ideal size for it. It now works as expected. 

There's still small problems with this code. On occasion the 'undo' stack gets out of date, and I haven't been able to figure out why. The symptom here is that if you make a change, and undo it, it properly undoes the change, but the editor is still marked as dirty. This isn't always the case, but it happens on occasion based on some initial condition I can't figure out. 

Either way, this is much better than before.

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
@@ -20,6 +20,8 @@ import org.eclipse.core.databinding.DataBindingContext;
 import org.eclipse.core.databinding.beans.BeanProperties;
 import org.eclipse.core.databinding.conversion.Converter;
 import org.eclipse.core.databinding.observable.value.IObservableValue;
+import org.eclipse.core.databinding.observable.value.IValueChangeListener;
+import org.eclipse.core.databinding.observable.value.ValueChangeEvent;
 import org.eclipse.core.databinding.validation.IValidator;
 import org.eclipse.core.databinding.validation.ValidationStatus;
 import org.eclipse.core.resources.IProject;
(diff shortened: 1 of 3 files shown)


adietish:
i tested it, this partially works for me.
I can change the connection now and the "changed" marker for the editor now appears. But it (the "changed" marker) will only appear once I the initial job that tries to load the resources from server1  fails. Changing the Eclipse project or the server still isnt working though. I can browse the Eclipse projects, but it wont get set to the widget once I selected a project in the popup. Same story for the service: I can browse and select the services (once I changed the connectino to server2), but the selected service wont appear in the editor.

robstryker:
Unfortunately I can't figure out in databinding how to do that...  The question is are you going to +1 this or request more changes?

robstryker:
testPR

adietish:
@robstryker thx, works for me.

adietish:
testPR

PR #1420, opened 2017-02-15
[JBIDE-23862] prefs: dont allow oc on path with " "

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.client/.classpath
+++ b/plugins/org.jboss.tools.openshift.client/.classpath
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry exported="true" kind="lib" path="lib/openshift-restclient-java-5.6.0.Final.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/openshift-restclient-java-5.7.0-SNAPSHOT.jar"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry exported="true" kind="lib" path="lib/okhttp-3.3.1.jar"/>
--- a/plugins/org.jboss.tools.openshift.client/META-INF/MANIFEST.MF
+++ b/plugins/org.jboss.tools.openshift.client/META-INF/MANIFEST.MF
@@ -6,7 +6,7 @@ Bundle-Version: 3.4.0.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-Vendor: %Bundle-Vendor
 Bundle-Localization: plugin
-Bundle-ClassPath: lib/openshift-restclient-java-5.6.0.Final.jar,
+Bundle-ClassPath: lib/openshift-restclient-java-5.7.0-SNAPSHOT.jar,
  lib/okhttp-3.3.1.jar,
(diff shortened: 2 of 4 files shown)


adietish:
@jeffmaury @bdshadow couldnt fix the issue with " " in the path to the oc binary. I am thus for now suggesting that we error in the preferences if such a path is being used.

bdshadow:
changed this PR a bit, i think it should be also merged today before the code freeze

jeffmaury:
I think we need @adietish comment before merging because I remember he had remarks about it

adietish:
@bdshadow @jeffmaury by looking at the PR things look fine. I was investigating this and didnt find a proper solution to fix spaces in paths for OSX (I tried minimalistic code). Let me test it properly.

adietish:
testPR

adietish:
@jeffmaury @bdshadow it looks like there are only issues on MacOS when the path to oc contains spaces (in jira @jeffmaury said that it works with spaces in Win10, @robstryker told me that it also works with spaces on Linux - @bdshadow can you confirm this please? @jeffmaury can you please specify in jira what you were referring to when talking of problems when the path to the workspace contains spaces?)
If it's true, that only MacOS has this problem, then the validaiton should only fail for spaces on MacOS (it currently does for all platforms).
Thoughts?

bdshadow:
@adietish i have problems of Fedora. I put the oc into "/home/dbocharo/oc folder with space".
The version is detected fine in the Preferences windows. However, when i tried to create a server adapter (where oc is actually used), i got the following error:
```java
org.eclipse.core.runtime.CoreException: Error while publishing server jboss-eap70-openshift (Service) at OpenShift 3 (10.1.2.2).  Could not sync all pods to folder /home/dbocharo/dev/openshift/runtime-eclipse_with_openshift/.metadata/.plugins/org.jboss.ide.eclipse.as.core/sample-project@jboss-eap70-openshift/deploy
	at org.jboss.tools.openshift.core.server.behavior.OpenShiftPublishController.publishStart(OpenShiftPublishController.java:73)
	at org.jboss.ide.eclipse.as.wtp.core.server.behavior.CachedPublisherProfileBehavior.publishStart(CachedPublisherProfileBehavior.java:47)
	at org.eclipse.wst.server.core.model.ServerBehaviourDelegate.publish(ServerBehaviourDelegate.java:942)
	at org.eclipse.wst.server.core.model.ServerBehaviourDelegate.publish(ServerBehaviourDelegate.java:774)
	at org.eclipse.wst.server.core.internal.Server.publishImpl(Server.java:3172)
	at org.eclipse.wst.server.core.internal.Server$PublishJob.run(Server.java:345)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:56)
Contains: Does your OpenShift binary location exist? Error starting process: Cannot run program ""/home/dbocharo/oc": error=2, No such file or directory
```

bdshadow:
I guess the problem happens [here in openshift-restclient](https://github.com/openshift/openshift-restclient-java/blob/c725ccf1069b5e92aa58905e419992d0a945bb9e/src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java#L156) where we create the command to execute and split it with spaces. The final `oc rsync` with parameters turns into a mess

bdshadow:
and i'm sure it doesn't work on any OS. At least, it must fail here.

bdshadow:
So your check should be added. It's a good quick fix to avoid these problems, even if we plan changes in restclient in the future

adietish:
went back to simply bumping the openshift-restclient-java since this fixes the issue. No change in JBoss Tools.
Thanks for being tenacious @bdshadow!

PR #1422, opened 2017-02-16
Jbide 23490

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@jeffmaury seems to work for me. Can you please also test this?

adietish:
since this is a new feature, ideally QE should also +1 this, @jrichter1 can you please?

jrichter1:
What I (perhaps wrongly) understand from the jira is, that I should be able to create and use server adapters linked to pods, even if a service is not available. Please correct me if I got this wrong.

Anyway, the findings so far:
 - I can create server adapters from pods - and they tell me what resource they are linked with (service or deployment config)
 - adapters created when a service exists work fine
 - adapters without a service look like their functionality does something, but I can't get to the actual deployed application to see what I've published - things like show in... just don't do anything

adietish:
@jrichter1 I believe that that's how it is supposed to work. Since there's no service, there's no route nor way to get to view the content from the outside. @jeffmaury ?

jeffmaury:
This feature is focused on the publish and debug. This is also targeted to non UI deployement (FUSE). There is no direct way from the UI to invoke the deployment but this is still possible through the port forward. We could have this done automatically in the future if a service is not available BTW having a service is not enough we must have a route so this is not new

adietish:
@jeffmaury @jrichter1 looks like things are thus working as expected. merging.

jeffmaury:
# **PLEASE DO NOT MERGE**
 until I gave an ok. Given the importance of the change, we should wait for the next sprint, we are to close to the GA

jeffmaury:
Closed by #1427

PR #1423, opened 2017-02-16
JBIDE-23903: Start watching resources from openshift server adapter

bdshadow:
I think it's worth starting to watch projects when we need a service in the Server Adapter.

@adietish @jeffmaury this patch fixes the bug for me. Please, test it


adietish:
+1, works for me

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
```diff
@@ -366,7 +367,9 @@ public static IService getService(IServerAttributes server, Connection connectio
 		}
 		String projectName = OpenShiftResourceUniqueId.getProjectName(uniqueId);
 		List<IService> services = connection.getResources(ResourceKind.SERVICE, projectName);
-		return OpenShiftResourceUniqueId.getByUniqueId(uniqueId, services);
+		IService service = OpenShiftResourceUniqueId.getByUniqueId(uniqueId, services);
+		WatchManager.getInstance().startWatch(service.getProject(), connection);
```
adietish: what if service is null? Shouldnt we null check here?

adietish: I double checked and it seems like starting the watcher twice would not harm, a check whether it already exists is in place preventing multiple watchers per project.

PR #1425, opened 2017-02-17
[JBIDE-23961] dont consider dev_mode as "debug is enabled"

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package org.jboss.tools.openshift.core.server.behavior;
 
+import static org.jboss.tools.openshift.core.server.OpenShiftServerUtils.toCoreException;
+
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
@@ -33,6 +35,7 @@ import org.eclipse.debug.core.model.IDebugTarget;
 import org.eclipse.jdt.debug.core.IJavaDebugTarget;
 import org.eclipse.jdt.debug.core.IJavaHotCodeReplaceListener;
 import org.eclipse.jdt.launching.SocketUtil;
+import org.eclipse.osgi.util.NLS;
 import org.eclipse.wst.server.core.IModule;
 import org.eclipse.wst.server.core.IServer;
 import org.eclipse.wst.server.core.ServerUtil;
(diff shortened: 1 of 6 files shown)


adietish:
@robstryker @jeffmaury @bdshadow can you please verify this?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -173,93 +177,98 @@ private void sleep(long t) {
...
 			beh.setRunMode(mode);
 			beh.setServerStarted();
 		} else {
 			beh.setServerStopped();
-			((ControllableServerBehavior)getControllableBehavior()).setRunMode(null);
+			beh.setRunMode(null);
 		}
 	}
 
+	private void setModulesPublishing() {
```
robstryker: THis method doesn't set that the modules *are* publishing, but rather, that they require publishing. What I mean is, this action doesn't begin any publish actions at all, nor does it mark them as being currently published. THis marks them as requiring a publish.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -173,93 +177,98 @@ private void sleep(long t) {
...
+		IModule[] modules = getServer().getModules();
+		for (int i = 0; i < modules.length; i++) {
+			((Server) getServer()).setModulePublishState(
+					new IModule[] { modules[i] }, IServer.PUBLISH_STATE_FULL);
+		}
+	}
+	
+	private void publishServer() {
```
robstryker: I would comment this method by indicating it kicks off an incremental publish action in a job.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -173,93 +177,98 @@ private void sleep(long t) {
...
+			}
+		}.schedule(PUBLISH_DELAY);
+	}
 
 	private void startDebugging(IServer server, IDeploymentConfig dc, DebuggingContext debugContext,
 			IProgressMonitor monitor) throws CoreException {
 		int remotePort = debugContext.getDebugPort();
-		if( remotePort == -1 ) {
+		if (remotePort == -1) {
 			debugContext.setDebugPort(8787);//TODO get default port from server settings?
```
robstryker: Will this hard-coded value stay? Or is it temporary?

adietish: this is temporary. This will be fixed in the jira(s) where we want to refactor the server adapters, separating different types

adietish:
this needs bumping test plugin and feature: #1444

PR #1427, opened 2017-02-22
[JBIDE-23490] Server adapter: provide ability to link pods/'deployment config'/'replication controller'(?) when no service is available

jeffmaury:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


bdshadow:
@jeffmaury found a few problems - [see my comment in JIRA](https://issues.jboss.org/browse/JBIDE-23490?focusedCommentId=13368338&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13368338).

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
```diff
@@ -92,26 +95,44 @@
...
+	                                                                                                                                   ResourceKind.POD));
+	
+	/**
+	 * Checks if the resource is allowed for OpenShift server adapter.
+	 * 
+	 * @param resource the OpenShift resource
+	 * @return true if allowed
+	 */
+	public static boolean isAllowedForServerAdapter(IResource resource) {
+	    return SERVER_ADDAPTER_ALLOWED_RESOURCE_TYPESS.stream().anyMatch(kind -> kind.equals(resource.getKind()));
```
jeffmaury: Types are not the same. Am I missing something?

bdshadow: `return SERVER_ADDAPTER_ALLOWED_RESOURCE_TYPESS.contains(resource.getKind());` should work fine

bdshadow:
@jeffmaury 
1. i scale up pods to 5
2. restart/restart in debug mode a server adapter, created from pod
firstly all pods are terminated and a new one is created for running/debugging purpose, and it is scaled up to 5 pods again.
Is it an intended behavior that pods are terminated and created again?

jeffmaury:
Yes because we need to toggle the debug mode

bdshadow:
@jeffmaury no, it isn't. Earlier it was restarting one pod for debug process

jeffmaury:
Are you sure ? Because updating the DC will cause the pods to be restarted by OCP

bdshadow:
@jeffmaury sorry, double-checked it. It was the same

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ProjectWrapper.java
```diff
@@ -65,6 +76,15 @@ protected void postUpdate(Collection<IResource> resources, Map<IResource, Abstra
 					resources);
 			newService.initWithResources(relatedResources);
 			newWrapper = newService;
+		} else if ((ResourceKind.DEPLOYMENT_CONFIG.equals(r.getKind()) ||
+		           (ResourceKind.REPLICATION_CONTROLLER.equals(r.getKind()) && !r.isAnnotatedWith(OpenShiftAPIAnnotations.DEPLOYMENT_CONFIG_NAME))) &&
+		           ServiceResourceMapper.getServices((IReplicationController) r, resources).isEmpty()) {
+		        ReplicationControllerWrapper dcWrapper = new ReplicationControllerWrapper(this, (IReplicationController) r);
+		        Collection<IResource> relatedResource =
```
bdshadow: Code from lines [83](https://github.com/jbosstools/jbosstools-openshift/pull/1427/files#diff-d73789e8250b3ee63bf704c4edf0e588R83) and [62](https://github.com/jbosstools/jbosstools-openshift/pull/1427/files#diff-d73789e8250b3ee63bf704c4edf0e588R62) can be separated in a method. 

also please, format it, it's hard to read

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ResourceDetailViews.java
```diff
@@ -0,0 +1,198 @@
...
+            IResource resource = (IResource) value;
+			namespaceText.setText(resource.getNamespace());
+			String labels = StringUtils.toString(resource.getLabels());
+            labels = org.apache.commons.lang.StringUtils.defaultString(labels); //replaces null by empty string
+			labelsText.setText(labels);
+        }
+
+		@Override
+		public boolean isViewFor(Object object) {
+			return object instanceof IResource;
```
bdshadow: Shouldn't `IProject` be here?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -119,8 +121,8 @@ public void launch(ILaunchConfiguration configuration, String mode,
 		
 		
 		try {
-			IDeploymentConfig dc = OpenShiftServerUtils.getDeploymentConfig(server);
-			toggleDebugging(mode, monitor, beh, server, dc);
+			IReplicationController rc = OpenShiftServerUtils.getReplicationController(server);
```
bdshadow: `rc` can be null here. Then it can fall in `toggleDebugging` (`debugContext` = null) -> `startDebugging` with NPE

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ResourceDetailViews.java
jeffmaury: No this abstract class is for resources that are inside a project

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/util/OpenShiftResourceUniqueId.java
```diff
@@ -61,13 +63,26 @@ public static String getResourceName(String uniqueId) {
...
+        }
 	}
 
+    public static String getKind(String uniqueId) {
+        if (StringUtils.isEmpty(uniqueId)) {
+            return null;
+        }
+        String[] comps = uniqueId.split("@");
+        if (comps.length == 2) {
+            return ResourceKind.SERVICE;
```
jeffmaury: Because of the old naming schema

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/util/OpenShiftResourceUniqueId.java
```diff
@@ -61,13 +63,26 @@ public static String getResourceName(String uniqueId) {
...
+        } else {
+            return comps[1];
```
jeffmaury: This should not happen

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -176,17 +182,17 @@ private void updateDeploymentConfigValues(IDeploymentConfig deploymentConfig,
...
-				deploymentConfig.setContainers(containers);
+				replicationController.setContainers(containers);
 			}
 		}
 		//TODO the list of env var to set in debug mode should probably be defined in the server settings instead
-		deploymentConfig.setEnvironmentVariable(DEBUG_PORT_KEY, String.valueOf(debugContext.getDebugPort()));
-		deploymentConfig.setEnvironmentVariable(DEV_MODE_KEY, String.valueOf(debugContext.isDebugEnabled()));//for node
-		deploymentConfig.setEnvironmentVariable(DEBUG_KEY, String.valueOf(debugContext.isDebugEnabled()));//for eap
+		replicationController.setEnvironmentVariable(DEBUG_PORT_KEY, String.valueOf(debugContext.getDebugPort()));
+		replicationController.setEnvironmentVariable(DEV_MODE_KEY, String.valueOf(debugContext.isDebugEnabled()));//for node
```
jeffmaury: Will be done by Dmitry's refactoring

adietish: @bdshadow can you please take note of this and make sure we take care of it in your refactoring for https://issues.jboss.org/browse/JBIDE-23900 ?

adietish:
@psrna @rhopp we should have QE +1 this new feature, please.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenshiftJMXConnectionProvider.java
```diff
@@ -50,11 +53,11 @@ public String getId() {
 	@Override
 	protected IConnectionWrapper createConnection(IServer server) {
 		IConnection openshiftCon = OpenShiftServerUtils.getConnection(server);
-		IService service = OpenShiftServerUtils.getService(server);
+		IResource resource = OpenShiftServerUtils.getResource(server);
```
adietish: what if resource == null? I see a possible NPE at line 59 as a result of it.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/PodDeploymentPathProvider.java
```diff
@@ -51,18 +51,22 @@
...
 		int imageDigestIndex = imageRef.indexOf(DOCKER_IMAGE_DIGEST_IDENTIFIER);
 		if (imageDigestIndex > 0) {
 			String imageDigest = imageRef.substring(imageDigestIndex);
 			IResource imageStreamTag = getImageStreamTag(connection, project, imageRef, imageDigest);
 			return getPodPath(imageStreamTag);
 		} else {
-			
-			return "";
+		    String imageMetaData = getImageMetaData(imageRef, project);
+		    if (imageMetaData != null) {
```
adietish: this if/else block seems weird:
#getPodPath returns null if the given imageMetaData is null. So this if/else blocks seems to effectively change the semantics by return "" instead. But this isnt consistent: at line 62 this isnt transformed. The null that is potentially returned by #getPodPath is returned as is. I thus doubt that this if/else is even required. Am I missing something?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
adietish: @bdshadow good catch. I believe that #startDebugging should thus null-check the debugContext.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -77,50 +78,50 @@ private OpenShiftDebugUtils(ILaunchManager launchManager) {
...
-	public DebuggingContext enableDebugMode(IDeploymentConfig deploymentConfig, DebuggingContext debugContext, IProgressMonitor monitor) throws CoreException {
-		Assert.isNotNull(deploymentConfig);
+	public DebuggingContext enableDebugMode(IReplicationController replicationController, DebuggingContext debugContext, IProgressMonitor monitor) throws CoreException {
+		Assert.isNotNull(replicationController);
 		Assert.isNotNull(debugContext);
 
 		IDebugListener listener = debugContext.getDebugListener();
 		if (debugContext.isDebugEnabled() && listener != null) {
-			IPod pod = getFirstPod(deploymentConfig);
+			IPod pod = getFirstPod(replicationController);
```
adietish: what if pod == null? Shouldnt we then NOT attach a listener?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/util/OpenShiftResourceUniqueId.java
```diff
@@ -61,14 +63,34 @@ public static String getResourceName(String uniqueId) {
...
+	 */
+    public static String getKind(String uniqueId) {
+        if (StringUtils.isEmpty(uniqueId)) {
+            return null;
+        }
+        String[] comps = uniqueId.split(String.valueOf(UNIQUE_ID_PROJECT_NAME_DELIMITER));
+        if (comps.length == 2) {
+            return ResourceKind.SERVICE;
+        } else {
+            return comps[1];
```
adietish: this should only be performed once we checked that there are actually >= 2 elements in comps. We otherwise potentially run into an ArrayIndexOutOfBoundsException.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -77,50 +78,50 @@ private OpenShiftDebugUtils(ILaunchManager launchManager) {
...
 			return;
 		}
-		updateDeploymentConfigValues(deploymentConfig, debugContext);
+		updateDeploymentConfigValues(replicationController, debugContext);
 		
-		IClient client = getClient(deploymentConfig);
+		IClient client = getClient(replicationController);
 		
-		DeploymentConfigListenerJob deploymentListenerJob = new DeploymentConfigListenerJob(deploymentConfig);
+		ReplicationControllerListenerJob deploymentListenerJob = new ReplicationControllerListenerJob(replicationController);
```
adietish: looks like deploymentListenerJob should be renamed to rcListenerJob

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -138,9 +139,14 @@ public void done(IJobChangeEvent event) {
...
+		if (ResourceKind.REPLICATION_CONTROLLER.equals(replicationController.getKind())) {
+		    for(IPod pod : ResourceUtils.getPodsFor(replicationController)) {
+		        client.delete(pod);
+		    }
+		}
 		try {
-			deploymentListenerJob.join(DeploymentConfigListenerJob.TIMEOUT, monitor);
+			deploymentListenerJob.join(ReplicationControllerListenerJob.TIMEOUT, monitor);
 		} catch (OperationCanceledException | InterruptedException e) {
 			e.printStackTrace();
```
adietish: we should replace this e.printStackTrace by proper logging.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -153,9 +159,9 @@ public void done(IJobChangeEvent event) {
 		
 	}
 
-	private void updateDeploymentConfigValues(IDeploymentConfig deploymentConfig,
+	private void updateDeploymentConfigValues(IReplicationController replicationController,
 			DebuggingContext debugContext) {
-		Collection<IContainer> originalContainers = deploymentConfig.getContainers();
+		Collection<IContainer> originalContainers = replicationController.getContainers();
 		if (originalContainers != null && !originalContainers.isEmpty() && debugContext.isDebugEnabled()) {
```
adietish: what if debugContext is null?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/ReplicationControllerListenerJob.java
```diff
@@ -66,7 +66,7 @@ public void connectionChanged(IConnection connection, String property, Object ol
 				String podName = candidate.getName();
 				if (!oldPods.contains(podName) &&
 					"Running".equals(candidate.getStatus()) &&
-					ResourceUtils.containsAll(dcSelector, candidate.getLabels())
+					ResourceUtils.containsAll(selector, candidate.getLabels())
```
adietish: to improve code readability: we could extract this logic (l64-l71), that basically checks if the given pod is a new, running pod to a method #isNewControlledPod and store the pod if the method returns true.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenshiftJMXConnectionProvider.java
jeffmaury: yes and also if the connection is null. Will be reported to the user probably. And this was an old code but will fix it

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/PodDeploymentPathProvider.java
jeffmaury: the if/else is required for deployment from docker images without sha256 id that is without image stream tags. This is what I used when deploying without a deployment config. And null or "" is handled the same way.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
jeffmaury: no because the listener manages the case

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
jeffmaury: Makes no sense for me

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/util/OpenShiftResourceUniqueId.java
jeffmaury: Not going to happen

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
jeffmaury: Maybe but not the topic of this PR

adietish:
@jeffmaury the PR now works for me, for the usecases that I tested. I have a small suggestion, though: When editing the server adapter I have no idea to what resource (service? deployment config?) the adapter is bound to:
image
I can find out though by hitting "Search". The dialog that pops up will have the current resource pre-selected and the details panel for it, tells me what resource type it is. It would be great though if the editor would tell me right out of the bow, without requiring me further clicks.

adietish:
@jeffmaury I also miss the ability to create a server adapter via context-menu. There's afaics no such item in any context menu now:
If there's no service pods are not shown in the explorer, so in consequence there's no "Server Adapter" in the explorer context menu of course.
But then in the Properties, where I can create a server adapter when selecting a service:
image
I should be able to do the same when selecting a deployment config. But there's no such item in the context menu:
image

adietish:
and there's another issue that I found:
1. make sure that you have an existing server adapter that you created using the code without this patch (to verify it, go into <your workspace project>/.settings/org.jboss.tools.openshift.core.prefs and check the key *org.jboss.tools.openshift.Service*. If it has a single "@" then the server adapter for this settings is an adapter that was created before this patch. Adapters created with this patch in place have 2x "@").
2. start your server adapter
Result:
Your server adapter errors complaining that it didnt find the resource
image

3. open up the server adapter editor
Result:
The adapter is missing the "Resource" (or "Service" for the older code).
image

We definitely should provide a migration path for existing server adapters. Otherwise we'll break existing users with existing server adapters.

rhopp:
After a long struggle with other things I got to try this. So far so good. You have my +1 for AM1 release.

adietish:
bumped openshift.js and openshift.js.feature in #1443

PR #1442, opened 2017-03-08
jbide-23494: unit-tests for WatchManager

bdshadow:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@bdshadow the 1st thing that pops into my mind is that you're using powermock which we discussed in JBoss Tools in earlier days: https://issues.jboss.org/browse/LOCUS-40
Powermock is requiring hacked versions of JUnit and Hamcrest and apparently only work if you make sure that the bundle pool is cleared: http://stackoverflow.com/questions/22456767/noclassdeffounderror-org-hamcrest-matchers-using-powermock-osgi#22466499
I'm not against it we should carefully weigh the implications though.
@jeffmaury any opinion on the matter?

bdshadow:
@adietish i didn't use powermock-osgi, because it seems to me that it's not longer maintained: https://code.google.com/archive/p/powermock-osgi/wikis/ReleaseNotes.wiki. All other information i found about it is from 2015 and earlier.
I just used libs, retrieved as [maven dependecies](https://github.com/jbosstools/jbosstools-openshift/pull/1442/files#diff-ee9983427aa147da8288e37915c4b141R50)
Haven't seen this discussion: https://issues.jboss.org/browse/LOCUS-40. Thank you!

I spent really much time to write these tests without Powermock, but it seems to me almost impossible. We have multiple singletons, lots of private stuff (which it better to leave private) and etc. Powermock really helps a lot. And this WatchManager things is really sensitive. We must have such tests.

bdshadow:
@adietish and one more thing, by the way. [I removed locus dependency](https://github.com/jbosstools/jbosstools-openshift/pull/1442/files#diff-2dc6738e0e172ecbbc0173805550a592L15). I remember i had some problems with having locus and pwermock, then i saw that:
- everything i need, and what is used from locus, exists in powermock libs
- found that [locus](https://github.com/jbosstools/jbosstools-locus) was intended for eclipse Orbit and the last commit was 9 months ago.
so i decided to replace it with powermock

adietish:
@bdshadow yes I fully agree, we definitely need unit test for the WatchManager and yes, I agree on all those statics and singletons that make writing tests without PowerMock a huge pain (that's btw. 1 of the reason that make me avoid statics wherever I can).
I'm just asking to weigh the pros/cons carefully.
Can you explain to me what the difference is btw using powermock-libs vs powermock-osgi?

adietish:
@bdshadow Can you please explain to me what the difference is btw using powermock-libs vs powermock-osgi?

adietish:
testPR

bdshadow:
@adietish sorry for making you ask twice. But i had to google a bit to give you a proper answer =)
I consider, that difference between them is the same as between any library and its osgi version. There're lots of information in the internet about "OSGIfying" libraries. The best one i think is: http://stackoverflow.com/questions/4931292/non-osgi-library-usage-in-an-osgi-application. Shortly, the difference isn't so much, if we use it only one bundle, in my opinion.

Once again, it's more obvious to use powermock-osgi for our purposes, but i decided to just import the libs into the classpath, because powermock-osgi isn't so actively maintained at the moment.

One more thing i also think about - is the creation of [this class](https://github.com/jbosstools/jbosstools-openshift/pull/1442/files#diff-579fa6fcbad261fd3491e4bdef1cd045R39). Please, read my JavaDoc for it. I've got a feeling that i wouldn't have such a problem, if i used powermock-osgi, as all these dependencies could be resolved then. Still very unsure about it.

adietish:
@bdshadow can you please point me to where ComparisonStrategy is used implicitly?

adietish:
@bdshadow also dare to address the sonar issues please?

bdshadow:
@adietish look at the stacktrace here: http://pastebin.com/hGD4cB9f, if launch already existing `OpenShiftServerUtilsTest` with `@RunWith(PowerMockRunner.class)`

adietish:
testPR

adietish:
@bdshadowit seems like we now have 1 failing test (verified locally):
```
-------------------------------------------------------------------------------
Test set: org.jboss.tools.openshift.test.internal.core.WatchManagerTest
-------------------------------------------------------------------------------
Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.023 sec <<< FAILURE! - in org.jboss.tools.openshift.test.internal.core.WatchManagerTest
testStopWatch(org.jboss.tools.openshift.test.internal.core.WatchManagerTest)  Time elapsed: 0.005 sec
testStartWatch(org.jboss.tools.openshift.test.internal.core.WatchManagerTest)  Time elapsed: 0.006 sec  <<< FAILURE!
org.mockito.exceptions.verification.TooLittleActualInvocations:
client.watch(<any>, <any>, <any>);
Wanted 12 times:
-> at org.jboss.tools.openshift.test.internal.core.WatchManagerTest.testStartWatch(WatchManagerTest.java:34)
But was 11 times:
-> at org.jboss.tools.openshift.internal.core.WatchManager$WatchListener.connect(WatchManager.java:289)

        at org.jboss.tools.openshift.test.internal.core.WatchManagerTest.testStartWatch(WatchManagerTest.java:34)
```

bdshadow:
@adietish the jenkins build has just finished successfully. Anyway i have an idea of why it is failing sometimes, i'll refactor this test

bdshadow:
@adietish though build wasn't marked "FAILED", just "UNSTABLE", found in the full console output, that this tests also fails. You can find it by searching for "org.jboss.tools.openshift.test.internal.core.WatchManagerTest"

bdshadow:
Fixed `WatchManagerTest` with `timeout`. Don't like this solution, but applied it also [here](https://github.com/jbosstools/jbosstools-openshift/pull/1442/files#diff-98c090f950fe26dd2eeb28740c89730fR66) and [here](https://github.com/jbosstools/jbosstools-openshift/pull/1442/files#diff-2cc876e37773360e2bca80aee2f49107R66). We run lot's of jobs from private nested classes. It's hard to test another way. 
Good thing about this `timeout` method (comparing to `after`) is that that test passes as soon as condition is satisfied even if time isn't finished yet. So it mustn't appeal to the time of test execution so much.

adietish:
testPR

adietish:
testPR

PR #1446, opened 2017-03-13
JBIDE-24081 - Add runtime detection support for cdk3

robstryker:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


jeffmaury:
Works for me however I found this little problem: when minishift is not in the PATH, then the runtime detector generates a problem but the result dialog shows that no runtime have been detected and the problem is not returned to the user.

PR #1459, opened 2017-03-29
[JBIDE-24083] make sure dev_mode is always on for nodejs apps

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
@@ -42,7 +42,6 @@ import org.eclipse.wst.server.core.ServerUtil;
 import org.eclipse.wst.server.core.internal.Server;
 import org.jboss.ide.eclipse.as.core.util.ClassCollectingHCRListener;
 import org.jboss.ide.eclipse.as.wtp.core.server.behavior.AbstractSubsystemController;
-import org.jboss.ide.eclipse.as.wtp.core.server.behavior.ControllableServerBehavior;
 import org.jboss.ide.eclipse.as.wtp.core.server.behavior.IControllableServerBehavior;
 import org.jboss.ide.eclipse.as.wtp.core.server.behavior.ILaunchServerController;
 import org.jboss.ide.eclipse.as.wtp.core.server.behavior.ISubsystemController;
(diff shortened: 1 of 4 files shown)


adietish:
@robstryker @jeffmaury review please? This makes sure dev_mode is always on in nodejs apps since this was missing and thus debugging wasnt enabled.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -118,6 +116,7 @@ public void launch(ILaunchConfiguration configuration, String mode,
         launch.addProcess(new ServerProcess(launch, server, getLabel(launch.getLaunchMode())));
 		beh.setServerStarting();
 		waitForDeploymentConfigAndPods(server, monitor);
+		enableDevModeForNodeJsProject(OpenShiftServerUtils.getDeploymentConfig(server), server);
```
jeffmaury: What if the server is not associated with a DC but rather with an RC

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -185,7 +189,11 @@ private void updateDeploymentConfigValues(IReplicationController replicationCont
 			}
 		}
 		//TODO the list of env var to set in debug mode should probably be defined in the server settings instead
-		replicationController.setEnvironmentVariable(DEBUG_PORT_KEY, String.valueOf(debugContext.getDebugPort()));
+		if (debugContext.isDebugEnabled()) {
+			replicationController.setEnvironmentVariable(DEBUG_PORT_KEY, String.valueOf(debugContext.getDebugPort()));			
+		} else {
```
jeffmaury: I would rather remove the variable

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
adietish: this is the original code that was in place before it was completely removed. 
I believe that changing the dc is the right choice since every rc would be derived from it.
How likely is it that there's no dc but only a rc?

adietish: turns out that the code that I added back into OpenShiftServerUtils#getDeploymentConfig(server) is identical to the currently existing OpenShiftServerUtils#getReplicationController(server). I thus simply did the change as suggested.

adietish:
merged into jbosstools-4.4.x and cherry-picked into master

PR #1460, opened 2017-03-31
[JBIDE-24183] corrected tests for Debugging enablement in context

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@jeffmaury @bdshadow the doubtful part is where debugging is only enabled when the *DEBUG_PORT* env var is set. We know from real-life experience that this is true for nodejs. Looking at the docker image labels though could suggest that 5858 is the default port:
```
"Labels": {
                ...
                "com.redhat.dev-mode": "DEV_MODE:false",
                "com.redhat.dev-mode.port": "DEBUG_PORT:5858",
                ...
```
One can find the very same for the eap docker image:
```
"Labels": {
                ...
               "com.redhat.dev-mode": "DEBUG:true",
                "com.redhat.dev-mode.port": "DEBUG_PORT:8787",
                ...
```
So if :5858 is *NOT* the default port, what is it?
And then logical would only be that eap would behave the same and only enable debugging if *DEBUG_PORT* was set.
The other solution is that the nodejs label is simply wrong.
Toughts?

adietish:
@bdshadow thx for those pointers! very helpful. Digging further into how to properly solve this.

adietish:
merged into jbosstools-4.4.x

PR #1463, opened 2017-04-06
[JBIDE-24232] dont NPE on unknown build strategies

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ResourceDetailsContentProvider.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ResourceDetailsContentProvider.java
@@ -38,6 +38,8 @@ import com.openshift.restclient.model.route.IRoute;
  */
 public class ResourceDetailsContentProvider implements ITreeContentProvider{
 
+	public static final String LABEL_STRATEGY = "strategy";
+
 	@Override
 	public Object[] getChildren(Object node) {
 		if(node instanceof IResource) {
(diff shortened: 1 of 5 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ResourceDetailsContentProvider.java
```diff
@@ -90,6 +90,21 @@ private void getDeploymentConfigChildren(Collection<ResourceProperty> properties
...
+		properties.add(new ResourceProperty("output to", config.getOutputRepositoryName()));
+		List<String> triggers = config.getBuildTriggers().stream()
+				.map(trigger -> trigger.getType().toString())
+				.collect(Collectors.toList());
+		properties.add(new ResourceProperty("build triggers", triggers));
+	}
+
+	private void addStrategyTypeProperties(Collection<ResourceProperty> properties, IBuildStrategy buildStrategy) {
+		if (buildStrategy == null) {
+			properties.add(new ResourceProperty("strategy", "<UNKNOWN>"));
```
bdshadow: I noticed that in other resources from "Template Details" window, if smth is not defined i has: "(Not Provided)" text. Take a look in [ResourceDetailsLabelProvider.getStyledText](https://github.com/jbosstools/jbosstools-openshift/blob/cb4239fc371ac1e70afacea96bd54e0174dc6a79/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/ResourceDetailsLabelProvider.java#L63). So possible it worth making it consitent and make it: `new ResourceProperty("strategy", "")` instead of `new ResourceProperty("strategy", "<UNKNOWN>")`. And then `ResourceDetailsLabelProvider` will make it "(Not Provided)"

adietish: the problem here is that there actually is something that^s defined, but it's unknown to the openshift-restclient-java, it cannot parse it, thus returning null. That's something different than "not provided". That's rather "unknown".

adietish:
@bdshadow I switched to using "(Unknown)" and added a test for this case.

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/ui/wizard/newapp/ResourceDetailsContentProviderTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+	private ResourceDetailsContentProvider contentProvider;
+
+	@Before
+	public void before() {
+		this.contentProvider = new ResourceDetailsContentProvider();
+	}
+
+	@Test
+	public void shouldReturnUnknownResourcePropertyGivenBuildConfigBuildßStrategyIsNull() {
```
bdshadow: `...BuildßStrategy...`
@adietish scharfes S hid in the method name. Could you, please, remove it and we can merge it

adietish: ohhh, ok :)

PR #1465, opened 2017-04-11
JBIDE-24250 - OpenShift Explorer sometimes don't show the objects in Properties view

jeffmaury:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@jeffmaury there is nothing Devstudio specific in the issue that this PR is fixing. I thus moved the jira from devstudio to jboss tools: https://issues.jboss.org/browse/JBIDE-24250.
Can you please update the commit-message to reference JBIDE-24250?

jeffmaury:
@adietish 2 remarks:

- can't pass null to load as it may cause NPE
- don't need to check the state as it is handled by load

PR #1479, opened 2017-04-26
jbide-24234: code cleanup + fix sonar blocker and critical issues

bdshadow:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
@@ -141,7 +141,7 @@ public class OpenShiftDebugUtils {
 						throw new OpenShiftCoreException(e);
 					}
 				}
-			};
+			}
 		});
 		
 		ConnectionsRegistrySingleton.getInstance().addListener(rcListenerJob.getConnectionsRegistryListener());
@@ -298,21 +298,20 @@ public class OpenShiftDebugUtils {
 		}
 	}
 	
-	public boolean isRunning(ILaunchConfiguration launchConfiguration, int localDebugPort) {
+	public boolean isRunning(ILaunchConfiguration launchConfiguration) {
 		boolean isRunning = getLaunches()
-				.filter(l -> !l.isTerminated() && launchMatches(l, launchConfiguration, localDebugPort))
+				.filter(l -> !l.isTerminated() && launchMatches(l, launchConfiguration))
 				.findFirst().isPresent();
 		return isRunning;
 	}
 	
 	
-	private boolean launchMatches(ILaunch l, ILaunchConfiguration launchConfiguration, int localDebugPort) {
+	private boolean launchMatches(ILaunch l, ILaunchConfiguration launchConfiguration) {
 		return Objects.equals(l.getLaunchConfiguration(), launchConfiguration);
 	}
 
 	public static String getRemoteDebuggerLaunchConfigurationName(IServer server) {
-		String name ="Remote debugger to "+server.getName();
-		return name;
+		return "Remote debugger to " + server.getName();
 	}
 	
 	public void terminateRemoteDebugger(IServer server) throws CoreException {
--- a/plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/connection/ExpressConnection.java
+++ b/plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/connection/ExpressConnection.java
@@ -60,7 +60,7 @@ import com.openshift.internal.client.utils.StreamUtils;
  * @author Xavier Coulon
  * @author Andre Dietisheim
  */
-public class ExpressConnection extends AbstractConnection implements Cloneable {
+public class ExpressConnection extends AbstractConnection {
 
 	private static final String SECURE_STORAGE_PASSWORD = "pass";
 
--- a/plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/connection/ExpressConnection.java
+++ b/plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/connection/ExpressConnection.java
@@ -60,7 +60,7 @@ import com.openshift.internal.client.utils.StreamUtils;
  * @author Xavier Coulon
  * @author Andre Dietisheim
  */
-public class ExpressConnection extends AbstractConnection {
+public class ExpressConnection extends AbstractConnection implements Cloneable {
 
 	private static final String SECURE_STORAGE_PASSWORD = "pass";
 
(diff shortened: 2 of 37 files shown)


bdshadow:
@robstryker @adietish @jeffmaury could you, please, review it and merge. It's not a difficult PR, but such a clean up is a good thing to be merged before the release. + it would help to avoid lots of messages from sonar for every PR, it would make them all green and would allow to see new critical/blocker issues (by sonar opinion) more explicitly.

plugins/org.jboss.tools.openshift.express.ui/src/org/jboss/tools/openshift/express/internal/ui/command/DeleteApplicationHandler.java
```diff
@@ -75,8 +75,7 @@ private IStatus deleteApplications(final List<IApplication> appsToDelete, Shell
 
 	private List<IUser> getUsers(List<IApplication> applications) {
 		List<IUser> users = new ArrayList<>();
-		if (applications == null
-				|| applications.size() == 0) {
+		if (applications.isEmpty()) {
```
adietish: shouldn't we null check the applications parameter?

plugins/org.jboss.tools.openshift.express.ui/src/org/jboss/tools/openshift/express/internal/ui/server/EGitUI.java
```diff
@@ -24,13 +24,10 @@
...
 public class EGitUI implements IEGitUI {
 
-	public EGitUI() {
-	}
-
 	@Override
 	public void publish(IProject project, final String remote, final String applicationName, final Job pushJob) throws CoreException {
 		final Repository repository = EGitUtils.checkedGetRepository(project);
-
+
```
adietish: this looks like an obsolete change, it just inserts a tab?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -298,21 +298,20 @@ public void setupRemoteDebuggerLaunchConfiguration(ILaunchConfigurationWorkingCo
...
 		boolean isRunning = getLaunches()
-				.filter(l -> !l.isTerminated() && launchMatches(l, launchConfiguration, localDebugPort))
+				.filter(l -> !l.isTerminated() && launchMatches(l, launchConfiguration))
 				.findFirst().isPresent();
 		return isRunning;
 	}
 	
 	
-	private boolean launchMatches(ILaunch l, ILaunchConfiguration launchConfiguration, int localDebugPort) {
+	private boolean launchMatches(ILaunch l, ILaunchConfiguration launchConfiguration) {
```
bdshadow: @adietish what do you mean? don't understand "dito", sorry :)

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
```diff
@@ -298,21 +298,20 @@ public void setupRemoteDebuggerLaunchConfiguration(ILaunchConfigurationWorkingCo
 		}
 	}
 	
-	public boolean isRunning(ILaunchConfiguration launchConfiguration, int localDebugPort) {
+	public boolean isRunning(ILaunchConfiguration launchConfiguration) {
 		boolean isRunning = getLaunches()
-				.filter(l -> !l.isTerminated() && launchMatches(l, launchConfiguration, localDebugPort))
+				.filter(l -> !l.isTerminated() && launchMatches(l, launchConfiguration))
```
bdshadow: i removed the last parameter from `launchMatches` method, because it isn't used

plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/server/ExpressServerPublishMethod.java
```diff
@@ -221,9 +221,9 @@ private IProject getModuleProject(IModule[] modules) {
 	protected PushOperationResult publish(final IProject project, final IServer server, final IProgressMonitor monitor) 
 			throws CoreException {
 		IProgressMonitor subMonitor = new SubProgressMonitor(monitor, 200);
-		try {
+		try (Repository repo = EGitUtils.getRepository(project)) {
```
bdshadow: +1
will fix it

plugins/org.jboss.tools.openshift.express.ui/src/org/jboss/tools/openshift/express/internal/ui/command/DeleteApplicationHandler.java
bdshadow: sonar reports this as an if-statement that always returns `true`. Indeed, it is a private method and  the check of `null` already exist in `execute` method.
What else must be mentioned is that sonar has a rule that an empty list must be returned from the method and not null. That's why it always expects that list is not null.

plugins/org.jboss.tools.openshift.express.ui/src/org/jboss/tools/openshift/express/internal/ui/server/EGitUI.java
bdshadow: +1. Strange, might have added it unintentionally

plugins/org.jboss.tools.openshift.express.core/src/org/jboss/tools/openshift/express/internal/core/connection/ExpressConnection.java
```diff
@@ -60,7 +60,7 @@
  * @author Xavier Coulon
  * @author Andre Dietisheim
  */
-public class ExpressConnection extends AbstractConnection {
+public class ExpressConnection extends AbstractConnection implements Cloneable {
```
bdshadow: agree. Can't remember, why did i do it. Will work on it

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
bdshadow: re-read your comment once again. The reason for not just ignoring it is that Sonar has the rule:  "Unused method parameters should be removed". And i agree with it, because imagine the case when you want to use some method, that has this unused parameter. And it's not easy for you retrieve it at the moment at your place of code to pass it into this method. Of course, you can read the method body and understand that it's not used and can be deleted, but it's unnecessary work during coding + you'd probably write a new private method for your needs. So i think it's better to refactor it while refactoring in this PR

adietish: @bdshadow sure while I fully agree with the general rule as much as I would also subscribe to "remove dead code" I am reluctant in case of our OpenShift v2 code because "never touch a running system" and "we won't invest anything here" seems equally important to me. If you are completely sure that we are not breaking anything (we should maybe have QE confirming) then we can have a go for this change.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugUtils.java
adietish: "same here", the same arguments would apply to this change.

adietish:
@rhopp do we have integration tests that assert the above changes?

PR #1480, opened 2017-04-26
JBIDE-24306 - remove outdated code that disables host in server editor

robstryker:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
@@ -12,7 +12,6 @@ package org.jboss.tools.openshift.internal.ui.server;
 
 import java.io.File;
 import java.io.UnsupportedEncodingException;
-import java.lang.reflect.Field;
 import java.net.MalformedURLException;
 
 import org.eclipse.core.databinding.Binding;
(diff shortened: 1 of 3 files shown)


apupier:
I created https://github.com/jbosstools/jbosstools-openshift/pull/1482
Might be good to use the 2 approcahes

robstryker:
testPR

robstryker:
testPR

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
```diff
@@ -616,7 +594,7 @@ private Job loadingProjectsJob(OpenShiftServerEditorModel model, IServerWorkingC
 	private void loadResources(final Composite container, OpenShiftServerEditorModel model) {
 		IServerWorkingCopy server = input.getServer();
 
-		final Connection connection = OpenShiftServerUtils.getConnection(server);
+		OpenShiftServerUtils.getConnection(server);
```
adietish: why is this operation even required if the result isnt used? If there's a reason for it we should add least comment it.

apupier: I don't know if this method has a side effect or not. So I kept it, I cleaned removed the unused parameter but didn't had time to check if the getConnection call has side effect (creating the connection if not yet available for instance)

robstryker: I can't see any side-effect at all. Seems to be a 100% static getter. Doesn't seem necessary at all. Removing it.

robstryker:
pushed to master and maintenance

PR #1481, opened 2017-04-26
[JBIDE-24312] fixed parsing "issuer" for SSL certificates

adietish:
# WORK IN PROGRESS, DONT MERGE YET
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@bdshadow @jeffmaury please test/review this. Latest push added issued to the items that we show: in the preferences (table & details view) but also in the ssl dialog.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/preferences/SSLCertificatesPreference.java
```diff
@@ -80,128 +111,114 @@ public void setAllowedByCertificate(X509Certificate certificate, boolean result)
...
+		}
+		return savedItems;
+	}
+
+	private List<HostCertificate> loadSavedCertificates() {
+		List<HostCertificate> savedItems = new ArrayList<>();
+		String value = getPreferenceStore().getString(ALLOWED_CERTIFICATES_NEW);
+		if (!StringUtils.isBlank(value)) {
+			String[] 	tokens = value.split(SEPARATOR);
+			for (int i = 0; i + 4 <= tokens.length;) {
```
jeffmaury: What the reason for the +4 increment as seems we store two items per certificate ?

jeffmaury: Plus I noticed that there is a missing separator between the end of the certificate and the start of the following. So I stored 2 certificates but when I restarted the list was empty because we got only 3 tokens thus the while is not executed

adietish:
@rhopp please test this critical fix.

adietish:
@jeffmaury plz test again. Updated the code.

PR #1482, opened 2017-04-27
Jbide 24306 support image without routef for jmx connection in server adapter

apupier:
- case of update might need further work to be covered
- no test provided


apupier:
testPR

robstryker:
Let's isolate work on this issue over at https://github.com/jbosstools/jbosstools-openshift/pull/1480

So we don't end up with 2 competing streams left and right.

PR #1500, opened 2017-05-10
[JBIDE-24295] using openshift-restclient-java 5.6.0.Final

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@bdshadow @jeffmaury please review, we need this before we freeze

plugins/org.jboss.tools.openshift.client/pom.xml
```diff
@@ -28,7 +28,7 @@
 		<commons-codec.version>1.6</commons-codec.version>
 		<commons-lang.version>2.6</commons-lang.version>
 		<commons-io.version>2.1</commons-io.version>
-		<enforcer.skip>true</enforcer.skip>
+		<!-- <enforcer.skip>true</enforcer.skip> -->
```
bdshadow: @adietish why did you comment it?

adietish: had to comment it when we started the current sprint using a SNAPSHOT client library. The build would have failed constantly on the enforcer rule. Now that the switch to the final version of the client we can get the enforcer back. But I ack that the current approach is kinda useless.

PR #1511, opened 2017-05-25
JBIDE-24452 ITests: Support local templates in ServiceRequirement

jrichter1:
Service requirement can now load template from a specified path or URL.
 - template=/path/to/template.json for local file
 - template=URL/to/template.json for remote files

@rhopp care to take a look?


adietish:
other than my nitpicking PR seems fine (only read the code).

PR #1524, opened 2017-07-06
JBIDE-19604 - improvement to new application wizard wrt secure storage

robstryker:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/common/ui/wizard/AbstractApplicationWorkbenchWizard.java
+++ b/plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/common/ui/wizard/AbstractApplicationWorkbenchWizard.java
@@ -65,20 +65,18 @@ public abstract class AbstractApplicationWorkbenchWizard extends Wizard implemen
 		@Override
 		public IWizardPage getNextPage() {
 			IWizardPage page = null;
-			if (isConnected()) {
-				if (wizard != null) {
-					page = wizard.getStartingPage();
-				}
-			} else if (DataBindingUtils.isValid(getDatabindingContext())
-					&& connect()) {
-				IConnection connection = getConnection();
-				this.wizard = getWizard(connection);
-				if (wizard != null) {
-					wizard.setConnection(connection);
-					page = wizard.getStartingPage();
+			if (isConnected() && wizard != null) {
+				page = wizard.getStartingPage();
+			} else if (DataBindingUtils.isValid(getDatabindingContext())) {
+				if (connect()) {
+					IConnection connection = getConnection();
+					this.wizard = getWizard(connection);
+					if (wizard != null) {
+						wizard.setConnection(connection);
+						page = wizard.getStartingPage();
+					}
 				}
 			}
-			
 			return page;
 		}
 


plugins/org.jboss.tools.openshift.common.ui/src/org/jboss/tools/openshift/common/ui/wizard/AbstractApplicationWorkbenchWizard.java
```diff
@@ -65,20 +65,18 @@ public boolean canFlipToNextPage() {
...
-			} else if (DataBindingUtils.isValid(getDatabindingContext())
-					&& connect()) {
-				IConnection connection = getConnection();
-				this.wizard = getWizard(connection);
-				if (wizard != null) {
-					wizard.setConnection(connection);
-					page = wizard.getStartingPage();
+			if (isConnected() && wizard != null) {
+				page = wizard.getStartingPage();
+			} else if (DataBindingUtils.isValid(getDatabindingContext())) {
```
jeffmaury: Not a big issue but why did you separate the 2 tests here ?

robstryker: I somehow didn't like the connect() (which does some action) in the 2nd half of an else statement, where I would typically be expecting checks on some boolean or some value comparison.  Basically, I thought "connect()" was hidden in the 2nd half and deserved a more prominent location since it actually takes some action.

adietish:
testPR

adietish:
@rhopp do we have bot-tests for this?

rhopp:
@adietish I've created JBIDE-24662 for this.

bdshadow:
@robstryker really can't understand your changes. There are no changes in the logic of the method, just division and union of the `if` and `else` conditions

robstryker:
Careful observation will show there is a difference ;) 

Imagine a situation where isConnected() returns true, but wizard == null.  In the previous code, it will enter the fist if statement, but not the nested if statement. Thus, page will remain unset. It will not enter the else if (DataBindingUtils.isValid(getDatabindingContext())  because it already entered the prior isConnected() if. 

Prior code:

	if (isConnected()) {  // code enters this if statement
		if (wizard != null) {  // code skips over this if statement
			page = wizard.getStartingPage();  // page remains null, since this if is not entered
		}
	} else if (DataBindingUtils.isValid(getDatabindingContext()) // code does not come here, as this is else
			&& connect()) {
		// do more stuff
	}

Under the new workflow, it is:


	if (isConnected() && wizard != null) { // code does not enter here, bc wizard = null
		page = wizard.getStartingPage();
	} else if (DataBindingUtils.isValid(getDatabindingContext())) { // code enters this else-if test
		// do more stuff, make more checks, etc. 
	}

It may appear it's just a re-org, but it actually makes sure that if the wizard is connected but wizard is still null, that we move forward. Furthermore, lines later, we make sure the wizard gets a value whenever possible, whereas before several occasions led the wizard to remaining null. 

So yeah, the patch does stuff, i promise ;)

robstryker:
I've re-tested the code, reviewed it, and in conjunction with andre's +1, I'm going to push it.  Thanks for the reviews all.

adietish:
@rhopp thanks for jira'ing

PR #1531, opened 2017-07-26
JBIDE-24768 - ensure changes to extended properties take effect after…

robstryker:
… finish

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


mmalina:
I tried this. I could reproduce the bug before applying the patch. Then I applied the patch and the bug disappeared. So +1.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/AdvancedConnectionEditor.java
```diff
@@ -343,9 +349,17 @@ public boolean isViewFor(Object object) {
 		return object instanceof ConnectionFactory;
 	}
 
-	private Connection getConnection() {
+	Map<String, Object> map = null;
```
adietish: @robstryker and then there's a page model already, what's the reasoning to add another model within the UI class where so far special care was applied separate model(-logic) from UI? Why not having pageModel#getExtendedProperties?

PR #1533, opened 2017-07-31
[JBIDE-24775] dont classcast when creating new v3 connection

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/listeners/ConfigureDependentFrameworksListener.java
@@ -38,7 +38,7 @@ public class ConfigureDependentFrameworksListener extends UnitedServerListener {
 	}
 	
 	private void scheduleConfigureFrameworksJob(final ServerEvent event) {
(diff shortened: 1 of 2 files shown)


adietish:
@robstryker please review

adietish:
the odd thing though is that extended properties are *NOT* saved though. See https://issues.jboss.org/browse/JBIDE-24768 which wants to *NOT* save them when cancelling the wizard. Hitting "Finish" though should save them, which it doesnt do - and i dont think that my patch is in any way responsible for this.

adietish:
@robstryker please explain in what your solution at https://bluejeans.com/6504809063/ is superior. To me it just more complex without added benefits (1. instance var map storing extended vars that needs to be synced with the orig source 2. distinction btw. non connection objects, v2 and v3 connections)

adietish:
in bluejeans we agree on
* using an instance var is need to temporarly store settings so that settings only get set to the connection once you hit "Finish", not saved if you hit "Cancel"
* map is a bad name for an instance var, it should be something like extendedProperties
* the extended properties map should be grouped with the other instance vars

adietish:
@robstryker merged your changes and modified according to what we discussed in BJ. Please review.

robstryker:
+1

adietish:
maven build is passing locally (and remotely). cdk-/openshift-integration-tests are failing across PRs. @rhopp can you please look into these failures?

PR #1534, opened 2017-08-01
[JBIDE-24501] dont NPE if pushing img that could not be looked up

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@jeffmaury please review

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/DockerConfigMetaData.java
```diff
@@ -61,16 +61,25 @@ public DockerConfigMetaData(IDockerImageInfo info) {
 	
 	@Override
 	public Set<String> exposedPorts() {
+		if (info == null ) {
+			return null;
```
adietish: not a good idea because we want to signal via null (or exception) that no metadata is present.

PR #1538, opened 2017-08-08
ITests: Extend wait for project creation & change label of builder image.

rhopp:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


rhopp:
@jkopriva review pls

test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/utils/OpenShiftLabel.java
```diff
@@ -299,7 +299,7 @@
 	public static class Others {
```
jcantrill: Not related to this PR, but this seems like a strange use of a class.  Typical java convention is to stuff these into an interface, which means you can then add them to any class, remove the 'public' modifier

jkopriva:
looks good to me, @bdshadow @adietish @jeffmaury please merge :)

PR #1545, opened 2017-08-17
jbide-24782: NullPointerException in ResourceUtils.getBuildConfigFor

bdshadow:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPage.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPage.java
@@ -57,6 +57,7 @@ import org.eclipse.jface.databinding.viewers.ObservableListTreeContentProvider;
 import org.eclipse.jface.databinding.viewers.ViewerProperties;
 import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.IDialogConstants;
+import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.jface.layout.GridLayoutFactory;
 import org.eclipse.jface.resource.JFaceResources;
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPage.java
```diff
@@ -497,9 +495,9 @@ private SelectionListener onImportProject(ServerSettingsWizardPageModel model, f
             @Override
             public void widgetSelected(SelectionEvent e) {
                 Map<com.openshift.restclient.model.IProject, Collection<IBuildConfig>> projectsAndBuildConfigs = new HashMap<>();
-                List<IBuildConfig> buildConfigs = new ArrayList<>();
-                ResourceUtils.getBuildConfigFor(model.getResource(), buildConfigs);
-                projectsAndBuildConfigs.put(model.getResource().getProject(), buildConfigs);
+                if (model.getResource() != null) {
+                	projectsAndBuildConfigs.put(model.getResource().getProject(), Collections.emptyList());
+                }
```
adietish: this looks as if we wont store no build configs at all (if the model holds a resource, we store an empt list of buildconfigs for the project of that resource). We also dont get the build configs any more. Without testing it, this doesnt look correct. Is it?

bdshadow: the list of build configs is retrieved in `ImportApplicationWizard`
Look into lines 501-502 , before my changes - they do nothing. `ResourceUtils.getBuildConfigFor` was called and it is getter - so it returns some value. But why isn't it assigned to any variable and used after? This code is useless. Look inside `ResourceUtils.getBuildConfigFor` and examples of its usage in other places and you'll get the idea.
So my fix just removes the NPE indeed and nothing else. As you said, test it and you'll see that it works.

bdshadow:
found that it causes some Exceptions later too. Need to fix it in  one more place too

bdshadow:
@adietish can be tested now

adietish:
@bdshadow other than the missing DCO things look fine to me. I'll merge it as soon as there's your required "signed-off by".

bdshadow:
I've added "signed-off by" to the first commit, but forgot for the second one. Seems all of them need it.

adietish:
@bdshadow yes absolutely. Given a PR that consists of several commits gets merged, I am not sure if the check is really helpful. Depends on the merge that would preserve the DCO or not (which I dont know)

adietish:
manually merged both commits, made sure there's a signed-off and pushed to origin master in commit 0e5add768f1c7c5b7bdf1e7945fc96c118a4ecf0

PR #1550, opened 2017-08-23
jbide-24781: Create OpenShift server adapters automatically when creating application

bdshadow:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
@@ -125,39 +125,16 @@ public enum OCBinary {
 	}
 	
 	/**
-	 * Checks if the oc binary is compatible for rsync publishing.
-	 * 
-	 * @param monitor the progress monitor
-	 * 
-	 * @return true if the oc binary is compatible
-	 * @see https://issues.jboss.org/browse/JBIDE-21307
-	 * @see https://github.com/openshift/origin/issues/6109
-	 */
-	public boolean isCompatibleForPublishing(IProgressMonitor monitor) {
-	    return isCompatibleForPublishing(null, monitor);
-	}
-	
-	/**
-	 * Checks if an arbitrary oc binary is compatible for rsync publishing
-	 * @param location
-	 * @param monitor
-	 * @return
-	 */
-	public boolean isCompatibleForPublishing(IConnection connection, IProgressMonitor monitor) {
-	    return new OCBinaryVersionValidator(getLocation(connection)).isCompatibleForPublishing(monitor);
-	}
-	
-	/**
 	 * Compute the error message for the OCBinary state and path.
 	 * 
 	 * @param valid if the oc binary is valid or not
 	 * @param location the location of the oc binary
 	 * @return the error message (may be null)
 	 */
-	public IStatus getOCBinaryStatus(IConnection connection, IProgressMonitor monitor) {
+	public IStatus getStatus(IConnection connection, IProgressMonitor monitor) {
 		String location = getLocation(connection);
 	    IStatus status = Status.OK_STATUS;
-	    if (!isCompatibleForPublishing(connection, monitor)) {
+	    if (new OCBinaryVersionValidator(getLocation(connection)).isCompatibleForPublishing(monitor)) {
 	        if (location == null) {
 	            status = OpenShiftCoreActivator.statusFactory().errorStatus(OpenShiftCoreMessages.NoOCBinaryLocationErrorMessage);
 	        } else if (!new File(location).exists()) {
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
@@ -10,13 +10,20 @@
  ******************************************************************************/ 
 package org.jboss.tools.openshift.internal.core.preferences;
 
+import java.io.File;
+
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.osgi.util.NLS;
 import org.jboss.tools.openshift.common.core.connection.IConnection;
 import org.jboss.tools.openshift.core.ICommonAttributes;
+import org.jboss.tools.openshift.core.OpenShiftCoreMessages;
 import org.jboss.tools.openshift.core.connection.IOpenShiftConnection;
 import org.jboss.tools.openshift.core.preferences.OpenShiftCorePreferences;
 import org.jboss.tools.openshift.internal.common.core.util.CommandLocationBinary;
+import org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator;
 
 public enum OCBinary {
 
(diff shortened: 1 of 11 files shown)


bdshadow:
@psrna @mmalina i think it also should be tested by QE, as it is a new feature

mmalina:
@jkopriva can you take a look?

jkopriva:
@bdshadow looks good to me, +1

adietish:
@bdshadow v2 had a dedicated wizard page upfront where the user could check/unchecked based on his wishes to create a server adapter. This is imho preferable to coming up with a message dialog  in the process (wizard should ideally provide all required settings and process without further questions once you hit "Finish").

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
```diff
@@ -139,4 +146,26 @@ public boolean isCompatibleForPublishing(IProgressMonitor monitor) {
...
 	}
+	
+	/**
+	 * Compute the error message for the OCBinary state and path.
+	 * 
+	 * @param valid if the oc binary is valid or not
+	 * @param location the location of the oc binary
+	 * @return the error message (may be null)
+	 */
+	public IStatus getOCBinaryStatus(IConnection connection, IProgressMonitor monitor) {
```
adietish: I see #getOCBInaryStatus being moved from the server page to OCBinary. Which seems like a great change  to me (in terms of encapsulation and so that it can be reused). 
1) Can now OCBinary#getLocation get private?
2) Shouldn't we now replace all #isCompatibleForPublishing by #getOCBinaryStatus?
3) Can't we rename #getOCBinaryStatus to #isValid ? (the name #getOCBinaryStatus is obviously repeating an obvious point: it's about the status of the binary. But this is obvious since we're in the OCBinary class anyhow. Then #getStatus seems not to nail the purpose since it only checks binary location and version, not some binary state.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
```diff
@@ -139,4 +146,26 @@ public boolean isCompatibleForPublishing(IProgressMonitor monitor) {
...
+		String location = getLocation(connection);
+	    IStatus status = Status.OK_STATUS;
+	    if (!isCompatibleForPublishing(connection, monitor)) {
```
adietish: why check the version 1st and the location afterwards? To me inversing this sequence makes more sense (1. Is a path provided? 2. does the path exist? 3. Does the version match the requirements)?

bdshadow:
@adietish didn't know about it, because didn't work with v2 at all. I'm ok to do it, though it's a lot to change in this pr to achieve it. So i'd be grateful, if you try it as it is now and tell me if it is uncomfortable and better to integrate into the wizard.
@jeffmaury i'd appreciate your comment too :)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizard.java
```diff
@@ -220,14 +232,37 @@ public void done(IJobChangeEvent event) {
...
+								@Override
+								public void done(IJobChangeEvent event) {
+									IService service = getService(resources);
+									List<org.eclipse.core.resources.IProject> importedProjects = wizard.getImportJob().getImportedProjects();
+									if (service != null && importedProjects.size() == 1) {
+										Display.getDefault().asyncExec(new Runnable() {
+
+								            @Override
+								            public void run() {
+								            	if (MessageDialog.openQuestion(getShell(), "Create server adapter", NLS.bind(
```
adietish: I dont think that we should ask the user. I think that we should provide a wizard page (like we do in v2) that holds a checkbox ([x] Create Server adapter for easy publishing).
Possibly we even should have the server adapter wizard page included in the new application wizard

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/newapp/NewApplicationWizard.java
```diff
@@ -252,7 +287,43 @@ public void run() {
...
+			}
+			
+			protected IRoute getRoute(Collection<IResource> resources) {
+				IResource route = getResourceOfType(resources, IRoute.class);
+				return route == null ? null : (IRoute)route;
+			}
+			
+			private IResource getResourceOfType(Collection<IResource> resources, Class<? extends IResource> type) {
+				for (IResource resource : resources) {
+					if (type.isInstance(resource)) {
```
adietish: so we take the 1st available service? what if there are more services?
Furthermore we can create server adapters for replication controllers, deployment configs, too

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
bdshadow: 1. `public String getLocation()` is already marked `@Deprecated`. `public String getLocation(IConnection connection)` is used in `AbstractOpenShiftCliHandler` and `OCBinaryOperation`. So i don't think that it's worth making it private now. Though need to look deeper and not as part of this pr.
2.  +1. Actually, all `isCompatibleForPublishing` are not used anymore anywhere. So they can be even removed now.
3. +1

adietish:
@bdshadow the current implementation works for me even though not as userfriendly as it could be. 
IMHO including a wizard page for the server adapter (which also could allow the user to select the service to bind the adapter to - we know what resources the template/builder image will create) is still preferrable. We could postpone this change to a enhancement that we tackle later. Works for you?
btw. we have a similar shortcoming in the import wizard which also asks the user while processing the wizard as it was "finished" by the user. The import, too, should have it's wizard page and not ask anything unless something unexpected is occurring.

bdshadow:
@adietish it is actually the main reason why i did it like this. Because if you don't import a project then it doesn't make sense to create a server adapter, because we need project for it. And this import isn't included into the wizard

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
bdshadow: makes sense, but it was just copied from https://github.com/jbosstools/jbosstools-openshift/pull/1550/files/4343beba8643b34ba04108828c17db2ebcf0ca5f#diff-5627fe7968b85348bfd960f2e0beae9aL206

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/preferences/OCBinary.java
bdshadow: @adietish read once again 3rd point. Maybe i didn't understand you completely,  but i think `getStatus` is the best option. `isValid` doesn't match well, because smth starting from "is" must return binary: true\false

adietish:
@bdshadow i agree that of course, the server adapter makes no sense if no project is imported.   Nevertheless the project could be imported by other means (ex. cloned and imported the source that is used by the template). Furthermore the wizard page for the server settings to hid the project part and assume the future import is the project that it should use.
I fully agree, like I said above, that we can postpone this enhancement to later in an additional jira.
This actually reminds me that importing a project could also have the option to create a server adapter.

PR #1552, opened 2017-08-24
JBIDE-24881 Change command to get PATH on Mac

mmalina:
The command we were using is generating some
redundant output that we don't want.

Signed-off-by: Martin Malina <mmalina@redhat.com>


mmalina:
I tried this both on my machine - macOS 10.12 Sierra - and on 10.11 El Capitan and it works. And this command is only used for mac so we should be fine.

mmalina:
OK, now that I'm reading about the interactive mode, there may be a catch. 

When the shell is running interactively, it changes its behavior in several ways.
Startup files are read and executed as described in Bash Startup Files.

So I tested this a little. When eclipse is started from Finder: 

If I only use 'bash -c "echo $PATH"' there will be a very short path: /usr/bin:/bin:/usr/sbin:/sbin:/Users/rasp/jbossqa/cdk/cdk-3.1.0
This suggests that .bash_profile is not invoked.

If I use 'bash -l -c "echo $PATH"' there will be a longer path, even something that I added to my .bash_profile just now:
bla:/Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/rasp/jbossqa/cdk/cdk-3.1.0

So this suggests that -l is enough to get .bash_profile invoked. And that was our goal from the start I think - because when Eclipse is launched from Finder, the PATH is very minimal.

PR #1553, opened 2017-08-28
JBIDE-24877 - cdk3 detector not careful enough

robstryker:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
AbstractCDKRuntimeDetector#validate loops over the entries as files:
```
	protected boolean validate(File root) {
		String[] required = getRequiredChildren();
		for (int i = 0; i < required.length; i++) {
			if (!(new File(root, required[i]).exists())) {
				return false;
			}
		}
		return true;
	}
```
safe enough it seems.
+1

PR #1556, opened 2017-08-30
JBIDE-24222 - fix xhyve issue for cdk3

robstryker:

This patch may fix the following JIRAs and should be tested against them:  JBIDE-24676, JBIDE-24188, JBIDE-24293, JBIDE-24222

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


mmalina:
This fixes the main issue for me - JBIDE-24222 - that CDK with xhyve didn't start in Eclipse properly.
The other issues which were about ugly output in the console, that's not really solved because that will require a special parameter in minishift to disable to progress indicator. I don't know what the effect on Windows will be, but I suggest committing this patch and then we'll check later.

mmalina:
So +1 to merge.

gbraad:
@mmalina the environment variable is not needed anymore.

PR #1559, opened 2017-09-01
jbide-24797 - apply default Eclipse code formatter

bdshadow:
Signed-off-by: Dmitrii Bocharov <dbocharo@redhat.com>

Here is the default eclipse formatting applied by the [formatter-maven-plugin](http://code.revelc.net/formatter-maven-plugin/formatter-maven-plugin/).

The main changes are here in pom.xml:https://github.com/jbosstools/jbosstools-openshift/pull/1559/files#diff-600376dffeb79835ede4a0b285078036R41

You can compare it with a slightly different formatting profile: https://github.com/jbosstools/jbosstools-openshift/pull/1560

PR #1560, opened 2017-09-01
jbide-24797_2: apply changed Eclipse formatter

bdshadow:

Here is the formatting slightly different from the standard Eclipse formatting profile

-  line length set to 140
- line wrapping is more intuitive (in my opinion)

The main changes are here in pom.xml: https://github.com/jbosstools/jbosstools-openshift/pull/1560/files#diff-600376dffeb79835ede4a0b285078036R41

You can compare it with default Eclipse formatting profile: https://github.com/jbosstools/jbosstools-openshift/pull/1559


adietish:
@bdshadow unfortunately we'd have to rebase or re-create this PR since changes got in after it was created

bdshadow:
@adietish rebased + added some more small changes to the formatter profile.

@jeffmaury also proposed a great idea - put the profile file somewhere into a common place, e.g. https://github.com/jbosstools/jbosstools-build - and use it during the build.
So i'lll make a PR soon for the jbosstools-build and then a new pr for this repo with corresponding changes in pom.xlm

adietish:
@bdshadow +1 for putting it into a common place. That's a suggestion that I spotted in various places.

bdshadow:
@adietish sorry, didn't get it under lots of discussions. So i'll configure it to be downloaded from some url and then applied

However, firstly i'd vote for merging this one, so that it means that we agreed about [eclipse-formatter.xml](https://github.com/jbosstools/jbosstools-openshift/pull/1560/files#diff-920569344b0de22da2ff1769cf8a9d18) file. Then i'll transfer it to some common place.

adietish:
@bdshadow build plugin, which is git reachable and thus accessible via URL?

jeffmaury:
I would prefer devdoc

adietish:
@bdshadow just to make sure that I didnt miss anything: the formatter that you're using is the default that's built in, correct?

bdshadow:
@adietish no. Please, see the first main comment for this PR

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/util/UITestUtils.java
```diff
@@ -18,48 +18,47 @@
...
-	public static void waitForState(IConnectionWrapper connectionWrapper, LoadingState state) throws InterruptedException, TimeoutException {
-		int count= 0;
-		while (connectionWrapper.getState() != state) {
-			Thread.sleep(100);
-			if (count++ > 1000) {
-				throw new TimeoutException();
-			};
-		}
-	}
+    public static void waitForState(IConnectionWrapper connectionWrapper, LoadingState state)
```
adietish: indenting tabs being replaced by spaces?

adietish:
@bdshadow I see here all identation tabs being replaced by spaces. See here as example: https://github.com/jbosstools/jbosstools-openshift/pull/1560/files#diff-8fb9d32f39e86e061fa1f656072a8ae3R52

adietish:
@bdshadow furthermore the reformatting seems to have destroyed things. If I open up the new application wizard I see having flaws, see here:
image

bdshadow:
@adietish yes, i've replaced tabs with spaces here. In my opinion, it's a more natural way (in another words, tab is a symbol that displays nothing, but we have already one for that - space, so why to have one more?). But i don't insist on it.
Furthermore, what i mainly changed in this pr, comparing to the default Eclipse:

1. Increased max comment line width from 80 to 100
2. Removed space insertion after the cast: compare [from this pr](https://github.com/jbosstools/jbosstools-openshift/pull/1560/files#diff-9f7a3ba30a993fe8c82cec4c238888b5R102) vs [default eclipse](https://github.com/jbosstools/jbosstools-openshift/pull/1559/files#diff-9f7a3ba30a993fe8c82cec4c238888b5R102).
3. Increased max line length from 120 to 140.
4. Some line wrapping politics for loops, conditionals and annotation parameters in case of excess of maximum line width.

As i said multiple times, we can just apply the other PR to make everything like in a default eclipse formatting. Just look through the other PR. Do you like it?

@adietish  I'm also going to investigate today why it has broken some things. It's really strange

adietish:
@bdshadow i dont think that there's arguing about formatting as long as readability is equivalent really. It's a matter for taste, but then to me the argument about space vs tab is moot to me. Why would you use redundant 4 spaces for a single tab? But as said, there's no arguing since it's a matter of taste as long as the readability of the code is equivalent. The only point I see here is that the majority of the code was using tabs, so I dont think that switching to spaces makes a lot of sense given that the changing commit gets much bigger with no added value.

bdshadow:
@adietish yes, i agree that it's moot. So lets leave it as it was - tabs, i agree.

What do you think about the space in casts? See the 2nd item in my previous comment. I really don't like the space after the cast. But if you're ok with it, then we can take a default eclipse profile from a neighbor PR.

bdshadow:
@adietish i can't imagine how the reformatting could cause such changes in the UI, that you screenshotted.

Here is the code https://github.com/jbosstools/jbosstools-openshift/pull/1560/files#diff-27efc2a3b54a1d6b3ddc18bcbf9b1f60 responsible for the projects labels and combo boxes on the top of the wizard

adietish:
@bdshadow if the space after a cast is what prevents us from using a default eclipse formatter I'm perfectly fine dropping it. I'm trying to not be picky, my primary concern really is readability and consistency. Can you reformat the things so that I can check the cast you're talking of?
Can you also point me to the formatter that you're using so that I can try it on a file=
@jeffmaury @robstryker can you please also chime in your comments and agree/disagree? We're all be impacted by this change so it's particularly important that we have a consensus here.

bdshadow:
@adietish you can find the default eclipse formating in the neighbor pr and the links to the differences in casts formatting in this comment above: https://github.com/jbosstools/jbosstools-openshift/pull/1560#issuecomment-327391530.
About the formatter - look here: https://github.com/jbosstools/jbosstools-openshift/pull/1560/files#diff-600376dffeb79835ede4a0b285078036 - change your pom.xml like this. The link to the docs of this plugin a posted in the comments to the jira

adietish:
@bdshadow I prefer the cast with the space, but I dont mind in the end really. I am just wondering if we really want to differ from the default (which requires us to setup the formatter whenever we use a fresh workspace) for those 2 differences. Either way, works for me.
@jeffmaury @robstryker thoughts? Please give feedback since you'll be impacted, too.

bdshadow:
@adietish yes, i thought we've already agreed about it.

The main question for me now is why do we have UI changed? I mean your screenshots above. [Here in the comment above](https://github.com/jbosstools/jbosstools-openshift/pull/1560#issuecomment-327450936) i pointed to this wizard source code, and can't imaging how these changes can lead to such things

jeffmaury:
I'm ok with the settings. However I have two questions:

- can we get samples for the various settings (if, loop, conditional)
- will the settings be automatically applied in Eclipse thanks to M2E ?

bdshadow:
@jeffmaury 
1) you can find samples in the PRs
2) this one we've discussed already too. The plugin has `format` and `validate` goals. By default `validate` is set and will run for the PRs. 

So in case your build is failing you either need to do `Ctrl-Shift-F` for changed files or run `mvn format` and it'll will everything for you.

jeffmaury:
@bdshadow Just to be clear:

- if I have a stock Eclipse install (so with the default Eclipse rules) and if I import a jbosstools-openshift then if I reformat one of jbosstools-openshift files, will they be reformatted using our profile or the default Eclipse profile

bdshadow:
@jeffmaury in this case default eclipse profile (or the one that is configured in your eclipse). Maven plugin has no power to go into Eclipse Preferences and change the profile there

adietish:
@jeffmaury @bdshadow your question about the result of formatting in a stock Eclipse is exactly what I was trying to express in my comments: I tend to think that we should only differ from the defaults if we have very convincing reasons (not just personal preferences) to do so. So far I dont think that neither extending the line with 120 -> 140 nor space/no space btw cast and variable are good enough reasons nor the other 2 changes specified at https://github.com/jbosstools/jbosstools-openshift/pull/1560#issuecomment-327391530
Don't you think so? I am open to any dissent, really.

bdshadow:
[as i've already said ](https://github.com/jbosstools/jbosstools-openshift/pull/1560#issuecomment-328529356) i agree to have default eclipse formatting.

So i'm for closing this PR and move discussion to the neighbor one.

bdshadow:
as i can see no objections, i'm closing this PR and moving all the the discussions to [the neighbor one with default eclipse formatter profile](https://github.com/jbosstools/jbosstools-openshift/pull/1559).

PR #1564, opened 2017-09-08
JBIDE-24491 - allow connection wizard to discover registry url

robstryker:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/AdvancedConnectionEditor.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/AdvancedConnectionEditor.java
@@ -11,7 +11,6 @@
 package org.jboss.tools.openshift.internal.ui.wizard.connection;
 
 import java.io.File;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -21,8 +20,6 @@ import org.eclipse.core.databinding.Binding;
 import org.eclipse.core.databinding.DataBindingContext;
 import org.eclipse.core.databinding.ValidationStatusProvider;
 import org.eclipse.core.databinding.beans.BeanProperties;
-import org.eclipse.core.databinding.observable.IObservable;
-import org.eclipse.core.databinding.observable.Realm;
 import org.eclipse.core.databinding.observable.list.IObservableList;
 import org.eclipse.core.databinding.observable.list.WritableList;
 import org.eclipse.core.databinding.observable.value.IObservableValue;
@@ -48,7 +45,6 @@ import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
-import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Display;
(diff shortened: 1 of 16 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/AdvancedConnectionEditor.java
```diff
@@ -276,6 +291,17 @@ protected GridLayoutFactory adjustAdvancedCompositeLayout(GridLayoutFactory grid
 		return composite;
 	}
 
+	private void discoverRegistryPressed() {
+		String ret = RegistryProviderModel.getDefault().getRegistryURL(pageModel.getSelectedConnection());
+		if( ret == null ) {
+			System.out.println("Nothing found");
```
jeffmaury: Could be removed

adietish: @robstryker nice with extension point et all. Can you please add badly needed unit test?

adietish:
@robstryker nice with extension point et all. Can you please add badly needed unit test?

adietish:
@robstryker I tested this, and I'm not sure that it works really.

steps:
#EXEC: edit existing connection to local cdk
#ASSERT: connection wizard shows up
image
#EXEC: click on "Discover"

Result:
Nothing happens. In the log you see "Not found". 
1. How come that nothing is found?
2. If nothing is found we should definitely tell the user

adietish:
testPR

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/AdvancedConnectionEditor.java
```diff
@@ -287,10 +289,14 @@ protected GridLayoutFactory adjustAdvancedCompositeLayout(GridLayoutFactory grid
 		return composite;
 	}
 
-	private void discoverRegistryPressed() {
+	private void discoverRegistryPressed(Shell shell) {
 		String ret = RegistryProviderModel.getDefault().getRegistryURL(pageModel.getSelectedConnection());
 		if( ret != null ) {
 			registryURLObservable.setValue(ret);
+		} else {
```
jeffmaury: Can we report this to the UI in a more UX way: I'm thinking about reporting an Info status to the databinding framework would be more user friendly

adietish:
testPR

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/connection/AdvancedConnectionEditor.java
```diff
@@ -276,6 +289,18 @@ protected GridLayoutFactory adjustAdvancedCompositeLayout(GridLayoutFactory grid
 		return composite;
 	}
 
+	private void discoverRegistryPressed(Shell shell) {
+		String ret = RegistryProviderModel.getDefault().getRegistryURL(pageModel.getSelectedConnection());
+		if( ret != null ) {
+			registryURLObservable.setValue(ret);
+		} else {
+			MessageBox messageBox = new MessageBox(shell, SWT.ICON_ERROR | SWT.OK);
+		    messageBox.setMessage("No registry provider found for the given connection.");
```
adietish: In my case the user error is that I didnt have the adapter started (launched the cdk from cmd-line). So some hint like "You may have to start your CDK server adapter" would be very helpful to the dummy user (like me)

adietish:
Tried it when I did not have the minishift adapter launched: got informed via message dialog that the lookup failed.
Tried it when I had the minishift adapter launched: the value for the registry url was overwritten. 
To me things look ok, we can still file jiras for possible enhancements:
* inform of failed lookup in the title area (not message dialog). I'm not all sure though if that's preferrable. 
* only insert the registry url if there's no value yet, or ask for confirmation before overwriting.

adietish:
build works for me locally without any errors. Jenkins is red though, failing with p2 dependencies. Merging.

PR #1573, opened 2017-09-22
Osfuse 548 clean code during investigation

apupier:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapModulesController.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapModulesController.java
@@ -15,6 +15,7 @@ import java.util.stream.Stream;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.MultiStatus;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.wst.server.core.IModule;
@@ -56,7 +57,7 @@ public class OpenShiftEapModulesController extends JBoss7FSModuleStateVerifier i
 	private MultiStatus syncDown(IProgressMonitor monitor) throws CoreException {
 		final RSync rsync = OpenShiftServerUtils.createRSync(getServer(), monitor);
 		final File localDeploymentDirectory = new File(getDeploymentOptions().getDeploymentsRootFolder(true));
-		final MultiStatus status = new MultiStatus(OpenShiftCoreActivator.PLUGIN_ID, 0, 
+		final MultiStatus status = new MultiStatus(OpenShiftCoreActivator.PLUGIN_ID, IStatus.OK, 
 				NLS.bind("Could not sync all pods to folder {0}.", localDeploymentDirectory.getAbsolutePath()), null);
 		rsync.syncPodsToDirectory(localDeploymentDirectory, status, ServerConsoleModel.getDefault().getConsoleWriter());
 		return status;
@@ -66,7 +67,7 @@ public class OpenShiftEapModulesController extends JBoss7FSModuleStateVerifier i
 		// do rsync local to remote
 		final RSync rsync = OpenShiftServerUtils.createRSync(getServer(), monitor);
 		final File localDeploymentDirectory = new File(getDeploymentOptions().getDeploymentsRootFolder(true));
-		final MultiStatus status = new MultiStatus(OpenShiftCoreActivator.PLUGIN_ID, 0, 
+		final MultiStatus status = new MultiStatus(OpenShiftCoreActivator.PLUGIN_ID, IStatus.OK, 
 				NLS.bind("Could not sync folder {0} to all pods.", localDeploymentDirectory.getAbsolutePath()), null);
 		rsync.syncDirectoryToPods(localDeploymentDirectory, status, ServerConsoleModel.getDefault().getConsoleWriter());
 		return status;
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
@@ -1120,7 +1120,7 @@ public class ResourceUtils {
 	}
 
 	private static String getLastPathSegment(String mangledUri) {
-		int index = mangledUri.lastIndexOf("/");
+		int index = mangledUri.lastIndexOf('/');
 		if (index >= 0) {
 			return mangledUri.substring(index + 1);
 		} else {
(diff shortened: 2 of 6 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/OpenShiftCoreMessages.java
```diff
@@ -20,6 +20,8 @@
     public static String OCBinaryLocationIncompatibleErrorMessage;
     public static String OCBinaryLocationDontExistsErrorMessage;
     public static String NoOCBinaryLocationErrorMessage;
+    public static String couldNoSyncPodsToFolder;
```
adietish: this is long running issue. The only thing I'd agree in is that we should be consistent, either extract all or no messages at all. To me it is very unlikely (and never happened in all the years that I am part of the project) that there will be any other language than english. Thus extracting messages brings no benefit but additional indirection. IMHO the best thing to do is to have things prepared (like using NLS.bind) but **NOT** extracting strings.

adietish: I still believe that we should **NOT** extract strings, but if we do the correct naming would be could**Not**SyncPodsToFolder

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/OpenShiftCoreMessages.java
```diff
@@ -20,6 +20,8 @@
...
+	public static String couldNoSyncFolderToPods;
```
adietish: I still believe that we should **NOT** extract strings, but if we do the correct naming would be could**Not**SyncFolderToPods

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapModulesController.java
```diff
@@ -50,14 +52,14 @@ private void deleteMarkers(String suffix) throws CoreException {
 		final File localDeploymentDirectory = new File(getDeploymentOptions().getDeploymentsRootFolder(true));
 		Stream.of(localDeploymentDirectory.listFiles())
 			.filter(p->p.getName().endsWith(suffix))
-			.forEach(p->p.delete());
+			.forEach(File::delete);
```
adietish: in what is change this superior? To me this is just an alternative way to achieve the same without added benefit.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapModulesController.java
```diff
@@ -50,14 +52,14 @@ private void deleteMarkers(String suffix) throws CoreException {
...
 	}
 	
 	private MultiStatus syncDown(IProgressMonitor monitor) throws CoreException {
 		final RSync rsync = OpenShiftServerUtils.createRSync(getServer(), monitor);
 		final File localDeploymentDirectory = new File(getDeploymentOptions().getDeploymentsRootFolder(true));
-		final MultiStatus status = new MultiStatus(OpenShiftCoreActivator.PLUGIN_ID, 0, 
-				NLS.bind("Could not sync all pods to folder {0}.", localDeploymentDirectory.getAbsolutePath()), null);
+		final MultiStatus status = new MultiStatus(OpenShiftCoreActivator.PLUGIN_ID, IStatus.OK, 
+				NLS.bind(OpenShiftCoreMessages.couldNoSyncPodsToFolder, localDeploymentDirectory.getAbsolutePath()), null);
```
adietish: as said, I think that we should **NOT** extract string, we should strive to be consistent and remove the message classes, moving the strings back into code while using NLS.bind etc.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapModulesController.java
```diff
@@ -66,8 +68,8 @@ private MultiStatus syncUp(IProgressMonitor monitor) throws CoreException {
 		// do rsync local to remote
 		final RSync rsync = OpenShiftServerUtils.createRSync(getServer(), monitor);
 		final File localDeploymentDirectory = new File(getDeploymentOptions().getDeploymentsRootFolder(true));
-		final MultiStatus status = new MultiStatus(OpenShiftCoreActivator.PLUGIN_ID, 0, 
-				NLS.bind("Could not sync folder {0} to all pods.", localDeploymentDirectory.getAbsolutePath()), null);
+		final MultiStatus status = new MultiStatus(OpenShiftCoreActivator.PLUGIN_ID, IStatus.OK, 
+				NLS.bind(OpenShiftCoreMessages.couldNoSyncFolderToPods, localDeploymentDirectory.getAbsolutePath()), null);
```
adietish: see above

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
```diff
@@ -1120,7 +1120,7 @@ private static String stripDotGit(String uri) {
 	}
 
 	private static String getLastPathSegment(String mangledUri) {
```
adietish: I believe that this method should be move to org.jboss.tools.openshift.common.core.utils.URIUtils

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapModulesController.java
apupier: SonarLint is configured to push users to use the method reference writing.
I don't have a strong opinion on that. I find it a little bit more readable but can understand that it can be a matter of taste.
I will revert it

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
apupier: I think that there are several methods that would be best placed in URIUTils such as stripTrailingSlashes, stripDotGit, getProjectNameForURI

I would prefer to keep it for another PR

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/OpenShiftCoreMessages.java
apupier: in Fuse Tooling we were asked to do it continuously to be ready "in case of", if you are going in the opposite way in JBoss Tools, our work is useless :-(

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceUtils.java
adietish: ok, works for me, can we file this move so that we dont forget?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapModulesController.java
adietish: I dont have an opinion either, I was just wondering. Works for me either way.

PR #1577, opened 2017-09-28
[JBIDE-24793] - Implement OSIO login

jeffmaury:

# READY TO BE MERGED NOW

## Testing
You should see a new icon in the toolbar, click and a browser will be launched to login to OpenShift.io (you only need a valid RHDP account not a full OpenShift.io). After login the retrieved token will be displayed and saved. Next time the same token will be returned if no refresh is needed. You can reset the saved token in preferences as well

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


features/org.jboss.tools.openshift.io.test.feature/feature.properties
```diff
@@ -0,0 +1,37 @@
...
+# This file should be translated.
+
+# "featureName" property - name of the feature
+featureName=JBoss OpenShift.io Tools Test
+
+# "providerName" property - name of the company that provides the feature
+providerName=JBoss by Red Hat
+
+# "description" property - description of the feature
+description=Tests for Tooling to work with OpenShift 3.
```
bdshadow: mistype, must be Openshift IO

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
```diff
@@ -0,0 +1,162 @@
...
+
+	public IAccountModel getModel() {
+		if (null == model) {
+			model = new AccountModel();
+		}
+		return model;
+	}
+
+	public AccountStatus getStatus(IAccount account) {
+		if (account.getAccessToken() == null) {
```
bdshadow: possibly, it would be good to add check for `null == account`

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/TokenProvider.java
```diff
@@ -0,0 +1,42 @@
...
+import org.eclipse.core.runtime.Platform;
+import org.jboss.tools.openshift.io.internal.core.OpenShiftIOCoreActivator;
+
+/**
+ * Interface to be implemented for the tokenProvider extension point. This is
+ * just an empty extension to Function<IResource, String> in order to keep
+ * loose coupling for extension point users.
+ */
+public interface TokenProvider extends Function<IResource, String> {
+	default String getToken(IResource resource) {
```
bdshadow: java 8 modern tricks :+1: i like it :)

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/exception/OpenshiftIORefreshException.java
```diff
@@ -0,0 +1,45 @@
...
+	public OpenshiftIORefreshException(IAccount account, int status) {
+		super("Error refreshing token for " + account.getId() + " on cluster " + account.getCluster().getId());
+		this.account = account;
+		this.status = status;
+	}
+
+	public OpenshiftIORefreshException(IAccount account, Throwable t) {
+		super("Error refreshing token for " + account.getId() + " on cluster " + account.getCluster().getId(), t);
+		this.account = account;
+		this.status = (-1);
```
bdshadow: nitpicking - why brackets here?

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/internal/core/OpenShiftIOCoreActivator.java
```diff
@@ -0,0 +1,34 @@
...
+	
+	public static final String PLUGIN_ID = "org.jboss.tools.openshift.io.core";
+	
+	public OpenShiftIOCoreActivator() {
+		super();
+		INSTANCE = this;
+	}
+
+	public static OpenShiftIOCoreActivator getDefault() {
+		return INSTANCE;
```
bdshadow: potential NPE, e.g. on line 31. If constructor hasn't been called yet, the `INSTANCE` is `null`

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/internal/core/model/AccountModel.java
```diff
@@ -0,0 +1,157 @@
...
+		} catch (BackingStoreException e) {
+			OpenShiftIOCoreActivator.logError(e.getLocalizedMessage(), e);
+		}
+		if (clusters.isEmpty()) {
+			addCluster(createCluster("OpenShift.io"));
+		}
+	}
+	
+	void fireEvent(Event event, IAccount account) {
+		listeners.stream().forEach(listener -> {
```
bdshadow: call `forEach` at once, without `stream()`

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/internal/core/model/AccountModel.java
```diff
@@ -0,0 +1,157 @@
...
+				break;
+			case ACCOUNT_REMOVED:
+				listener.accountRemoved(this,  account);
+				break;
+			}
+		});
+	}
+	
+	void fireEvent(Event event, ICluster cluster) {
+		listeners.stream().forEach(listener -> {
```
bdshadow: call `forEach` at once, without `stream()`

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/internal/core/model/AccountModel.java
```diff
@@ -0,0 +1,157 @@
...
+	@Override
+	public void removeCluster(ICluster cluster) {
+		clusters.remove(cluster);
+		fireEvent(Event.CLUSTER_REMOVED, cluster);
+		removed.add(cluster.getId());
+	}
+
+	@Override
+	public void save() {
+		clusters.stream().forEach(ICluster::save);
```
bdshadow: call `forEach` at once, without `stream()`

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/internal/core/model/AccountModel.java
```diff
@@ -0,0 +1,157 @@
...
+		Preferences accountRoot = getAccountsPreferences();
+		ISecurePreferences accountSecureRoot = getSecureAccountsPreferences();
+		removed.stream().forEach(id -> {
```
bdshadow: call `forEach` at once, without `stream()`

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/internal/core/model/Cluster.java
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	@Override
+	public void save() {
+			try {
+				Preferences accountsNode = AccountModel.getAccountsPreferences();
+				Preferences clusterNode = accountsNode.node(getId());
+				ISecurePreferences clusterSecureNode = AccountModel.getSecureAccountsPreferences().node(getId());
+				clusterNode.put(ENDPOINT_URL_KEY, getEndpointURL());
+				removed.stream().forEach(id -> {
```
bdshadow: call `forEach` at once, without `stream()`

plugins/org.jboss.tools.openshift.io.ui/.project
```diff
@@ -0,0 +1,40 @@
...
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.zeroturnaround.eclipse.rebelXmlBuilder</name>
```
bdshadow: i think this one can be removed

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/OpenShiftIOUIActivator.java
```diff
@@ -0,0 +1,34 @@
...
+	
+	public static final String PLUGIN_ID = "org.jboss.tools.openshift.io.ui";
+	
+	public OpenShiftIOUIActivator() {
+		super();
+		INSTANCE = this;
+	}
+
+	public static OpenShiftIOUIActivator getDefault() {
+		return INSTANCE;
```
bdshadow: potential NPE, e.g. on line 31. If constructor hasn't been called yet, the `INSTANCE` is `null`

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
jeffmaury: You're not supposed to call with no account. Does not make sense and I will probably throw the same exception

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/exception/OpenshiftIORefreshException.java
jeffmaury: Habits from an old C dev.....

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/internal/core/OpenShiftIOCoreActivator.java
jeffmaury: Not in the OSGI world

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/OpenShiftIOUIActivator.java
jeffmaury: Not in the OSGI world

bdshadow:
@jeffmaury i have an 
> org.eclipse.swt.SWTException: An exception was raised in JavaScript

full stack trace: https://pastebin.com/whZrmDVj. The same if trying to login directly (not through github) by username/password
ezgif com-video-to-gif

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
```diff
@@ -0,0 +1,162 @@
...
+		long lastRefreshed = account.getLastRefreshedTime();
+		long current = System.currentTimeMillis();
+		if (current > account.getAccessTokenExpiryTime()) {
+			if (current > account.getRefreshTokenExpiryTime()) {
+				return AccountStatus.NEEDS_LOGIN;
+			} else {
+				return AccountStatus.NEEDS_REFRESH;
+			}
+		}
+		if (((current - lastRefreshed) > OSIOCoreConstants.DURATION_24_HOURS)
```
adietish: might improve readability if this was extracted to #wasRefreshed24hAgo

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
```diff
@@ -0,0 +1,162 @@
...
+				|| ((current - lastRefreshed) > ((account.getAccessTokenExpiryTime() - current) / 2))) {
```
adietish: same here, method would add quickly graspable info about what this check is trying to achive

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
```diff
@@ -0,0 +1,162 @@
...
+
+	public String getToken(IResource t) {
+		String token = null;
+
+		ICluster cluster = getModel().getClusters().get(0);
+		List<IAccount> identities = cluster.getAccounts();
+		if (identities.isEmpty()) {
+			token = performLogin(cluster, null);
+		} else {
+			IAccount account = identities.get(0);
```
adietish: what if several accounts?

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
```diff
@@ -0,0 +1,162 @@
...
+				|| ((current - lastRefreshed) > ((account.getAccessTokenExpiryTime() - current) / 2))) {
+			return AccountStatus.NEEDS_REFRESH;
+		}
+		return AccountStatus.VALID;
+	}
+
+	public String getToken(IResource t) {
+		String token = null;
+
+		ICluster cluster = getModel().getClusters().get(0);
```
adietish: what if several clusters?

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/exception/OpenshiftIORefreshException.java
adietish: maybe -1 should be extracted to a public constant so that consumers can test against this state?
As far as I can see this status is - in normal cases - the http response code. So -1 has the special meaning "there's no status". Something that I'd state more prominently like ```public/private static final NO_STATUS = -1```

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/dialog/BrowserBasedLoginDialog.java
```diff
@@ -0,0 +1,111 @@
...
+				if (event.total <= 0)
+					return;
+				int ratio = event.current * 100 / event.total;
+				progressBar.setSelection(ratio);
+			}
+
+			@Override
+			public void completed(ProgressEvent event) {
+				progressBar.setSelection(0);
+				System.out.println("URL=" + browser.getUrl());
```
bdshadow: I think it's better to log it through `OpenShiftIOUIActivator`

plugins/org.jboss.tools.openshift.io.core/plugin.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?eclipse version="3.5"?>
+<plugin>
+   <extension-point id="org.jboss.tools.openshift.io.core.tokenProvider" name="tokenProvider" schema="schema/org.jboss.tools.openshift.io.core.tokenProvider.exsd"/>
```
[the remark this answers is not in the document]
jeffmaury: because I want to have loose coupling so that extension user can cast to Function<IResource, String>

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/preferences/MainPreferencePage.java
```diff
@@ -0,0 +1,121 @@
...
+
+	void onButton(SelectionEvent event) {
+		button.setEnabled(false);
+		removed = true;
+		updateLabel();
+		
+	}
+	void updateLabel() {
+		String msg;
+		ICluster cluster = AccountService.getDefault().getModel().getClusters().get(0);
```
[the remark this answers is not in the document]
jeffmaury: Current implem is single cluster single account. Will create JIRAs

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/preferences/MainPreferencePage.java
```diff
@@ -0,0 +1,121 @@
...
+
+	/* (non-Javadoc)
+	 * @see org.eclipse.jface.preference.PreferencePage#performOk()
+	 */
+	@Override
+	public boolean performOk() {
+		if (removed) {
+			ICluster cluster = AccountService.getDefault().getModel().getClusters().get(0);
+			if (!cluster.getAccounts().isEmpty()) {
+				IAccount account = cluster.getAccounts().get(0);
```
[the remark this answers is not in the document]
jeffmaury: You got it

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/OSIOUtils.java
```diff
@@ -0,0 +1,81 @@
...
+	public static long decodeExpiryFromToken(String token) {
+		String payloads[] = token.split("\\.");
+		Claims claims = (Claims) Jwts.parser().parse(payloads[0] + '.' + payloads[1] + '.').getBody();
+		return  claims.get("exp", Date.class).getTime();
+	}
+
+	public static String computeLandingURL(String endpointURL, String devstudioOsioLandingPageSuffix) {
+		try {
+			URL url = new URL(endpointURL);
+			URL landingURL = new URL(url.getProtocol(), url.getHost(), url.getPort(), devstudioOsioLandingPageSuffix);
```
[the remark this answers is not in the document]
jeffmaury: Maybe URI could to the trick as well. Will check

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
jeffmaury: +1

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
jeffmaury: +1

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
jeffmaury: Unlikely to happen

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
jeffmaury: Unlikely to happen

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
adietish: so why would you support those if not likely?

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
```diff
@@ -0,0 +1,162 @@
...
+			}
+		}
+		if (((current - lastRefreshed) > OSIOCoreConstants.DURATION_24_HOURS)
+				|| ((current - lastRefreshed) > ((account.getAccessTokenExpiryTime() - current) / 2))) {
+			return AccountStatus.NEEDS_REFRESH;
+		}
+		return AccountStatus.VALID;
+	}
+
+	public String getToken(IResource t) {
```
adietish: parameter t is never used

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/DefaultLoginProvider.java
```diff
@@ -0,0 +1,55 @@
...
+				}
+			});
+			return response[0];
+			
+		} else {
+			return loginInUI(cluster, account);
+		}
+	}
+
+	public LoginResponse loginInUI(ICluster cluster, IAccount account) {
```
adietish: Account is never used? This looks like it's handled otherwise in this specific case or it's not used and thus the interface should be changed, too

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/processor/DefaultRequestProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+
+import org.eclipse.swt.browser.Browser;
+import org.jboss.tools.openshift.io.core.LoginResponse;
+import org.jboss.tools.openshift.io.core.OSIOUtils;
+import org.jboss.tools.openshift.io.internal.ui.OpenShiftIOUIActivator;
+
+public class DefaultRequestProcessor implements RequestProcessor {
+
+	private static String SCRIPT = "           var request = (function() {\r\n" +
```
adietish: final?

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/OSIOCoreConstants.java
```diff
@@ -0,0 +1,31 @@
...
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.openshift.io.core;
+
+public final class OSIOCoreConstants {
```
adietish: add private constructor to avoid instantiations?

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
jeffmaury: For further enhancements: the model is design to support it but the implem is yet limited.

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
```diff
@@ -0,0 +1,162 @@
...
+		}
+		return token;
+	}
+
+	private String performLogin(ICluster cluster, IAccount account) {
+		if (null != provider) {
+			LoginResponse response = provider.login(cluster, account);
+			if (null != response) {
+				if (null == account) {
+						String id = OSIOUtils.decodeEmailFromToken(response.getAccessToken());
```
adietish: extract l109-112 into ```private IAccount createAccount(response, cluster)```

plugins/org.jboss.tools.openshift.io.core/src/main/java/org/jboss/tools/openshift/io/core/AccountService.java
jeffmaury: It is there for future evolution: the resource is there in order to provide a kind of context, for example the selected OpenShift.io account could be set at the project level.

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/DefaultLoginProvider.java
jeffmaury: No there is no way to pass account info to the browser. But if we switch to a full HTTP lib implementation, then account would be used.

plugins/org.jboss.tools.openshift.io.core/plugin.xml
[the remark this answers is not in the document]
adietish: how would both of these not be possible with an osgi service?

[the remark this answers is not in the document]
jeffmaury: I thought registering/retrieving an OSGI service was only possible through a Class object but it would be possible with any string. Will try if I can before the code freeze but the LoginProvider would stay as is because I planned to used priority based selection

bdshadow:
@jeffmaury i have 3 more things
1) The "Remove" button is active by default when i go to the Preferences page, even if i'm not logged-in to OSIO and have nothing to remove. It becomes inactive as soon as i click it.

2) i don't think that it's good to log the token to the Error Log or show it in the window after the user is logged-in. Think of the following situations: a) user wants to do a screencast and he don't want to show his token of course to the whole world. b) we store a token in a secure storage, it means it mustn't be shown anywhere and then explicitly print it everywhere.

3) I try to log-in, but click "Close". I get an error window with the message "Error login to cluster Openshift.io". I don't think that an error window is the correct way to interact with user here. There were no errors indeed, a user just changed his mind and decided not to login to OSIO.

jeffmaury:
@bdshadow 
1. Will disable the button if model is empty
2. Good security remark, will remove logging
3. An error is returned because this component is to be called by third party try to get a token and if user will stop the login phase we need to signal them an error (in terms I cannot return a token)

PR #1578, opened 2017-10-04
JBIDE-25055 - initial impl for cdk 3.2 support

robstryker:
This PR is for testing purposes. It is not 100% Complete. 

I am not 100% confident in it's implementation but need feedback.

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/CDK32Server.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/CDK32Server.java
@@ -1,0 +1,51 @@
+/******************************************************************************* 
+ * Copyright (c) 2017 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
+ ******************************************************************************/ 
+package org.jboss.tools.openshift.cdk.server.core.internal.adapter;
+
+import java.util.ArrayList;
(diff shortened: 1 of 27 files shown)


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/ui/internal/CDK3ServerWizardFragment.java
```diff
@@ -110,20 +110,44 @@ protected String findError() {
...
 		return retString;
 	}
 	
+
+	protected String isVersionCompatible(MinishiftVersions versions) {
+		String cdkVers = versions.getCDKVersion();
+		if( cdkVers == null ) {
+			return "Cannot determine CDK version.";
+		}
+		if( cdkVers.startsWith("3.0.") || cdkVers.startsWith("3.1")) {
```
adietish: i see this check (negated or not in several places), worth extracting and reusing?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/CDK32Server.java
```diff
@@ -0,0 +1,51 @@
...
+			return true;
+		}
+		return false;
+	}
+		
+	public static String[] getArgsWithProfile(IServer server, String[] args) {
+		if( supportsProfiles(server)) {
+			String profileName = server.getAttribute(CDK32Server.PROFILE_ID, (String)null);
+			if( profileName != null ) {
+				ArrayList<String> al = new ArrayList<String>();
```
adietish: plz use ```List``` in the type declaration

odockal:
Using 3.2+ adapter and cdk 3.2-alpha.1: performing Start/Stop/Restart operations of server adapter works, OS connection is created during starting. Discovery button (in editing OS connection) shows proper registry url, Docker Explorer shows new connection, I can log into OS via web browser. Things seems to be working on my side as well. I have to refactor cdk itests (new adapter, new check of cdk version) but this PR seems alright. I will test it more tomorrow.

PR #1582, opened 2017-10-06
[JBIDE-24868] debugging: increase (route) timeout, disable probe

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/LivenessProbe.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/LivenessProbe.java
@@ -1,0 +1,105 @@
+/******************************************************************************* 
+ * Copyright (c) 2017 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
+ ******************************************************************************/
+package org.jboss.tools.openshift.internal.core.server.debug;
+
+import java.util.Collection;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.osgi.util.NLS;
+import org.jboss.tools.openshift.core.server.OpenShiftServerUtils;
+
+import com.openshift.restclient.model.IContainer;
+import com.openshift.restclient.model.IDeploymentConfig;
+import com.openshift.restclient.model.probe.IProbe;
+
+/**
+ * A class that allows to manipulate an existing liveness probe for a given deployment config.
+ * 
+ * @author Andre Dietisheim
+ * 
+ * @see <a href="https://kubernetes.io/docs/api-reference/v1.8/#probe-v1-core"/>
+ */
+public class LivenessProbe {
+
+	public static final int INITIAL_DELAY = 1 * 60 * 60; // 1h
+	
+	private IDeploymentConfig dc;
+
+	LivenessProbe(IDeploymentConfig dc) {
+		this.dc = dc;
+	}
+	
+	/**
+	 * Sets the initial delay of the liveness probe for the given deployment config.
+	 * No action is taken if there's no such probe. Returns {@code true} if the
+	 * probe was modified, {@code false} otherwise.
+	 * 
+	 * @param delay
+	 * @param monitor
+	 * @return
+	 * @throws CoreException 
+	 */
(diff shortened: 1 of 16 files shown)


bdshadow:
@adietish please, add "Signed-off-by"

adietish:
@bdshadow @robstryker review please?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugMode.java
```diff
@@ -155,25 +161,114 @@ public OpenShiftDebugMode disableDevmode() {
...
+		if (route == null) {
+			OpenShiftCoreActivator.pluginLog().logInfo(
+					NLS.bind("Could not increase timeout for debugging: could not find any route for resource {0}", resource.getName()));
+			return;
+		}
+
+		OpenShiftCoreActivator.pluginLog().logInfo(
+				NLS.bind("Setting haproxy timeout for route {0}", route.getName()));
+		
+		if (!hasTimeoutAnnotation(route)) {
```
jeffmaury: What if the deployment already has a timeout but it is a low one. I think we should back it up in one of our own annotation so that we can recover from several invocations after JBoss Tools restarts

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugMode.java
```diff
@@ -155,25 +161,114 @@ public OpenShiftDebugMode disableDevmode() {
 	 * @throws CoreException
 	 */
 	public OpenShiftDebugMode execute(IProgressMonitor monitor) throws CoreException {
-		IDeploymentConfig dc = getDeploymentConfig(context, monitor);
+		Connection connection = OpenShiftServerUtils.getConnection(context.getServer());
+		IResource resource = OpenShiftServerUtils.getResource(context.getServer(), monitor);
+		IDeploymentConfig dc = getDeploymentConfig(resource, connection, monitor);
```
jeffmaury: What if there is no DC but an RC ?

adietish:
testPR

adietish:
latest version of this PR sets a route timeout and "disables" (increases the initial delay)  livenessProbe if it exists. Both values are also restored once you get from debug to run again.
@jeffmaury @bdshadow can you please test it?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugMode.java
adietish: that's part of another jira that we have.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/LivenessProbe.java
```diff
@@ -0,0 +1,104 @@
...
+	public boolean setInitialDelay(DebugContext context, IProgressMonitor monitor) throws CoreException {
+		monitor.subTask(NLS.bind("Setting liveness probe initial delay to {0}", INITIAL_DELAY));
+
+		IProbe probe = getLivenessProbe(dc);
+		if (probe == null) {
+			return false;
+		}
+
+		int initialDelay = probe.getInitialDelaySeconds();
+		if (initialDelay >= INITIAL_DELAY) {
```
jeffmaury: I don't understand the rationale for this case and why don't you save the original value as an annotation in the deployment config ?

adietish: @jeffmaury if the existing delay is larger than what we'd want for debugging, why should we change it?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/server/debug/OpenShiftDebugMode.java
```diff
@@ -155,25 +155,72 @@ public OpenShiftDebugMode disableDevmode() {
...
+		if (route != null) {
+			safeSend(route, connection, monitor);
+		}
+	}
+
+	private boolean updateDc(IDeploymentConfig dc, Connection connection, IProgressMonitor monitor)
+			throws CoreException {
+		boolean dcUpdated = updateDebugmode(dc, context, monitor) 
+				| updateDevmode(dc, context, monitor) 
+				| updateLifenessProbe(dc, context, monitor);
```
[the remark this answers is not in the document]
jeffmaury: Sonar is so dummy

adietish:
@jeffmaury great, thx!

PR #1588, opened 2017-10-10
JBIDE-25167 - both cdk3 runtime detectors validate against cdk3 / 3.1

robstryker:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/detection/CDK3RuntimeDetector.java
```diff
@@ -38,17 +38,15 @@
 	public static final String PROP_CDK_VERSION = "cdk.version";
 	
 	public static final String OVERRIDE_MINISHIFT_LOCATION = "OVERRIDE_MINISHIFT_LOCATION";
-	
+
 	@Override
 	protected boolean validate(File root) {
 		boolean matchesHomeMinishift = isHomeDirectory(root.getParentFile()) && 
 				".minishift".equals(root.getName()) && super.validate(root);
-		if( matchesHomeMinishift )
```
robstryker: This caused both detectors to match if the folder in question was ~/.minishift

PR #1592, opened 2017-10-12
[JBIDE-25172] - default profile and 'cdk' file wont exist if specific…

robstryker:
… profile was setup

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/CDKConstants.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/CDKConstants.java
@@ -30,8 +30,11 @@ public interface CDKConstants {
 	
 	
 	public static final String CDK_RESOURCE_VAGRANTFILE = "Vagrantfile";
+	public static final String CDK_RESOURCE_DOTMINISHIFT= ".minishift";
 	public static final String CDK_RESOURCE_DOTCDK= ".cdk";
 	public static final String CDK_RESOURCE_CDK= "cdk";
+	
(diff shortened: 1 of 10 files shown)


robstryker:
This needs more work

robstryker:
This patch solves JBIDE-25172 and JBIDE-25170. 

JBIDE-25170 is the issue where an empty profile tag should still send '--profile minishift' to the command line arguments, in case the user has changed the default profile of the system. 

JBIDE-25172 fixes the adapter being unable to locate credentials when a user has a fresh cdk3.2 install and chooses to set up only a specific profile (./minishift setup-cdk --profile test555)  and never sets up the default profile. In this case, ~/.minishift/cdk file will not exist, and the adapter must instead look for ~/.minishift/profiles/test555/cdk

These two patches are closely related to each other and cannot be committed separately without refactoring the two patches. I prefer to squash them and commit them as an indivisible group.

adietish:
I can replicate the issue without the suggested changes in code. Now testing with the PR.

adietish:
the PR is fixing the usecase as described in jira.

adietish:
@robstryker I am wondering why the server adapter wizard nows suggests by default the "minishift" profile?

adietish:
The other usecase discussed with @robstryker also works for me:

* EXEC: setup-cdk  without a profile
* EXEC: setup-cdk with --profile test
* ASSERT: make sure test is the active profile: ```minishift profile set test```. ```minishift  profile list``` shows that test is the active profile
* EXEC: in Eclipse try to launch a cdk server adapter that has no profile set. 

**BEFORE THE PR**
Profile **test** is launched

**AFTER THE PR**
Profile **minishift** (the default profile) is launched.

PR #1600, opened 2017-10-23
[JBIDE-25169] - Deadlock when tokenProvider is invoked during start

jeffmaury:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/DefaultLoginProvider.java
+++ b/plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/DefaultLoginProvider.java
@@ -31,6 +31,8 @@ import org.jboss.tools.openshift.io.internal.ui.dialog.BrowserBasedLoginDialog;
  */
 public class DefaultLoginProvider implements LoginProvider {
 
+	private static final int TIMEOUT_JOB_ON_UI_THREAD = 10000;
+
 	private class LoginJob extends UIJob {
 		private boolean runninginUI = false;
 		private boolean shouldRun = true;
@@ -81,7 +83,7 @@ public class DefaultLoginProvider implements LoginProvider {
 		job.schedule();
 		IProgressMonitor monitor = new NullProgressMonitor();
 		try {
-			if (job.join(10000, monitor)) {
+			if (job.join(TIMEOUT_JOB_ON_UI_THREAD, monitor)) {
 				return job.getResponse();
 			} else {
 				throw new InterruptedException();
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/DefaultLoginProvider.java
```diff
@@ -25,23 +31,80 @@
...
 
+	LoginResponse runInJob(ICluster cluster, IAccount account) {
+		LoginJob job = new LoginJob(cluster, account);
+		job.schedule();
+		IProgressMonitor monitor = new NullProgressMonitor();
+		try {
+			if (job.join(10000, monitor)) {
+				return job.getResponse();
+			} else {
+				throw new InterruptedException();
```
adietish: what else than us here is throwing InterruptedException (Eclipse shutting down?)? If it's just us then I'm wondering if we couldn't simply have the code of l92 directly following here without the throw.
Furthermore I see us trying to join twice before we finally give up:
1) l84
2) l94
What is the reason for trying twice?

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/DefaultLoginProvider.java
```diff
@@ -25,23 +31,80 @@
...
+		
+		public LoginJob(ICluster cluster, IAccount account) {
+			super("Logging to OpenShift.io");
+			this.cluster = cluster;
+			this.account = account;
+		}
+
+		@Override
+		public IStatus runInUIThread(IProgressMonitor monitor) {
+			if (shouldRun && !monitor.isCanceled()) {
```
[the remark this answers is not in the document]
jeffmaury: It is there for safety because cancelling the job may be delayed due to lock mechanisms and as this job will show some UI I though it was important to be safe

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/DefaultLoginProvider.java
```diff
@@ -25,23 +31,80 @@
...
 			return loginInUI(cluster, account);
 		}
 	}
 
+	LoginResponse runInJob(ICluster cluster, IAccount account) {
+		LoginJob job = new LoginJob(cluster, account);
+		job.schedule();
+		IProgressMonitor monitor = new NullProgressMonitor();
+		try {
+			if (job.join(10000, monitor)) {
```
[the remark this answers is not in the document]
jeffmaury: Will do

plugins/org.jboss.tools.openshift.io.ui/src/main/java/org/jboss/tools/openshift/io/internal/ui/DefaultLoginProvider.java
jeffmaury: The contract of the join method may throw InterruptedException

jeffmaury: We are trying twice because we want to prevent deadlock (that you may have when an LSP is started when you open the editor). So we first allocate 10s for the job to be on the UI or completed and if we see the job is running on the UI then we wait for the job to finish.

PR #1601, opened 2017-10-24
JBIDE-24438 - Read port from Connection URL

apupier:
read port from Server Adapter Connection URL instead of using an
hardcoded one. It is also taking care when no port is specified (such as
open.paas.redhat.com and surely most of publicly available instances)

I don't know how to unit test it as there are several static methods and tests are not in the same package (no fragment used). Suggestions?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenshiftJMXConnectionProvider.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenshiftJMXConnectionProvider.java
@@ -10,6 +10,8 @@
  ******************************************************************************/ 
 package org.jboss.tools.openshift.core.server;
 
+import java.net.MalformedURLException;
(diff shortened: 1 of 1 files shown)


adietish:
@robstryker dare to review this please?

robstryker:
So... I like the patch for ensuring that the connection is correct when using createConnection(). 

However, if the details are likely to change (ie a pod is shut down and a new one brought up), wont this jmx connection then go stale? Don't you need a listener somewhere to respond and either change the details of the connection or remove it and create a new one?

apupier:
> However, if the details are likely to change (ie a pod is shut down and a new one brought up), wont this jmx connection then go stale? Don't you need a listener somewhere to respond and either change the details of the connection or remove it and create a new one?

yes, the JMX connection won't be working anymore, see https://issues.jboss.org/browse/JBIDE-25288
but it is not due to this port. The port of the global connection will remain the same.

jeffmaury:
testPR

PR #1603, opened 2017-10-25
[JBIDE-25115] server editor: warn "restart to activate changes"

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorModel.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorModel.java
@@ -59,61 +59,44 @@ public class OpenShiftServerEditorModel extends ServerSettingsWizardPageModel {
 
 	private void initChangelisteners(IServerWorkingCopy server, ServerEditorSection section) {
 		this.devmodeKeyObservable = BeanProperties.value(PROPERTY_DEVMODE_KEY).observe(this);
-		devmodeKeyObservable.addValueChangeListener(
-				event -> {
-					if (initializing) {
-						return;
-					}
-					section.execute(
-							new SetStringCommand(server, "SetDevmodeKey", 
-									OpenShiftServerUtils.ATTR_DEVMODE_KEY, 
-									(String) event.diff.getOldValue(), (String) event.diff.getNewValue()));
-				});
+		devmodeKeyObservable.addValueChangeListener(event -> {
+			if (initializing) {
+				return;
+			}
+			section.execute(new SetStringCommand(server, "SetDevmodeKey", OpenShiftServerUtils.ATTR_DEVMODE_KEY,
+					(String) event.diff.getOldValue(), (String) event.diff.getNewValue()));
+		});
 
 		this.debugPortKeyObservable = BeanProperties.value(PROPERTY_DEBUG_PORT_KEY).observe(this);
-		debugPortKeyObservable.addValueChangeListener(
-				event -> {
-					if (initializing) {
-						return;
-					}
-					section.execute(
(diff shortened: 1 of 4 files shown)


adietish:
@bdshadow dare to review please?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
```diff
@@ -762,13 +774,31 @@ public boolean isCanceled() {
 
 	@Override
 	public IStatus[] getSaveStatus() {
-		return new IStatus[] { formPresenterSupport.getCurrentStatus() };
+		return new IStatus[] { getFormPresenterStatus() };
+	}
+
+	private IStatus getFormPresenterStatus() {
+		IStatus saveStatus = Status.OK_STATUS;
+		if (formPresenterSupport != null
```
bdshadow: i don't think that `formPresenterSupport` can be `null`.
If i miss something, then these null checks must be also added to `dispose` method

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
```diff
@@ -144,8 +146,18 @@ public void createSection(Composite parent) {
...
 		section.setClient(container);
 
-		this.formPresenterSupport = new FormPresenterSupport(
-				new FormEditorPresenter(getManagedForm().getForm().getForm()), dbc);
+		IObservableValue<Boolean> dirtyStatusObservable = DataBindingUtils.createDirtyStatusObservable(getServerEditor());
+		dbc.addValidationStatusProvider(new MultiValidator() {
+
+			@Override
+			protected IStatus validate() {
+				if (Boolean.TRUE.equals(dirtyStatusObservable.getValue())) {
```
bdshadow: can be just `if (dirtyStatusObservable.getValue()) {`

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorModel.java
```diff
@@ -436,7 +437,11 @@ public IStatus redo(IProgressMonitor monitor, IAdaptable adapt) {
 	public void setInitializing(boolean initializing) {
 		this.initializing = initializing;
 	}
-	
+
+	public boolean isInitializing() {
```
bdshadow: haven't found any usage of this new method. Can be removed

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorSection.java
adietish: I dont think so since dirtyStatus can also be just null.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/OpenShiftServerEditorModel.java
adietish: I see it being set from the job that loads the resources:

OpenShiftServerEditorSection#LoadResourcesJob:
```
	private class LoadResourcesJob extends Job {
		...
		@Override
		protected IStatus run(IProgressMonitor monitor) {
			model.setInitializing(true);
			...
			} finally {
				model.setInitializing(false);
			}
			...
```

PR #1606, opened 2017-10-31
JBIDE-25308 - minishift_home is not passed to launch command

robstryker:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDK3LaunchController.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDK3LaunchController.java
@@ -56,7 +56,7 @@ public class CDK3LaunchController extends AbstractCDKLaunchController
 
 	@Override
 	public void initialize(ILaunchConfigurationWorkingCopy wc) throws CoreException {
-		final IServer s = ServerUtil.getServer(wc);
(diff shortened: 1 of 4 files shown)


robstryker:
testPR

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDK3LaunchController.java
```diff
@@ -98,12 +98,23 @@ static String getProfileString(IServer server) {
 		}
 		return profiles;
 	}
+	
+	protected IServer getServerFromLaunch(ILaunchConfigurationWorkingCopy wc) throws CoreException {
```
adietish: I dont see the benefit of this additional method.

adietish:
@robstryker I let you dismiss @jeffmaury's review and merge.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/controllers/CDK3LaunchController.java
robstryker: I think I had originally put it there for the test to override, but it became unnecessary. Still, doesn't hurt anything.

PR #1608, opened 2017-11-03
[JBIDE-25121] - Connection wizard: Image Registry URL is not decorated when not valid (and validation seems weird)

jeffmaury:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@jeffmaury decorations now show up, great!
But the validator seems to consider "https:" as a valid URL, is it?

jeffmaury:
Yes because 'https:' is semantically valid for an URL

adietish:
@jeffmaury hmm, but that's not how I'd read https://tools.ietf.org/html/rfc3986#section-3 where it looks like query and fragment only seems optional.

PR #1613, opened 2017-11-08
Jbide 25303 provide spring boot profile for open shift

apupier:
Early PR to gather feedback and ensure going in the right direction.

I created most of the plugins/features. Will we need a separate test plugin? it seems that you are reusing the same plugins for almost all tests.
I checked manually that it returns "true" when a SpringBoot project is used.

What is the best way to test this kind of things? I have not found tests for EAP or NodeJS profiles. Can you point me to them?


adietish:
@jeffmaury this is my fault, it is what I suggested in the discussion. But yes, you're absolutely right.

apupier:
I created this other PR to have it in the same plugin than for EAP: https://github.com/jbosstools/jbosstools-openshift/pull/1614

PR #1624, opened 2017-11-22
JBIDE-24804 - ResourceLabelsPageModel should never set instance var t…

robstryker:
…o unmodifiable list

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/common/ResourceLabelsPageModel.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/common/ResourceLabelsPageModel.java
@@ -36,7 +36,7 @@ public class ResourceLabelsPageModel extends ObservablePojo implements IResource
 
 	@Override
 	public void setLabels(List<Label> labels) {
-		firePropertyChange(PROPERTY_LABELS, this.labels, this.labels = labels);
+		firePropertyChange(PROPERTY_LABELS, this.labels, this.labels = new ArrayList<Label>(labels));
 	}
 
 	@Override


robstryker:
I actually feel strongly that this is the right solution. We don't want random people who call setLabels to be able to add an unmodifiable list, which would prevent addLabel working in the future. We also might not want someone who is a listener to the model to be able to change the model directly without using the model's methods itself. 

So I strongly believe that it is appropriate for the events to be fired with unmodifiable lists, and for setLabels(etc) to properly protect the model by instantiating a new instance of a modifiable list. 

It's also worth noting that this was Dmitri's solution on the JIRA, and I agree with his solution.

jeffmaury:
I agree with you that setLabels is the right place to set a modifiable lib. But then why are we duplicating the list as an unmodifiable one before calling setLabels. I would keep your code but modify addLabel, removeLabel and updateLabel not to pass an unmodifiable list

adietish:
@jeffmaury where do you see the code duplicating into an unmodifiable list and then calling #setLabels? Couldn't find such occurrences, just the property firing which is using those unmodifiable list which doesnt seem utterly wrong to me?

jeffmaury:
Have a look at the `addLabel` method as an example

bdshadow:
i agree with @jeffmaury and i don't think it makes sense to `fireIndexedPropertyChange` with wrapping lists into `Collections.unmodifiableList` all the time

robstryker:
It makes perfect sense to send out unmodifiable lists. This way the listener who gets the list can't make edits to the list except through the approved API of 'addLabel'.   

If you send out a normal list, that can be modified, any listener can simply add items to the list. This will bypass the model's own internal recordkeeping. If a listener decides to add or remove items from the list directly, any OTHER listeners won't get events!

jcantrill:
It is my understanding that consumer's of a fired event are receiving:

* What property changed
* The new value(s)
* The old value(s)

It's my opinion is that 'old' values should be unmodifiable; its a 'fact' used to make some determination and not really designed to be alterable.   The code should probably copy the set into a new one so that it can be modified by subsequent calls.

jeffmaury:
I'm not against unmodifiable lists. I'm just curious why we need so may copies when processing a single operation (add/remove label)

adietish:
@jeffmaury #addLabel creates 2 new collections:
1) the old list, so that it can fire it
2) an unmodifiable copy of the new list (so that listeners can't modify it).

I dont see what's wrong with that agreeing with the arguments brought up by @robstryker and @jcantrill. 

+1 for this PR.

PR #1632, opened 2017-12-04
JBIDE-25149 - code cleanup

robstryker:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/CDK32Server.java
```diff
@@ -43,7 +44,7 @@ public static boolean supportsProfiles(IServer server) {
 				profileName = CDK32Server.MINISHIFT_DEFAULT_PROFILE;
 			}
 			if( profileName != null ) {
-				ArrayList<String> al = new ArrayList<String>();
+				List<String> al = new ArrayList<String>();
```
jeffmaury: You can use the diamond notation for the ArrayList creation

PR #1641, opened 2017-12-19
[JBIDE-25514] - Build to not allow goal dependency:tree

jeffmaury:
- Removed version of maven-dependency-plugin as its given by parent

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


nickboldt:
Assuming you've removed these versions because they're updated in parent pom to 3.0.2 in pluginManagement section, then +1. Not sure why you had "NO-MVN-MAN-VER" in there, that seems strange.

PR #1645, opened 2017-12-22
 [JBIDE-25111] impl'd deleting individually selected resources 

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
testPR

adietish:
@jkopriva can you please test this please?

jkopriva:
@adietish can you please look at the jenkins failure? I am also not able to build repo with maven. Thanks!

adietish:
testPR

adietish:
@jkopriva sorry for the inconvenience. I missed the format failures before leaving for holidays. I pushed the fixes.

jkopriva:
@adietish no problem, I will run ITests against it and take a closer look at functionality

jeffmaury:
Works for me but I have a little UI issue: when the label is complex and thus log, there is no way to see it fully and seems the tooltip is misleading:
image

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/resource/DeleteResourcesWizardModel.java
```diff
@@ -0,0 +1,176 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Red Hat, Inc. Distributed under license by Red Hat, Inc.
```
jeffmaury: Should be 2018

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/resource/DeleteResourcesWizardModel.java
```diff
@@ -0,0 +1,176 @@
...
+import com.openshift.restclient.model.IPod;
+import com.openshift.restclient.model.IReplicationController;
+import com.openshift.restclient.model.IResource;
+
+/**
+ * @author Andre Dietisheim
+ */
+public class DeleteResourcesWizardModel extends ObservablePojo {
+
+	private static final String[] ALL_RESOURCE_KINDS = { ResourceKind.BUILD, ResourceKind.BUILD_CONFIG,
```
jeffmaury: I'm wondering if we can't get this list from the client lib ? What if we have a kubernetes deployment (k8s Deployment instead of OpenShift DeploymentConfig)

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/resource/ResourceLabelFilter.java
```diff
@@ -0,0 +1,49 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Red Hat, Inc.
```
jeffmaury: Should be 2018

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/internal/common/core/util/KeyValueFilterFactoryTest.java
```diff
@@ -0,0 +1,194 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Red Hat, Inc.
```
jeffmaury: Should be 2018

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/resource/DeleteResourcesWizardModel.java
adietish: @jeffmaury all types that are supported by the client lib are constants in the ResourceType class. But then not all of them are types that we want/can to delete (ex. Policy, PolicyByinding etc.)

adietish:
@jeffmaury: I corrected 2017->2018 and I suggest that we now merge this as is while filing jiras for the issues. I can then resolve these in the current sprint.

jkopriva:
@adietish Is there a possibility to add some progress dialog? When you are using slower OpenShift instance (e.g. https://console.starter-us-east-2.openshift.com) it takes a few seconds (1-2) until the list is loaded, but you do not see if something is happening or not.
Also, when I delete pod, the new pod is spawned, is this expected? Or when we deleting a pod, application should be scaled to 0?

adietish:
@jkopriva +1 on using a progress bar. It looks like OpenShift online got considerably slower these days. When I developed this I tested it with online and loading happened instantly while these days it takes several seconds. As suggested above I think that we should merge as-is and fix in the current sprint. Agree?

jkopriva:
@adietish ok, we can merge it, I will create a JIRA for progress bar and I will assign it to you.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/util/ResourceKindAndNameComparator.java
```diff
@@ -0,0 +1,36 @@
...
+		}
+		return compareName(first, second);
+	}
+
+	private int compareName(IResource first, IResource second) {
+		return first.getName().compareTo(second.getName());
+	}
+
+	private int compareKind(IResource first, IResource second) {
+		return ObjectUtils.compare(first.getKind(), second.getKind());
```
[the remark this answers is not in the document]
adietish: @jeffmaury nothing that I had introduced, I just extracted this class from within another one so that I can reuse it. So I'm guess that it's about avoiding null checks?

adietish:
@jeffmaury @jkopriva latest update added scrollbars to the table (so that one can scroll the table in case the label column wont show all labels (without scrolling).

adietish:
@jeffmaury sorry, my fault, composite visibility behaves differently on different platforms. I didnt think at it. I fixed it in the latest update. Please re-review. Thx!

adietish:
@jeffmaury missing test added to commit. We should really try to figure out how to show the scrollbars in Windows.

adietish:
@jeffmaury I suggest that we file the scrollbars issue to jira and try to fix it for GA. In the meantime, let's merge this in. Agree?

PR #1646, opened 2018-01-05
Rewrite OCBinary requirement to be configurable

odockal:
- Downloaded oc is copied from extracted folder instead of being symlinked
   - OCBinary takes multiple parameters, bool: cleanup, setInPrefs, ocExists, downloadNewestOC
   - String param: pathToOC

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/preference/page/OpenShift3PreferencePage.java
+++ b/test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/preference/page/OpenShift3PreferencePage.java
@@ -36,7 +36,7 @@ public class OpenShift3PreferencePage extends PreferencePage {
 	 * @param path path to oc binary
 	 */
 	public void setOCLocation(String path) {
-		new LabeledText(OpenShiftLabel.TextLabels.OC_LOCATION).setText(path);
+		getOCLocation().setText(path);
 	}
 	
 	/**
@@ -44,6 +44,10 @@ public class OpenShift3PreferencePage extends PreferencePage {
 	 */
 	public void clearOCLocation() {
 		setOCLocation("");
+	}
+	
+	public LabeledText getOCLocation() {
+		return new LabeledText(OpenShiftLabel.TextLabels.OC_LOCATION);
 	}
 }
 
--- a/test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/requirement/OpenShiftCommandLineToolsRequirement.java
+++ b/test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/requirement/OpenShiftCommandLineToolsRequirement.java
@@ -18,85 +18,206 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.net.MalformedURLException;
+import java.net.URI;
 import java.net.URL;
 import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.util.Properties;
 
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.osgi.util.NLS;
 import org.eclipse.reddeer.common.logging.Logger;
 import org.eclipse.reddeer.junit.requirement.Requirement;
+import org.eclipse.reddeer.workbench.ui.dialogs.WorkbenchPreferenceDialog;
 import org.jboss.tools.openshift.reddeer.exception.OpenShiftToolsException;
+import org.jboss.tools.openshift.reddeer.preference.page.OpenShift3PreferencePage;
 import org.jboss.tools.openshift.reddeer.requirement.OpenShiftCommandLineToolsRequirement.OCBinary;
+import org.jboss.tools.openshift.reddeer.utils.DatastoreOS3;
 import org.jboss.tools.openshift.reddeer.utils.FileHelper;
+import org.jboss.tools.openshift.reddeer.utils.SystemProperties;
+import org.jboss.tools.openshift.reddeer.utils.TestUtils;
 
 /**
  * Requirement to download and extract OpenShift command line tools binary which is necessary 
- * for some functionality of OpenShift tools.
+ * for some functionality of OpenShift and CDK tools. 
+ * Loads up properties file with links to different oc versions.
  *  
  * @author mlabuda@redhat.com
  * @author adietish@redhat.com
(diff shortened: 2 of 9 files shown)


test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/preference/page/OpenShift3PreferencePage.java
```diff
@@ -45,5 +45,9 @@ public void setOCLocation(String path) {
 	public void clearOCLocation() {
 		setOCLocation("");
 	}
+	
+	public LabeledText getOCLocation() {
+		return new LabeledText(OpenShiftLabel.TextLabels.OC_LOCATION);
```
jkopriva: @odockal you should distinguish between linux and win: on Linux is label "'oc' executable location", but on windows is "'oc.exe' executable location"

jkopriva:
@odockal I suggest to leave only 2 versions of oc client: latest stable and version 1.1 and move these parameters to pom.xml. If it would be in pom.xml user can specify in maven build which version of oc should be downloaded - no need to change something in code.

@adietish is there a reason for having multiple version of oc client? In my opinion we should test only latest stable, and also we need older version(something under 1.1.1), because in connection wizard is check for oc version("OpenShift Client version 1.1.1 or higher is required to avoid rsync issues.").

test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/preference/page/OpenShift3PreferencePage.java
odockal: @jkopriva got it, thanks!

test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/requirement/OpenShiftCommandLineToolsRequirement.java
```diff
@@ -23,48 +23,130 @@
...
+					StandardCopyOption.REPLACE_EXISTING).toAbsolutePath().toString();
+		} catch (IOException e) {
+			throw new OpenShiftToolsException(NLS.bind("Could not copy {0} to {1}:\n{2}", 
+					new Object[] { OCBinaryFile.get().getFile().getAbsolutePath(), downloadedOCBinary.getAbsolutePath(), e }));
 		}
 	}
 
+	@SuppressWarnings("unused")
+	@Deprecated
 	private void createSymlink(File downloadedOCBinary) {
```
jkopriva: @odockal This could be deleted. Use of this method leads to an error on Windows machines: "A required privilege is not held by the client..."

odockal:
Changes updated. I have added properties into pom.xml as well as I kept oclinks.properties files in cases where tests would be run within IDE and one cannot use pom defined properties. In such case one must put them into VM arguments which will make it complicated, then and only then properties from file will be loaded...

jkopriva:
@odockal if works for me, user can specify oc version, otherwise default version is used. +1
@adietish can we proceed with merge?

test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/utils/SystemProperties.java
```diff
@@ -29,19 +30,38 @@
...
 	public static String getRequiredProperty(String key, String[] validValues, String errorMessage) {
+		return getRequiredProperty(key, null, validValues, errorMessage);
+	}
+
+	public static String getRequiredProperty(String key, String defaultValue, String[] validValues, String errorMessage) {
 		assertTrue(StringUtils.isNotBlank(key));
 		String value = System.getProperty(key);
-		assertTrue(errorMessage, StringUtils.isNotBlank(value));
+		if (StringUtils.isBlank(value)) {
+			if (defaultValue != null && !defaultValue.isEmpty()) {
```
adietish: could use StringUtils#isBlank for better readability

test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/utils/TestUtils.java
```diff
@@ -142,6 +148,41 @@ public static String getValueOrDefault(String value, String defaultValue) {
...
+	public static String getProjectAbsolutePath(String... path) {
+
+		// Construct path
+		StringBuilder builder = new StringBuilder();
+		for (String fragment : path) {
+			builder.append("/" + fragment);
+		}
+
+		String filePath = "";
+		filePath = System.getProperty("user.dir");
```
adietish: why declare and initialise with an empty string that immediately gets overwritten by the system property?

test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/requirement/OpenShiftCommandLineToolsRequirement.java
```diff
@@ -18,85 +18,204 @@
...
  *
  */
 public class OpenShiftCommandLineToolsRequirement implements Requirement<OCBinary> {
 
 	private static final String CLIENT_TOOLS_DESTINATION = "binaries";
 	private static final String SUFFIX_TAR_GZ = ".tar.gz";
 	private static final String SUFFIX_ZIP = ".zip";
+	private WorkbenchPreferenceDialog dialog;
+	private OpenShift3PreferencePage page;
+	private static final Properties properties = TestUtils.readPropertiesFile(TestUtils.getProjectAbsolutePath("resources" + File.separator + "oclinks.properties"));
```
adietish: I'd group this with the other static entries at l58

test-framework/org.jboss.tools.openshift.reddeer/src/org/jboss/tools/openshift/reddeer/requirement/OpenShiftCommandLineToolsRequirement.java
```diff
@@ -18,85 +18,204 @@
...
+		
+		/**
+		 * If set to true, given oc will be set into 
+		 * Preferences -> JBoss Tools -> OpenShift 3 -> oc executable location field
+		 * @return true or false
+		 */
+		boolean setOCInPrefs() default false;
+		
+		/**
+		 * Will delete downloaded oc, is not used when ocExists=true
```
adietish: isn't "cleanup=false" missing in the javadoc?

odockal: @adietish all comments taken into account, ready to push it. Unless you are still reviewing...

PR #1662, opened 2018-01-30
[JBIDE-25670] deleting resources in an enforced sequence

adietish:
Before: each resource was deleted in an individual thread resulting in inappropriate removal ordering (ex. pod being deleted before rc, causing pod tp be recreated etc.)

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@robstryker dare to review please?

adietish:
@jeffmaury this change fixes the issue that one can see when using open.pass.redhat.com. The other change is that now resources that cannot be deleted are reported. I'm not sure if that's fine because there are mostly resources that one cannot delete like some secrets. On the other hand I feel like we should report the errors, even if they're expected. Any thoughts?

PR #1663, opened 2018-02-01
jbide-25551: disable formatter until Oxygen.3 release in March 2018

bdshadow:
after trying a workaround described here https://github.com/revelc/formatter-maven-plugin/issues/260: wrap javadocs with html into @formatter:off/on, i decided to disable the plugin by now, because this workaround didn't help for some reason.

So we need to wait for `org.eclipse.jdt.core` release of version equal or above 3.14.0 in March 2018.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,7 @@
 					<!-- skipFormatting disables both format and validate goals, 
 					think of it as a switch for the whole plugin;
 					To apply from terminal use -Dformatter.skip=false -->
-					<!-- <skipFormatting>true</skipFormatting> --> 
+					<skipFormatting>true</skipFormatting>
 						<directories>
 							<directory>src</directory>
 						</directories>


ctubbsii:
FWIW, version 2.7.0 of the formatter-maven-plugin does not have this problem, and should work fine. However, it may have a bug with respect to enum formatting that I have not experienced myself, but was fixed in 2.7.1. So, if you need a formatter, but 2.7.1 is problematic, 2.7.0 should still be good for most people.

adietish:
@bdshadow to have a workable solution right now I agree on disabling the formatter for now. I believe that we should check out 2.7.0 and see if it works for after the freeze.

adietish:
@ctubbsii thx for your comments on the matter. 2.7.0 is actually what I would have suggested using :) We're very tight in terms of schedule though.

PR #1671, opened 2018-02-14
CDK credentials error test

odockal:
New cdk itests for checking erroneous state during cdk start up with wrong credentials passed into server adapter

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


odockal:
https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Devstudio/view/devstudio_integration_tests/job/cdk.itests/317/

odockal:
Moving all label/shell/jobs names into one class in next PR.

test-framework/org.jboss.tools.cdk.reddeer/src/org/jboss/tools/cdk/reddeer/core/enums/CDKHypervisor.java
```diff
@@ -0,0 +1,67 @@
...
+	public static CDKHypervisor getDefaultHypervisor() {
+		switch(CDKRuntimeOS.get().getRuntimeName()) {
+			case "linux":
+				return KVM;
+			case "darwin":
+				return XHYVE;
+			case "win":
+				return HYPERV;
+			default:
+				return null;
```
jkopriva: What happen, if we return null? Shouldn't we throw an exception?

jkopriva:
+1

PR #1689, opened 2018-03-05
JBDS-4682 - return empty array instead of default impl null

robstryker:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/CDKServer.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/CDKServer.java
@@ -20,6 +20,7 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.equinox.security.storage.StorageException;
 import org.eclipse.wst.server.core.IModule;
 import org.eclipse.wst.server.core.IServerWorkingCopy;
+import org.eclipse.wst.server.core.ServerPort;
 import org.eclipse.wst.server.core.internal.Server;
 import org.eclipse.wst.server.core.model.ServerDelegate;
 import org.jboss.ide.eclipse.as.core.util.ServerNamingUtility;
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/core/internal/adapter/CDKServer.java
```diff
@@ -192,5 +193,11 @@ public boolean skipRegistration() {
 	public boolean skipUnregistration() {
 		return getServer().getAttribute(CDKServer.PROP_SKIP_UNREG, false);
 	}
+	
+	@Override
+	public ServerPort[] getServerPorts() {
+		return new ServerPort[0];
```
dgolovin: This fixes NPE 👍 

Would that make sense to add real ports (should not be part of this PR) that could be monitored in running CDK?

robstryker: I never found the WTP's port monitoring to be good at all. I don't think I was ever able to get any useful information out of it ever.

robstryker: http://tools.jboss.org/blog/2015-05-28-port-monitoring.html

This is a description of how the port monitoring works. I've found in the past that it breaks a lot of ports or becomes unstable for things other than the basic scenario. 

Whats worse, though, is you're telling the tooling to contact a new random port locally when it actually wants to access otherhost:otherport.  For example, if you want to spy on your VM's EAP instance, and that instance is at 192.168.1.101:8080,   it will set up a new port on localhost (localhost:27511) and any requests sent to localhost:27511 will be forwarded over to 192.168.1.101:8080.   

So now you need to get your tooling to communicate on localhost:27511.  How would we even do that with an openshift server? Often times these servers are using the hostname from the server editor. Or they're making requests via the 'oc' command.  

Trying to get this stuff through port monitoring is a huge pain. And it's really not worth the hassle. It could necessitate a complete redesign of a lot of core assumptions.

PR #1692, opened 2018-03-08
DebugLaunchConfigs NPE on workspace shutdown

robstryker:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
@@ -289,7 +289,10 @@ public class OpenShiftLaunchController extends AbstractSubsystemController
 
 			@Override
 			public void onDebugChange(DebugContext context, IProgressMonitor monitor) throws CoreException {
-				DebugLaunchConfigs.get().terminateRemoteDebugger(getServer());
+				DebugLaunchConfigs configs = DebugLaunchConfigs.get();
+				if( configs != null ) {
+					configs.terminateRemoteDebugger(getServer());
+				}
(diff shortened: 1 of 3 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftLaunchController.java
```diff
@@ -389,6 +392,9 @@ private ILaunch attachRemoteDebugger(IServer server, int localDebugPort, IProgre
 		monitor.subTask("Attaching remote debugger...");
 		ILaunch ret = null;
 		DebugLaunchConfigs launchConfigs = DebugLaunchConfigs.get();
+		if( launchConfigs == null ) {
+			throw toCoreException(NLS.bind("Could not modify launch config for server {0}", server.getName()));
```
dgolovin: It should be `Could not get launch config for server {0} to attach remote debugger`, should it not? This way it would be in some way related to action 'Attaching remote debugger' and it would by obvious why it cannot be attached.

adietish:
@robstryker is there any usecase to reproduce the NPE?

adietish:
@robstryker Is there any way to reproduce this?

jeffmaury:
I got the NPE when starting OpenShift server adapter in debug mode and then stopping Eclipse. But as this is a race condition between plugins this is not guaranted to always occur

adietish:
@jeffmaury in other words you'd +1 this fix since it is addressing the NPE when it happens? 
Fixing "why it happens" is not possible to my knowledge since we cannot control how the OSGI runtime shuts down bundles.

jeffmaury:
Yes +1 for the functionnality, agrees with @dgolovin for code review

PR #1704, opened 2018-03-20
jbide-25590: connection is not refreshed if you modify its url

bdshadow:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
+++ b/plugins/org.jboss.tools.openshift.common.core/src/org/jboss/tools/openshift/common/core/connection/ConnectionsRegistry.java
@@ -247,25 +247,7 @@ public class ConnectionsRegistry {
 	}
 
 	public void update(IConnection currentConnection, IConnection updatedConnection) {
-		ConnectionURL updatedConnectionUrl = null;
-		try {
-			updatedConnectionUrl = ConnectionURL.forConnection(updatedConnection);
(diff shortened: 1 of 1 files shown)


adietish:
works for me, +1
Could we please also have a test (that could at least verify that firing happens)?

bdshadow:
@adietish sure, working on it
@jeffmaury  maybe we handle this case, but how should it expand (i mean, to the project layer? or to the services? if to the services layer then should we do this for all projects?)

jeffmaury:
I would expand to the project level if the previous one was expanded

bdshadow:
@jeffmaury ok, i'll take a look if it is not difficult to implement. If so, i'll include it into this pr

bdshadow:
@jeffmaury i can't find how to understand that the connection(or project, or service) is expanded. Do you have an idea how to check it?

adietish:
@bdshadow I suggest that we merge this for now and provide the enhancement in a further jira.

adietish:
@bdshadow you can tell the TreeViewer to expand via TreeViewer#expandToLevel

bdshadow:
@adietish :( i was making a test. Still wanted to merge it
Also the question is not HOW to expand the tree. Look at [Jeff's comment above](https://github.com/jbosstools/jbosstools-openshift/pull/1704#issuecomment-374878221): "... if the previous one was expanded". How to understand that the previous one was expanded? In general - how to detect that TreeViewer is expanded to some level?

PR #1712, opened 2018-03-29
 [JBIDE-22245] keyvaluewiz: corrected all "Label"->"Key"

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/common/ResourceLabelsPage.java
```diff
@@ -200,7 +229,7 @@ protected TableViewer createTable(Composite tableContainer) {
 					public String getValue(Label label) {
 						return label.getName();
 					}
-				})).name("Name").align(SWT.LEFT).weight(2).minWidth(100).buildColumn()
+				})).name("Key").align(SWT.LEFT).weight(2).minWidth(100).buildColumn()
```
bdshadow: use `LABEL_KEY` constant here, added above

adietish:
@bdshadow corrected both, thx for pointing out.

PR #1720, opened 2018-04-06
Change default minishift_home in new cdk / ms server fragment

robstryker:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/cdk/server/ui/internal/CDK32ServerWizardFragment.java
```diff
@@ -93,6 +93,14 @@ public void modifyText(ModifyEvent e) {
 		}; 
 		msHomeText.addModifyListener(msHomeModListener);
 	}	
+	
+	private String getDefaultMinishiftHome() {
+		String msHome = System.getenv(CDK32Server.ENV_MINISHIFT_HOME);
```
dgolovin: According to documentation for System.getEnv(String) :
```
The returned map will never contain null keys or values.
```
that means check for null is not needed.

PR #1734, opened 2018-04-27
jbide-25909: connection is not refreshed after editing credentials

bdshadow:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/EditConnectionHandler.java
```diff
@@ -28,11 +30,16 @@
 
 	@Override
 	public Object execute(ExecutionEvent event) throws ExecutionException {
-		IConnection connection = UIUtils.getFirstElement(HandlerUtil.getCurrentSelection(event), IConnection.class);
-		return openConnectionWizard(connection, event);
+	    ConnectionWrapper connectionWrapper = UIUtils.getFirstElement(HandlerUtil.getCurrentSelection(event), ConnectionWrapper.class);
```
adietish: We here depend on a v3 specific class ConnectionWrapper. This will thus only work in v3 (we dont have v2 any more anyhow). Moving this from openshift.common.ui to openshift.ui  makes perfect sense. But I think that we therefore should also correct the activation expression from testing the common IConnection to the v3 ConnectionWrapper (openshift.ui/plugin.xml, l1485):
```
<adapt type="org.jboss.tools.openshift.common.core.connection.IConnection" />
```

adietish:
other than the suggested change in the prior review the PR fixes the issue for me.

bdshadow:
@adietish fixed

adietish:
@bdshadow jenkins currently fails with 
```
[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:1.1.0:compare-version-with-baselines (default) on project org.jboss.tools.openshift.client: Only qualifier changed for (org.jboss.tools.openshift.client/3.4.3.Final-v20180514-1327). Expected to have bigger x.y.z than what is available in baseline (3.4.3.Final-v20180412-2005-B2685) -> [Help 1]
```

bdshadow:
rebased

PR #1738, opened 2018-05-14
[JBIDE-25967] - Properties view: Add OC version for OpenShift connection

jeffmaury:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/ConnectionPropertySource.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/ConnectionPropertySource.java
@@ -14,6 +14,7 @@ import java.util.Set;
 import java.util.TreeSet;
 
 import org.apache.commons.lang.StringUtils;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.ui.views.properties.IPropertyDescriptor;
 import org.eclipse.ui.views.properties.IPropertySource;
@@ -24,6 +25,8 @@ import org.jboss.tools.openshift.common.core.connection.IConnectionsRegistryList
 import org.jboss.tools.openshift.core.ICommonAttributes;
 import org.jboss.tools.openshift.core.connection.IOpenShiftConnection;
 import org.jboss.tools.openshift.internal.common.ui.utils.OpenShiftUIUtils;
+import org.jboss.tools.openshift.internal.core.preferences.OCBinary;
+import org.jboss.tools.openshift.internal.core.preferences.OCBinaryVersionValidator;
 
 public class ConnectionPropertySource implements IPropertySource {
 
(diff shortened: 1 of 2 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/property/ConnectionPropertySource.java
```diff
@@ -122,6 +127,11 @@ public Object getPropertyValue(Object id) {
 			if (KUBERNETES_MASTER_VERSION.equals(id)) {
 				return openshiftConnection.getKubernetesMasterVersion();
 			}
+			if (OC_VERSION.equals(id)) {
+				String path = OCBinary.getInstance().getLocation(openshiftConnection);
+				OCBinaryVersionValidator validator = new OCBinaryVersionValidator(path);
+				return validator.getVersion(new NullProgressMonitor()).toString();
```
adietish: I'd think that we should not run #getVersion in the UI thread since it launches a process, a heavy operation.

adietish: @jeffmaury a bit hacky but could work: when asked for the value for oc in #getPropertyValue return an empty string (or null). Create a subclass of UneditablePropertyDescriptor that launches a job (to determine the oc version) if it gets an empty/null string as value (in CellEditor#doSetValue) that then calls super.doSetValue to update the editor value?
WDYT?

jeffmaury: This is what I tried to do but it does not work because the cell editor is created only when the user clicks on the cell in the UI and setting the value does not seem to work because the text field is readonly

adietish: IC, what about a job that is launched when the value is asked for the 1st time. it would fetch the version and cause the table to be refreshed, querying the model again which would now return the evaluated value (returning empty string/null when launching the job)?
I also see a ConnectionPropertySource#setPropertyValue that the job could use to update the value?

jeffmaury: There is no way to reflect the change to the UI. The setPropertyValue is called by the UI layer when the user changes the value to report the change to the model.

adietish: I still believe that running a process in the UI thread is something a plugin which behaves like a good Eclipse citizen should avoid whenever possible. Among the worst cases one could imagine could be where oc is on an NFS share (with great latency) which then would cause the UI to freeze when the properties for the connection are displayed. 
What about displaying instead whether the connection overrides the global oc preference and display the path to oc (which matches the wizard when editing a connection)?

jeffmaury:
testPR

PR #1744, opened 2018-05-24
jbide-25756: fix: connection displays project that was deleted

bdshadow:


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/NewProjectHandler.java
```diff
@@ -55,7 +55,7 @@ public static void openNewProjectDialog(final Connection connection, final Shell
 		Display.getDefault().asyncExec(new Runnable() {
 			@Override
 			public void run() {
-				List<IProject> oldProjects = connection.getResources(ResourceKind.PROJECT);
+				List<IProject> oldProjects = connection.getActiveProjects();
```
adietish: this new method is a bit unexpected but works fine, no doubt. 
Just a modest suggestion, but that's not required: I probably would have added a #getResources method which allows filtering which I think would be more in line with the existing API

bdshadow:
@adietish i changed it. Take a look please

adietish:
testPR

PR #1745, opened 2018-05-25
[JBIDE-25905] route page: reuse existing table item instances

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/PodPortValidator.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/PodPortValidator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015 Red Hat, Inc.
+ * Copyright (c) 2016 Red Hat, Inc.
  * Distributed under license by Red Hat, Inc. All rights reserved.
  * This program is made available under the terms of the
  * Eclipse Public License v1.0 which accompanies this distribution,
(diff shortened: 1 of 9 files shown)


adietish:
@jeffmaury @bdshadow this jira was caused by weird bugs in jface databinding: 
* the implementation was using ObservableMapLabelProvider to label data entries. This label provider works well as long as the input dataset stays the same. As soon as it's replaced (as it happens when one is pressing "Reset" in the UI), it wont update its cache (attributeMaps) and thus wont find the data that it wants to create labels for (empty labels in the screenshots attached to jira).  So I removed ObservableMapLabelProvider and replaced it with CellLabelProviders (which is what we use in other places in the tooling). The labeling then works fine but unveils another databdinging bug:
* Changes in boolean properties for an entry is not detected as a change and thus wont trigger a refresh in the tableviewer. I worked around this by listening to changes to the routing port property, manually triggering a refresh of the tableviewer.

adietish:
@jeffmaury @bdshadow please review

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/PodPortValidator.java
```diff
@@ -16,73 +16,93 @@
...
+			status = validateValue(port);
+			if (status.isOK()) {
+				status = validateUniqueness(port);
+			}
+		}
+		return status;
+	}
+
+	private IStatus validateUniqueness(final String port) {
+		for (IServicePort existing : ports) {
```
bdshadow: can be simpler:`if ports.contains(existing.getTargetPort())`

adietish: @bdshadow this unfortunately wont work. ports contains IServicePort objects. What we get as parameter is a string. We need to check if among the service ports objects, there's already one whose target port matches the given (String) parameter.

PR #1746, opened 2018-05-30
Revert "[JBIDE-25905] route page: reuse existing table item instances"

adietish:
Reverts jbosstools/jbosstools-openshift#1745


adietish:
closed since master is unfrozen now.

PR #1748, opened 2018-06-04
jbide-24979: Scale up/down/to: NPE if you have only rc, no dc

bdshadow:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/ScaleDeploymentHandler.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/ScaleDeploymentHandler.java
@@ -49,6 +49,7 @@ import org.jboss.tools.openshift.internal.common.ui.utils.DisposeUtils;
 import org.jboss.tools.openshift.internal.common.ui.utils.UIUtils;
 import org.jboss.tools.openshift.internal.core.util.ResourceUtils;
 import org.jboss.tools.openshift.internal.ui.OpenShiftUIActivator;
+import org.jboss.tools.openshift.internal.ui.models.IReplicationControllerWrapper;
 import org.jboss.tools.openshift.internal.ui.models.IResourceWrapper;
 import org.jboss.tools.openshift.internal.ui.models.IServiceWrapper;
 import org.jboss.tools.openshift.internal.ui.utils.ResourceWrapperUtils;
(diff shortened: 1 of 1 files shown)


adietish:
testPR

bdshadow:
@adietish yes, i had the same. And did nothing with it as it is really not connected twith this jira

adietish:
@bdshadow I agree on the fact that having the pod(s) show up twice needs a new jira, but I'd think that doing nothing (with no feedback) for some pod (that are shown) should be adressed in this jira? Wrong?

bdshadow:
@adietish i agree that it's not ok to do nothing without feedback.
The reason, why nothing happens, is because we can't find rc for such pods, as we try to find services for them first and then replication controller. And this kind of pods don't have service as a parent, because their parent (in terms of `IResourceWrapper`s) is project.
The reason why we show it like this is in this PR: https://github.com/jbosstools/jbosstools-openshift/pull/1709/files#diff-fe25f5ebed5fcd00f5f86c6e04e1aba0R188. So my question is - is it a correct representation of the hierarchy of the resources? If yes - i'll try to retrieve the rc somehow in this jira. If no - it's a matter of another jira

adietish:
@bdshadow thx for the explanations. The pods "belong" to the rc, which is displayed. What I dont understand is that this relation seems to be detected (pods are displayed as children to the rc), since I can scale the pods displayed as children to the rc. But I can't when they're displayed as children to the project. Looks like the visual representation defines if a pod can scale or not. 
On the other hand I get that doing resource queries (to overcome the missing relation to the rc in IResourceWrapper for the erroneously displayed pods) isn't possible since we're in UI thread here. 
I guess that we can leave this "bug" alone since we want to fix the double display bug and keep the pods that are displayed as children to the rc only, making this "bug" disappear. It could hit us later though. 
Agree?

bdshadow:
@adietish agree :) 
PS
You are saying that 

> relation seems to be detected 

But it's not detected. We put it like this ourselves: https://github.com/jbosstools/jbosstools-openshift/pull/1709/files#diff-fe25f5ebed5fcd00f5f86c6e04e1aba0R192 and line 187 too.

adietish:
@bdshadow so pods could end up listed below an rc that doesn't control them? sounds bad and worth a jira?

bdshadow:
@adietish yes: https://issues.jboss.org/browse/JBIDE-26087

PR #1751, opened 2018-06-05
[JBIDE-26079] explorer: edit projects when selecting a project

adietish:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
after the change, Edit > Projects... is available when a project is selected in the explorer:
image

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizard.java
```diff
@@ -30,10 +31,13 @@ public ManageProjectsWizard(Connection connection) {
 		this(null, connection);
 	}
 
+	public ManageProjectsWizard(IProject project) {
```
bdshadow: what's the point of adding this constructor?

bdshadow:
it works for me, as described in the jira. However, when i select a project and then edit -> Projects, Openshift Projects wizard is not smth that i expect.
I mean it's fine for connections level, but not for projects. When do Edit -> Project for a project, i want to change its name or display name or smth like this. But i don't want to remove it (i have a separate button for this) of create a new one. 
wdyt?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizard.java
adietish: @bdshadow if a project is selected, the wizard will derive the connection from the project. In the other case a connection is selected, so there's no need to derive the connection.

bdshadow: yes, i see it. But i haven't found any usage of this constructor. I even tried to comment this lines - works fine. So it looks to me like no need in new lines of code

bdshadow:
@adietish i just think it's logically correct to select connection, when you want to manage projects of this connection.
About renaming projects - ok, but we can edit a description and a display name.
About moving deletes to the "Edit" menu - i don't think it's a good idea. Remove != Edit.
Anyway if you consider it more handy, i'm ok about it

adietish:
@bdshadow I believe that this is still handy.
@jeffmaury @jkopriva your vote?

adietish:
@jeffmaury +1'ed the change
@jkopriva your opinion?

jkopriva:
@adietish +1, works for me

adietish:
@jkopriva thx!
@bdshadow dare to +1, please?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/project/ManageProjectsWizard.java
adietish: @bdshadow ok, removed it

PR #1767, opened 2018-07-03
jbide-26084: Explorer: pods are shown twice if they're controlled by an

bdshadow:
rc (no dc, no service)


bdshadow:
@adietish take a look, please! In my opinion it's important to be merged before the code freeze

adietish:
works for me too for normal template base apps while I have a case (msa app "hello") where the pod still shows up twice, but this is most likely caused by erroneous labels.
image
Filing to a new JIRA, +1 for this change.

bdshadow:
the main problem is that we don't know the main characteristic to use to
determine that pod has no "parents"

PR #1772, opened 2018-07-16
 [JBIDE-25700] added warning when using oc < 3.11 on linux

adietish:
# Pull Request Checklist
## General
* Is this a blocking issue or new feature? If yes, QE needs to +1 this PR

## Code
* Are method-/class-/variable-names meaningful?
* Are methods concise, not too long?
* Are catch blocks catching precise Exceptions only (no catch all)?

## Testing
* Are there unit-tests?
* Are there integration tests (or at least a jira to tackle these)?
* Is the non-happy path working, too?
* Are other parts that use the same component still working fine?

## Function
* Does it work?

PR #1774, opened 2018-07-17
[JBIDE-25700] added warning when using oc < 3.11 on linux

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@odockal @jkopriva please review. You can test this PR instead of https://github.com/jbosstools/jbosstools-openshift/pull/1772 since it includes the latter.

When testing, please now also test
* Port forwarding
* Pod Log

Both should complain if there's no oc binary defined in the preferences or preferences point to a bogus oc binary (which is not executable or is not an oc binary)

odockal:
@adietish 
Preferences validation of oc: 
Test case 1:
1. open the preferences, switch to OpenShift 3 item:
2. oc is not set
3. ASSERT: got warning saying "Could not determine ..."
4. ASSERT: can switch to another item
5. ASSERT: Apply and close button is pressed, preferences closed
Actual result: One cannot apply and close after visited Openshift 3 page without oc being set, even when switching to another page. 
Workaround: Close preferences, open and do not switch to Openshift 3 page, then Apply and close works again.

Test case 2.
1. open the preferences, switch to OpenShift 3 item:
2. choose oc version < 3.11
3. ASSERT: warning message that oc is under 3.11, apply and close can be pressed (linux only)

Test Case 3:
1. open the preferences, switch to OpenShift 3 item:
2. choose oc version < 3.11
3. ASSERT: warning message that oc is under 3.11, apply and close can be pressed (linux only)
4. Preferences opened again, Openshift 3 item is choosen
5. ASSERT: oc can be erased, apply and close pressed, preferences closed
Actual result: Cannot press apply and close even it is enabled (should be enabled, but must allow user to be pressed)
Workaround: Erase location to oc, press apply and Cancel dialog

I did not tested port forwarding, though I tried pod log, I can get log (opened in console view using template eap71-basic-s2i) but I also got some wrong coding there:
screenshot_20180718_181940

adietish:
@odockal I fixed the issues with "Apply and Close" that you found.
The encoding is a different issue. Either this is a regression of JBIDE-21869 or we need a new issue for this.

adietish:
@odockal 
As you pointed out in the earlier PR it is important to fix this in the following usecases (which are all covered by the different issues linked to JBIDE-25700):
Given that you don't have an oc configured (and/or on linux, an oc < 3.11)
* create a new app from file menu (or central): should not error, only warning displayed
* create a new connection: should not error, only warning displayed
* create a new server adapter: error displayed (since server adapter requires a valid oc)
* launch pod log: error displayed only if oc is not set or oc is not recognized
* launch port forwarding: error displayed only if oc is not set or oc is not recognized

odockal:
@adietish oh, alright. So I manually tested what you had requested.
Assuming that you don't have an oc configured, unrecognized oc, oc defined with a path that contains " " (on non-windows) or (on linux) an oc < 3.11:

- create a new app from file menu (or central): should not error, only warning displayed - _Seems works this way, except that according to your assumptions I had no oc configured, which was only warning showed, still OK._
- create a new connection: should not error, only warning displayed - _Seems works this way, same situation as above -> oc is not configured, still OK._
- create a new server adapter: error displayed (since server adapter requires a valid oc) - Server adapter can be created without any problem, while starting Problem Occured dialog is shown making complaints about oc cli tools, after accepting the error the adapter is still started, no further warning/error is shown.

Assuming that you dont have oc configured or configured an unrecognized oc:

- launch pod log: error displayed only if oc is not set or oc is not recognized - OpenShift client Error dialog pops up
- launch port forwarding: error displayed only if oc is not set or oc is not recognized - OpenShift client Error dialog pops up

When setting oc (v3.9.31) as my oc executable, all of above works with warnings but without any errors.

adietish:
@odockal: @jkopriva pointed out, and he's perfectly right, that with the errors in the server adapter one currently couldn't create a server adapter on linux, since there's no oc 3.11 available yet. So we decided that we's have only warnings in the server adapter, too.
So if I read you right, this PR works fine for you? +1?

odockal:
@adietish Check also this one out: https://issues.jboss.org/browse/JBIDE-26223

adietish:
@odockal I had this issue initially, a problem of initial wait states. It's not related to this PR though. We should fist focus on this PR have it done.

adietish:
rebased and pushed to get the latest changes in master.

odockal:
@adietish Thanks for rebasing, let wait for PR check to finish. Otherwise, +1 from my side. I would let @jkopriva to have last words as he was running itests suite for this PR.

jkopriva:
@adietish @odockal I have just found only one issue: https://issues.jboss.org/browse/JBIDE-26225, not sure, if this is related to changes introduced in this PR.

I have went through all usecases mentioned in this and previous PR and it looks good.

I have also run ITests suite against this PR(before rebase) and after changes to creation server adapter it was OK.

So from my side also +1.

adietish:
@jkopriva thx for spotting. it might be related, but I'd fix this is a separate PR to not further bloath this already huge PR. Agree?

jkopriva:
@adietish yes, that would be the best option :)

PR #1785, opened 2018-07-31
jbide-26175: cannot create server adapter for eap71-basic-s2i template

bdshadow:

A very easy and little fix, but needs to be tested a lot, as can have some unpredictable repercussions.


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapDeployPathController.java
```diff
@@ -36,7 +36,7 @@ protected String getOutputNameFromSettings(IServerAttributes server, IModule mod
 			String ret = super.getOutputNameFromSettings(server, module);
 			if (ret == null && module.equals(findProjectModule(server))) {
 				String suffix = ServerModelUtilities.getDefaultSuffixForModule(module);
-				ret = "ROOT" + suffix;
+				ret = module.getName() + suffix;
```
jeffmaury: That would probably fix for EAP 7.1 but for 7.0 as the OpenShift profile is probably not set you will not get the proper module name. Can you test with EAP 7.0 ?

bdshadow: @jeffmaury you're right. It fails. But i think it must be part of https://issues.jboss.org/browse/JBIDE-22138?focusedCommentId=13608828&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13608828. @adietish wdyt?

adietish: @bdshadow Yes, since the reason for the module name not being handled correctly lies in JBIDE-22138 aka upstream. As long as upstream isn't fixed the decision whether we want to apply this fix boils down to whether we want eap-70 or eap-71 templates to fail. I'd opt for sticking to the status quo aka not applying this fix.
@jeffmaury @bdshadow thoughts?

jeffmaury:
As J2EEDeployableFactory is a singleton can't we call the clearCache from our stuff before we generate the module name ?

adietish:
@jeffmaury I get the idea and yes, I agree: even though very dirty we could try to work around the bug in JBIDE-22138 and manually clear the module factory cache(s).

jkopriva:
I am also against applying this patch, we need to have more complex solution to have both(eap70 and eap71) templates working.
As @jeffmaury mentioned, with this patch eap71 is working, but eap70 template does not work.

bdshadow:
@adietish @jeffmaury @jkopriva "openshift" profile is now activated when an openshift server adapter is created. Anyway it doesn't solve our problems. Please, take a look and test it. WDYT?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapDeployPathController.java
```diff
@@ -33,16 +40,40 @@ protected ModuleDeploymentPrefsUtil createModuleDeploymentPrefsUtil() {
...
+		        Method clearCacheMethod = J2EEDeployableFactory.class.getDeclaredMethod("clearCache", IProject.class);
+		        clearCacheMethod.setAccessible(true);
+		        clearCacheMethod.invoke(JEEDeployableFactory.jeeInstance(), getEclipseProject(server));
+            } catch (IllegalAccessException
+                    | IllegalArgumentException
+                    | InvocationTargetException
+                    | NoSuchMethodException
+                    | SecurityException e) {
+                // TODO Auto-generated catch block
+                e.printStackTrace();
```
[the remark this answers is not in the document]
adietish: @bdshadow dare to replace this stacktrace printing by logging?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapDeployPathController.java
```diff
@@ -33,16 +40,40 @@ protected ModuleDeploymentPrefsUtil createModuleDeploymentPrefsUtil() {
...
+		    try {
+		        Method clearCacheMethod = J2EEDeployableFactory.class.getDeclaredMethod("clearCache", IProject.class);
+		        clearCacheMethod.setAccessible(true);
+		        clearCacheMethod.invoke(JEEDeployableFactory.jeeInstance(), getEclipseProject(server));
+            } catch (IllegalAccessException
+                    | IllegalArgumentException
+                    | InvocationTargetException
+                    | NoSuchMethodException
+                    | SecurityException e) {
+                // TODO Auto-generated catch block
```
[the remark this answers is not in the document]
adietish: @bdshadow maybe mark this clearly as a workaround for JBIDE-22138 (with link and explanation) in javadoc?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapDeployPathController.java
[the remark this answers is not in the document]
adietish: @bdshadow dare to please replace this stacktrace printing by logging?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapDeployPathController.java
```diff
@@ -33,16 +40,40 @@ protected ModuleDeploymentPrefsUtil createModuleDeploymentPrefsUtil() {
...
 			}
 			return ret;
 		}
 
 		protected IModule findProjectModule(IServerAttributes server) {
-			return OpenShiftServerUtils.findProjectModule(OpenShiftServerUtils.getDeployProject(server));
+			return OpenShiftServerUtils.findProjectModule(getEclipseProject(server));
+		}
+		
+		private IProject getEclipseProject(IServerAttributes server) {
```
adietish: any particular reason for extracting the call to the util method to it's own *private* method?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPage.java
```diff
@@ -1088,8 +1101,27 @@ public boolean isPageComplete() {
 
 	public IServer saveServer(IProgressMonitor monitor) throws CoreException {
 		model.updateServer();
+		setOpenshiftMavenProfileActive(model.getDeployProject());
 		return model.saveServer(monitor);
 	}
+	
+	@SuppressWarnings("restriction")
+    private void setOpenshiftMavenProfileActive(IProject newDeployProject)  throws CoreException  {
```
adietish: maybe we could make this method even more useful by refactoring it to something like ```activateMavenProfile(String profilename, IProject project)```?

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPage.java
```diff
@@ -1088,8 +1101,27 @@ public boolean isPageComplete() {
...
+	    IFile pom = newDeployProject.getFile(IMavenConstants.POM_FILE_NAME);
```
adietish: the file may only be retrieved if the project has the maven nature, so the call could be easily moved to l1112 making it more logical, I think.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPage.java
```diff
@@ -1088,8 +1101,27 @@ public boolean isPageComplete() {
...
+            IMavenProjectFacade facade = MavenPlugin.getMavenProjectRegistry().create(pom, true, null);
+            List<ProfileData> profiles = profileManager.getProfileDatas(facade, null);
+            List<String> activeProfiles = profiles.stream()
+                    .filter(p -> ProfileState.Active.equals(p.getActivationState()))
+                    .map(p -> p.getId())
+                    .collect(Collectors.toList());
+            if (profiles.stream().anyMatch(p -> OPENSHIFT_MAVEN_PROFILE.equals(p.getId())) && !activeProfiles.contains(OPENSHIFT_MAVEN_PROFILE)) {
+                activeProfiles.add(OPENSHIFT_MAVEN_PROFILE);
+                profileManager.updateActiveProfiles(facade, activeProfiles, false, true, new NullProgressMonitor());
+            }
```
adietish: l1111-1122 could be refactored to a set of 3 methods which then would be reusable - if then moved to a common bundle :
* ProfileData getActiveProfiles(IProject project)
* boolean isActiveProfile(String profileId, IProject project) - optional, could be included in setActiveProfile
* void setActiveProfile(String profileId)

Looks like a benefit if done so, agree?

nickboldt:
testPR

adietish:
@bdshadow You can rebase this PR in order to fix the jenkins errors. I bumped openshift (in https://github.com/jbosstools/jbosstools-openshift/pull/1788) in order to fix the baseline check errors that caused those.

jkopriva:
@bdshadow I have retested it with last commit and I was able to create server adapter for eap70 and eap71 template/

bdshadow:
@jkopriva thank you
@adietish i also tested it against https://github.com/redhat-helloworld-msa/helloworld-msa/blob/master/hello.adoc (second cmd approach) and it works fine.

Now i'm going to refactor my code according to the remarks and test it against java 9.

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPage.java
```diff
@@ -1088,8 +1101,27 @@ public boolean isPageComplete() {
...
 	public IServer saveServer(IProgressMonitor monitor) throws CoreException {
 		model.updateServer();
+		setOpenshiftMavenProfileActive(model.getDeployProject());
 		return model.saveServer(monitor);
 	}
+	
+	@SuppressWarnings("restriction")
+    private void setOpenshiftMavenProfileActive(IProject newDeployProject)  throws CoreException  {
+	    IFile pom = newDeployProject.getFile(IMavenConstants.POM_FILE_NAME);
+	    if (newDeployProject.hasNature(IMavenConstants.NATURE_ID) && pom != null) {
```
adietish: checking IFile "pom" against being null isn't enough I believe. IContainer#getFile always returns an instance, regardless whether it exists or not. I think that we should also check ```pom.isAccessible()```

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPage.java
```diff
@@ -1088,8 +1101,27 @@ public boolean isPageComplete() {
...
+	@SuppressWarnings("restriction")
+    private void setOpenshiftMavenProfileActive(IProject newDeployProject)  throws CoreException  {
+	    IFile pom = newDeployProject.getFile(IMavenConstants.POM_FILE_NAME);
+	    if (newDeployProject.hasNature(IMavenConstants.NATURE_ID) && pom != null) {                   
+            final IProfileManager profileManager = MavenProfilesCoreActivator.getDefault().getProfileManager();
+            IMavenProjectFacade facade = MavenPlugin.getMavenProjectRegistry().create(pom, true, new NullProgressMonitor());
+            List<ProfileData> profiles = profileManager.getProfileDatas(facade, new NullProgressMonitor());
+            List<String> activeProfiles = profiles.stream()
+                    .filter(p -> ProfileState.Active.equals(p.getActivationState()))
+                    .map(p -> p.getId())
```
adietish: this lambda can be replaced by a method reference for improved readability: ```ProfileData::getId```

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ServerSettingsWizardPage.java
```diff
@@ -1088,8 +1101,27 @@ public boolean isPageComplete() {
...
+	
+	@SuppressWarnings("restriction")
+    private void setOpenshiftMavenProfileActive(IProject newDeployProject)  throws CoreException  {
+	    IFile pom = newDeployProject.getFile(IMavenConstants.POM_FILE_NAME);
+	    if (newDeployProject.hasNature(IMavenConstants.NATURE_ID) && pom != null) {                   
+            final IProfileManager profileManager = MavenProfilesCoreActivator.getDefault().getProfileManager();
+            IMavenProjectFacade facade = MavenPlugin.getMavenProjectRegistry().create(pom, true, new NullProgressMonitor());
+            List<ProfileData> profiles = profileManager.getProfileDatas(facade, new NullProgressMonitor());
+            List<String> activeProfiles = profiles.stream()
+                    .filter(p -> ProfileState.Active.equals(p.getActivationState()))
```
adietish: writing down what we discussed/found out:
with a ~/.m2/settings.xml file, that also has profiles, filtering only on activation states this will have all those activated:
image
One needs to also filter autoactive profiles:
```
.filter(p -> ProfileState.Active.equals(p.getActivationState())
                        && !p.isAutoActive())
```
This will produce the correct results:
image
image

adietish:
Another thing that we should do: 
* deactive the openshift profile if one stops (or deletes the server adapter)
* activate the openshift profile if one (re-)starts the adapter

I thus think that activating the profile should be done where the adapter gets started, not upon creation like it is now

adietish:
extracted the maven profile to it's own class MavenProfile (which can be activate and eventually deactivated) in org.jboss.tools.openshift.core.
Further work required to have it activated upon adapter start and deactivated upon adapter stop.

adietish:
@bdshadow rebased and added test for MaveProfile

bdshadow:
Tested with java 9. Works fine

adietish:
@bdshadow When trying the eap70 basic template, I have the project switched to the correct "openshift" profile but still the "kitchensink.war" artifact deployed. Weird enough, I see the correct ```clearCache(IProject)``` method being called by our workaround.

bdshadow:
@adietish actually i removed the usage of property name. I debugged it for eap70 and eap71 and it was returning the same both: `module.getName()` and property. Moreover, if null is returned, you can find in the parent class that this logic with the property retrieval is run

I don't know why it didn't work for you with eap70. I've tested it lots of times already, and once i also had such a problem, and i wrote to you about it. The everything was fine. I have know idea why this happens sometimes

adietish:
we agreed that deploying using the deploy name, rather the module name is the right way to do it.
But then we face the next issue:
When switching the deploy name, the name of the war changes. A new (exploded) war is created in the temporary server-adapter directory in Eclipse, but the old (exploded) war is not removed. This causes both wars to be deployed

steps:
1) have app, workspace project and server adapter in place
2) make sure that the server-adapter temporary directory in Eclipse only has 1 exploded war (either ROOT.war or jboss-kitchensink.war - delete the other one) - temporary server-adapter directory in Eclipse is at ```<workspace>/.metadata/.plugins/org.jboss.ide.eclipse.as.core/<adapter-name>/deploy```
3) switch the maven profile of the workspace project (if it has the "openshift" profile, uncheck it. If it has it, check it)
4) Result: a new deployed war directory is created in the temporary server-adapter directory. The existing one stays in place, it is not removed. Therefore rsync will copy both (exploded) wars to OpenShift

adietish:
to solve the above one could simply clean the local "deploy" directory (local sever-adapter storage) before it gets rsynced up to the OpenShift pod.
The problem with that is that this wont remove the existing deployment in the (wildfly on the) pod. Resulting is an additional deployment with the new war name that conflicts with the existing one (that was not removed).
To remove that one, we'd have to kill the <war-name>.deployed marker (see https://docs.jboss.org/author/display/WFLY10/Application+deployment#Applicationdeployment-MarkerFiles) which we currently dont have locally (another bug: we do rsync down pod->"deploy" directory, but the marker file is not present locally.

bdshadow:
by now it's better to wait for @robstryker as we can be missing smth and solution can be easier (or in the contrary we don't see smth complex). I discussed it with @adietish yesterday. It seems to me that he agreed.

adietish:
@bdshadow I added an additional commit which makes sure that modules get notified of the cleared module cache. Things now work for me:

* publishing to an eap71-basic-s21 app -> publishes as kitchensink.war (as seen in the console)
* publishing to an eap70-basic-s21 app -> publishes as ROOT.war (as seen in the console)

@bdshadow @jkopriva could you please test this PR and give it a +1/-1?

PR #1803, opened 2018-09-25
jbide-26159: Server adapter: port forwarding isn't stopped if you delete a pod

bdshadow:

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
@@ -24,7 +24,12 @@ import java.util.Set;
 
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.jdt.launching.SocketUtil;
+import org.jboss.tools.openshift.common.core.connection.ConnectionsRegistryAdapter;
+import org.jboss.tools.openshift.common.core.connection.ConnectionsRegistrySingleton;
+import org.jboss.tools.openshift.common.core.connection.IConnection;
+import org.jboss.tools.openshift.common.core.connection.IConnectionsRegistryListener;
 import org.jboss.tools.openshift.core.connection.ConnectionsRegistryUtil;
+import org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator;
 import org.jboss.tools.openshift.internal.core.ocbinary.OCBinaryOperation;
 
 import com.openshift.restclient.capability.CapabilityVisitor;
@@ -40,6 +45,7 @@ public class PortForwardingUtils {
 
 	/** Internal registry of {@link IPod}'s port-forwarding. */
 	private static final Map<IPod, IPortForwardable> REGISTRY = new HashMap<>();
+	private static final Map<IPod, IConnectionsRegistryListener> POD_LISTENERS = new HashMap<>();
 
 	private PortForwardingUtils() {
 	}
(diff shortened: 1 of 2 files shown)


bdshadow:
@jeffmaury 

-  it's not about debugging, it's more about port forwarding - look at the issue.
-  yes, that's what Andre said about, working on it. It'll be triggered on a message from the server (through the Watcher)

jeffmaury:
OK, I got confused by the title of this PR 👎

bdshadow:
@jeffmaury it has been changed recently by @adietish according to the description of the jira, as the title really didn't reflect the main idea

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
```diff
@@ -151,9 +157,23 @@ protected void runOCBinary() {
...
 		}, null);
-		if (portForwarding != null) {
-			REGISTRY.put(pod, portForwarding);
-		}
+        if (portForwarding != null) {
+            REGISTRY.put(pod, portForwarding);
+            IConnectionsRegistryListener podListener = new ConnectionsRegistryAdapter() {
+                @Override
+                public void connectionChanged(IConnection connection, String property, Object oldValue, Object newValue) {
+                    if (newValue == null && oldValue instanceof IPod) {
```
jeffmaury: Where do you check that the deleted pod is the one you started a port forward ?

adietish: I guess something like ```REGISTRY.containsKey(pod)``` is missing.
@bdshadow WDYT?

adietish: the following does the trick for me:
```
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/portforwarding/PortForwardingUtils.java
@@ -162,7 +162,9 @@ public class PortForwardingUtils {
             IConnectionsRegistryListener podListener = new ConnectionsRegistryAdapter() {
                 @Override
                 public void connectionChanged(IConnection connection, String property, Object oldValue, Object newValue) {
-                    if (newValue == null && oldValue instanceof IPod) {
+                    if (newValue == null
+                               && oldValue instanceof IPod
+                               && REGISTRY.containsKey(oldValue)) {
                         try {
```

steps:
1) ASSERT: have 2 applications in 2 different projects with at least 1 pod runnin
2) EXEC: start port forwarding for both pods (2nd one needs to be started by checking "[x] find free local ports"
3) EXEC: in cmd line, verify that 2 "oc port-forwarding" processes (see exact semantic above) are running
4) EXEC: kill 1 pod
5) ASSERT: the port forwarding for the pod, that you just killed, was terminated, the other port port forwarding is still running

bdshadow: @jeffmaury @adietish if you look inside `PortForwardingUtils.stopPortForwarding((IPod)oldValue, null);`, you'll find that first of all there is a check `PortForwardingUtils.isPortForwardingStarted(pod)`, which actually checks if a pod was put into `REGISTRY` before

bdshadow: @adietish checked your steps. Everything works for me without additional check ` && REGISTRY.containsKey(oldValue)`. As i said above, this check already exists inside `PortForwardingUtils.isPortForwardingStarted(pod)`

adietish: @bdshadow yes I think that you're right. Resolving. @jeffmaury dare to comment please?

jeffmaury: +1

PR #1804, opened 2018-09-25
[JBIDE-25830] try find oc in minishift home

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/cdk/server/core/adapter/CDK3Server.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/cdk/server/core/adapter/CDK3Server.java
@@ -15,6 +15,7 @@ import java.io.File;
 import org.eclipse.core.runtime.Platform;
 import org.jboss.tools.openshift.common.core.utils.StringUtils;
 import org.jboss.tools.openshift.internal.cdk.server.core.CDKConstants;
+import org.jboss.tools.openshift.internal.cdk.server.core.listeners.CDKServerUtility;
 
 public class CDK3Server extends CDKServer {
 
@@ -66,12 +67,7 @@ public class CDK3Server extends CDKServer {
 	}
 
 	public String getMinishiftHome() {
-		String home = System.getProperty("user.home");
-		String defaultMinishiftHome = new File(home, CDKConstants.CDK_RESOURCE_DOTMINISHIFT).getAbsolutePath();
-		String msHome = getServer().getAttribute(CDK3Server.MINISHIFT_HOME, defaultMinishiftHome);
-		if (StringUtils.isEmpty(msHome))
-			msHome = defaultMinishiftHome;
-		return msHome;
+		return CDKServerUtility.getMinishiftHomeOrDefault(getServer());
 	}
 
 	public static boolean matchesCDK3(String version) {
@@ -84,6 +80,9 @@ public class CDK3Server extends CDKServer {
 	 */
 	public boolean isCDKInitialized() {
 		String home = getMinishiftHome();
+		if (StringUtils.isEmpty(home)) {
+			return false;
+		}
 		File homeF = new File(home);
 		if (homeF.exists() && homeF.isDirectory()) {
 			File cdk = new File(homeF, "cdk");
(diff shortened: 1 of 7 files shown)


jeffmaury:
@adietish ok to review but can you fix the compilation failure first.

adietish:
@jeffmaury ouch, sorry, didn't see those. Fixing.

adietish:
@jeffmaury corrected. Plz have a look.

jeffmaury:
Did a test with the following conf:

- CDK was started from the command line
- No oc configured at the workspace level
- defined a new CDK server adapter and started it

Got a new OpenShift connection defined but no oc is associated

image

jeffmaury:
Did not manage to get it working. Here's my conf:

- CDK already started (cause it's a mess on my machine to start it)
- no oc configured at the workspace level
- defined a new CDK3.2+ adapter and ask for start
- a new OpenShift connection is defined but no oc is associated:

image

adietish:
@jeffmaury do you have an oc in ~/.minishift/cache/oc/... ? I'm extracting the minishift home attribute in the cdk server adapter which is pointing to ~/.minishift/ in my case. if found I then look into the subdirectories within cache/oc to find an executable: https://github.com/jbosstools/jbosstools-openshift/pull/1804/files#diff-fb7a0bf44bcc6c82199840d2b0fa9a77R162
Any idea what of the above is failing for you?

jeffmaury:
I did a bit of debugging session and found that the method findOCInMinishiftHome returns null because the server attribute ENV_VAR_MINISHIFT_HOME is null although my server definition is the following:

image

odockal:
@adietish It is not working for me as well. I have default minishift home at ~/.minishift, I DO NOT have set env. var. MINISHIFT_HOME, which should be default case and after starting cdk 3.2+ adapter OS connection does not contain ~/.minishift/cache/**/oc.

jkopriva:
@adietish I have minishift on path, so after starting CDK server adapter, OpenShift connection is created with correct oc from minishift

adietish:
@jkopriva thx for pointing out. But I don't know if this is related. If it was it's a bit unexpected.
@jkopriva @odockal @adietish  dug further into this and it turns out that minishift home attribute isn't always set. It looks expected as there are several places in the code where the adapter falls back to some default minishift home location. I unified these codes into a single utility class. This then tells me that we shouldn't merge this before GA, better postpone it. Even more since it's an enhancement.

jkopriva:
+1 for postponing after GA

adietish:
ok, postponing after 4.9.0.GA

robstryker:
> the method findOCInMinishiftHome returns null because the server attribute ENV_VAR_MINISHIFT_HOME is null 

I'm a bit confused here. First, that variable doesn't seem to exist in this PR:

[rob@rawbdorable jbosstools-openshift] (JBIDE-25830)$ grep -R "ENV_VAR_MINISHIFT_HOME" *
[rob@rawbdorable jbosstools-openshift] (JBIDE-25830)$

I did find "ENV_MINISHIFT_HOME"  (missing "_VAR_") in getDefaultMinishiftHome().  However, this method never seems to return null. In fact, it seems impossible for it to return null;

	public static String getDefaultMinishiftHome() {
		String msHome = System.getenv(CDK32Server.ENV_MINISHIFT_HOME);
		if( StringUtils.isEmpty(msHome)
				|| !new File(msHome).exists()) {
			msHome = new File(System.getProperty("user.home"), CDKConstants.CDK_RESOURCE_DOTMINISHIFT)
					.getAbsolutePath();
		}
		return msHome;
	}

Even if System.getenv(etc) returns null, logic will flow into the next line, where a default value of System.getProperty("user.home") + "/.minishift" basically gets returned. This value is returned 100% whether the folder exists or does not exist. 

In fact, the only way I can get this to fail as described is to declare a new cdk server adapter with a new clean minishift_home, receive the prompt to run setup-cdk, accept the prompt, wait for startup to complete, at which point it will attempt to make the openshift connection, and then DELETE the entire minishift_home AFTER the cdk starts but BEFORE an openshift connection is made. 

If i wait until after the connection is created, the oc location is stored in the connection object, even if the oc file / folder are deleted after the fact. 

So really the only way I can make this fail is to set a breakpoint and rm -rf an entire folder at a key moment. Otherwise the patch seems clean to me.

adietish:
@jeffmaury @odockal can you please test this again?

adietish:
@jeffmaury thx!

PR #1842, opened 2019-04-23
JBIDE-25814 Select recent connection in server adapter wizard instead of

jkopriva:
first

Signed-off-by: Josef Kopriva <jkopriva@redhat.com>

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ChooseOpenshiftConnectionComposite.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ChooseOpenshiftConnectionComposite.java
@@ -48,6 +48,7 @@ import org.jboss.tools.openshift.internal.core.ocbinary.OCBinary;
 import org.jboss.tools.openshift.internal.ui.OpenShiftUIActivator;
 
 import com.openshift.restclient.OpenShiftException;
+import com.openshift.restclient.authorization.IAuthorizationContext;
 
 public class ChooseOpenshiftConnectionComposite extends Composite {
 	private IConnectionChangedListener connectionChangeListener;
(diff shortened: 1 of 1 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/server/ChooseOpenshiftConnectionComposite.java
```diff
@@ -177,8 +177,14 @@ public void widgetSelected(SelectionEvent e) {
...
-			connectionCombo.select(0);
-			setSelectedConnection(connections.get(0));
+			IConnection recentConnection = ConnectionsRegistrySingleton.getInstance().getRecentConnection();
+			for(String connectionItem : connectionCombo.getItems()) {
+				if(connectionItem.contains(recentConnection.getHost())) {
+					connectionCombo.select(connectionCombo.indexOf(connectionItem));
+					break;
+				}
+			}
+			setSelectedConnection(recentConnection);
```
adietish: I would suggest to extract l180-187 to a separate method and improve readability and maybe reuse it. 
Another suggestion then is to use a ```ComboViewer``` instead of a ```Combo``` to eliminate all the item/connection matching operations that one can find in this class.
Such duplicate and unnecessary code is ex. at
* l101
* l116
* l181
* l219

adietish:
I then have another thing to beg for:
There are a few sonar issues in this class where it could be great if we could fix those (most easy option is to install sonar into your Eclipse and eliminate the need to lookup in sonarcloud):

* ```List<String> connectionNames = new ArrayList<String>();```
->
```List<String> connectionNames = new ArrayList<>();```
* ```allCons.forEach((con) -> connectionNames.add(createLabel(con)));```
->
```allCons.forEach(con -> connectionNames.add(createLabel(con)));```

adietish:
Then there's this to change (l234):
```if ("Basic".equals(authScheme)) {```
->
```if (IAuthorizationContext.AUTHSCHEME_BASIC.equals(authScheme)) {```

adietish:
@jkopriva thx for the corrections! +1

jkopriva:
@adietish Thx for review!

PR #1847, opened 2019-06-14
[JBDS-4777] Extract docker image uri from container spec

adietish:
Allow docker image uri (where the docker image labels for
deployment-dir, debug-mode & debug port are present) to be extracted
from dc>spec>template>spec>containers if there's no
dc>spec>trigger>ImageChange

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
@jeffmaury please review

adietish:
> Just a comment: I remember that the server adapter was supposed to work if there is no DC but only an RC but seems the code is bound to DC. I don't think this is related to your patch but we should probably open an issue at least to verify
Agreed, its true that it wont work and it's true that it's not related to my patch. We have an existing issue for this problem though: https://issues.jboss.org/browse/JBIDE-24981

PR #1848, opened 2019-06-24
[JBIDE-26408] warn that debugging nodejs only supports node.js <=7

adietish:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/IDialogProvider.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/IDialogProvider.java
@@ -1,0 +1,16 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.openshift.core;
+
+public interface IDialogProvider {
+
+	void warn(String title, String message, String preferencesKey);
+}
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/IDialogProvider.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/IDialogProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015 Red Hat, Inc.
+ * Copyright (c) 2019 Red Hat, Inc.
  * Distributed under license by Red Hat, Inc. All rights reserved.
  * This program is made available under the terms of the
  * Eclipse Public License v1.0 which accompanies this distribution,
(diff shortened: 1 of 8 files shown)


adietish:
@jeffmaury please review. 
This is a simple warning dialog when server adapter for Node.js starts in debug mode: we only support Node.js <= v7. A checkbox allows the user not to get bothered any further. Me feels that we need this given that without it, users have no idea why this doesnt work, they have the debugger complaining not to be able to connect to the v8 on OpenShift. Given that we're not able to detect the Node.js that's running, warning upfront seems the only possible solution.
On the technical level, we need to open a dialog in a core plugin where there are no ui packages available. I thus had to create an extension point in org.jboss.tools.openshift.core (that org.jboss.tools.openshift.ui would contribute to) that allows access to simple ui messaging (only #warn for now).

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/IDialogProvider.java
```diff
@@ -0,0 +1,16 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Red Hat, Inc.
```
jeffmaury: Should be 2019

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
```diff
@@ -231,6 +231,7 @@ private static void updateServer(IServerWorkingCopy server) {
 		server.setAttribute(IDeployableServer.DEPLOY_DIRECTORY_TYPE, IDeployableServer.DEPLOY_CUSTOM);
 		server.setAttribute(IDeployableServer.ZIP_DEPLOYMENTS_PREF, true);
 		server.setAttribute(Server.PROP_AUTO_PUBLISH_TIME, DEFAULT_AUTO_PUBLISH_DELAY);
+
```
jeffmaury: Is this change really required ?

plugins/org.jboss.tools.openshift.js/src/org/jboss/tools/openshift/js/server/behaviour/OpenShiftNodejsLaunchController.java
```diff
@@ -16,19 +16,27 @@
...
 	public OpenShiftNodejsLaunchController() {
 		super();
 	}
 
 	@Override
 	protected void startDebugging(OpenShiftServerBehaviour beh, DebugContext context, IProgressMonitor monitor) {
+		// https://issues.jboss.org/browse/JBIDE-26408
+		Dialogs.INSTANCE.warn("Node.js 8+ unsupported	", 
+				"Debugging is only supported up to Node.js 7. If your Node.js is newer, debugging will not work.\n"
+				+ "If using an OpenShift template to create the application, set 'NODEJS_PROPERTY' to '6'",
```
jeffmaury: Why do we recommand Node 6 when the text claims we support up to Node 7 ?

adietish:
@jeffmaury thx for correcting things right away.

plugins/org.jboss.tools.openshift.js/src/org/jboss/tools/openshift/js/server/behaviour/OpenShiftNodejsLaunchController.java
adietish: I amended this change to your commit.

PR #1853, opened 2019-07-08
JBIDE-26721 Remove smartimport dependency from OpenShift ITests

jkopriva:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


itests/org.jboss.tools.openshift.ui.bot.test/META-INF/MANIFEST.MF
```diff
@@ -2,14 +2,13 @@ Manifest-Version: 1.0
...
 Bundle-SymbolicName: org.jboss.tools.openshift.ui.bot.test;singleton:=true
-Bundle-Version: 3.7.0.qualifier
+Bundle-Version: 3.8.0.qualifier
 Bundle-Activator: org.jboss.tools.openshift.ui.bot.test.Activator
 Require-Bundle: org.eclipse.ui.ide,
  org.eclipse.ui.editors,
  org.eclipse.ui.console,
  org.eclipse.core.databinding,
  org.eclipse.debug.core,
- org.eclipse.egit.ui.smartimport;bundle-version="4.8.0";resolution:=optional,
```
adietish: Ok this fixes it, thx. I still dont understand why it fails if the bundle is marked as optional.
The other question I have is if this failed since October (I see this being added in October)? Or was smartimport removed from TP or?

adietish: according to @jkopriva egit.ui.smartimport bundle doesnt exist any more, it was merged into another one. Explains it.

PR #1856, opened 2019-07-16
[JBIDE-26744] trigger full deploy upon saving server in deployment (editor) page

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


robstryker:
I'm going to +1 this, but just want to list some concerns.
1) If your server has multiple deployed modules, you're setting them all as requiring a full publish, even if the user only customized one of them
2) I still don't know how, on the publish side, you're cleaning up the old deployment names. If you're only allowing these changes to a server that has no deployed modules, (and thus no cleanup, like wfly and eap) then this is fine. But please make sure you test that old deployment names in your deploy folder are cleaned up.

adietish:
@jkopriva can you please test this fix for the issue you experienced in https://issues.jboss.org/browse/JBIDE-26694?

jkopriva:
LGTM, we also delete old deployment - could lead in for example name collisions for datasources.

PR #1858, opened 2019-07-19
[JBIDE-26746] remove old deployment when outputName changes

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
@@ -11,6 +11,7 @@
 package org.jboss.tools.openshift.core.server.behavior;
 
 import java.io.File;
+import java.util.Arrays;
 import java.util.stream.Stream;
 
 import org.eclipse.core.resources.IProject;
@@ -34,6 +35,8 @@ import org.jboss.tools.as.core.server.controllable.subsystems.StandardFileSystem
 import org.jboss.tools.common.util.FileUtils;
 import org.jboss.tools.openshift.common.core.utils.StringUtils;
 import org.jboss.tools.openshift.core.server.OpenShiftServerUtils;
+import org.jboss.tools.openshift.core.server.OutputNamesCacheFactory;
+import org.jboss.tools.openshift.core.server.OutputNamesCacheFactory.OutputNamesCache;
 import org.jboss.tools.openshift.core.server.RSync;
 import org.jboss.tools.openshift.core.server.behavior.eap.OpenshiftEapProfileDetector;
 import org.jboss.tools.openshift.internal.core.OpenShiftCoreActivator;
@@ -52,6 +55,10 @@ public class OpenShiftPublishController extends StandardFileSystemPublishControl
 		if (!modulesIncludesMagicProject(server, deployProject)) {
 			publishRootModule(monitor, deployProject, localDirectory);
 		}
+		
+		final File localFolder = getDeploymentsRootFolder();
+		syncPodsToDirectory(localFolder, monitor);
+		deleteOldDeployments(getServer());
 	}
 
 	private void publishRootModule(final IProgressMonitor monitor, final IProject deployProject, final File localDirectory)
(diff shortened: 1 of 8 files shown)


plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
```diff
@@ -108,6 +111,8 @@ public void publishFinish(IProgressMonitor monitor) throws CoreException {
 		super.publishFinish(monitor);
 
 		final File localFolder = getDeploymentsRootFolder();
+		syncPodsToDirectory(localFolder, monitor);
```
robstryker: Why is syncPodsToDirectory being done during publishFinish?  Shouldn't that be done at the start of the publish operation, like publishStart or something? My concern is if this is done after publish, it may be buggy.

For example:
   User changes Sample.war.  publishStart() is called, publishModule(etc) is called.  You copy in the new changes of Sample.war,  then you syncPodsToDirectory and end up overwriting your changes? Not sure if this is correct but its my current concern.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
```diff
@@ -186,4 +199,33 @@ protected boolean isEapProfile() {
...
+		Arrays.stream(server.getModules())
+			.forEach(module -> deleteOldDeployment(module, deploymentsRootFolder));
+	}
+
+	private void deleteOldDeployment(final IModule module, final File deploymentsRootFolder) {
+		OpenShiftServer delegate = OpenShiftServerUtils.getServerDelegate(getServer());
+		ModuleOutputNames moduleOutputNames = delegate.getModuleOutputNames();
+		if (moduleOutputNames.isModified(module)) {
+			deleteOldDeployment(module, moduleOutputNames, deploymentsRootFolder);
+			delegate.getModuleOutputNames().reset();
```
robstryker: This seems like a bug. If you're iterating through deleteOldDeployments(etc) and each loop then calling deleteOldDeployment, what happens if 5 deployments have changed (Wonka1.war through Wonka5.war)?   On the first iteration, you see Wonka1.war was modified. You reset the output names. On the second iteration, Wonka2 is marked as unchanged, because you've already reset the output names. It would have been marked as changed if you didn't just reset the cache on the previous iteration.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
```diff
@@ -186,4 +199,33 @@ protected boolean isEapProfile() {
 		String profile = ServerProfileModel.getProfile(getServer());
 		return OpenshiftEapProfileDetector.PROFILE.equals(profile);
 	}
+
+	public void deleteOldDeployments(final IServer server) throws CoreException {
+		File deploymentsRootFolder = getDeploymentsRootFolder();
+		Arrays.stream(server.getModules())
+			.forEach(module -> deleteOldDeployment(module, deploymentsRootFolder));
```
robstryker: Seems the output name cache should be reset here, not below.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
adietish: Fair point but I dont think that this matters given that the local files are newer and thus wont be overwritten. Still, I'll move sync down to publishing start given that it's more logical.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
adietish: true, should reset each module on it's own, not gloabbly.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/OpenShiftPublishController.java
adietish: yes that's where individual resets should happen.

adietish:
@robstryker review again, please?

robstryker:
+1 I guess :D   If it works ;)

adietish:
@jkopriva can you please test this?

jkopriva:
@adietish Hi, I have tried it, old deployment is correctly deleted after deployment name change.

PR #1859, opened 2019-08-08
[JBIDE-26754] wildfly: first undeploy and then only redeploying with new name

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
@@ -753,4 +753,10 @@ public class OpenShiftServerUtils {
 		return collection;
 	}
 
+	public static void setModulePublishState(int state, IModule[] modules, Server server) {
+		Arrays.stream(modules).forEach(module -> 
+			server.setModulePublishState(new IModule[] { module }, state));
+	}
+
+
 }
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapPublishController.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapPublishController.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 Red Hat, Inc.
+ * Copyright (c) 2017-2019 Red Hat, Inc.
  * Distributed under license by Red Hat, Inc. All rights reserved.
  * This program is made available under the terms of the
  * Eclipse Public License v1.0 which accompanies this distribution,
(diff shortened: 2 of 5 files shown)


adietish:
I now remove the ".deployed" marker file before syncing up to the pod. Nevertheless, the new deployment still fails with the same error.

adietish:
I now remove the .dodeploy marker and loop, syncing down & waiting for the .undeployed marker (which indicates that undeployment happened). Once this was detected the new deployment is copied over and sync'ed up. This now works for me. The only bug I experience is where:
1) ASSERT: have server adapter, pod, local project.
2) ASSERT: have the deployment name changed
3) EXEC: kill the pod
4) ASSERT: new pod is created with the war being deployed using the original name
5) EXEC: change the deployment name again
Result:
Removal of the "old" deployment fails (since it's not the name in step 2) but the original name as in step 4). Deployment is executed using the name specified in 5) but fails in wildfly/eap with the error specified in JIRA.

adietish:
@jkopriva can you please give it a try? It is also important that the original use case where a deployment to tomcat (I faced special problems with wildfly/eap) works.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/OpenShiftServerUtils.java
```diff
@@ -753,4 +753,10 @@ public static CoreException toCoreException(String msg) {
 		return collection;
 	}
 
+	public static void setModulePublishState(int state, IModule[] modules, Server server) {
```
robstryker: WTP has two different concepts of what an IModule[] array means. 
One is an array of module items, just a list basically, which may or may not have anything to do with each other. 
The second is a specific module and its parent modules in an array representing something like a tree-path to a single module. 

Your method signature or javadoc here should make clear whether you are setting the publish state of every module in the list (setModulesPublishStates) or a single module.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapPublishController.java
```diff
@@ -10,29 +10,78 @@
...
-		return super.publishModule(IServer.PUBLISH_CLEAN, deltaKind, module, monitor);
+	private void deleteFiles(File... files) {
+		Arrays.stream(files).forEach(file -> {
+			if (file.exists()) {
+				FileUtils.remove(file);
+			}
+		});
+	}
+
+	private void waitForUndeployed(File moduleDeployment, IProgressMonitor monitor) throws CoreException, InterruptedException {
```
robstryker: I worry that if you try to delete something which was incompletely deployed or copied over but not actually deployed by the server, you may end up waiting the full timeout because you're waiting for the server to recognize the undeployment of something it never fully deployed. Not sure how you'd accurately detect this in advance to reduce the chance of long timeouts.

adietish:
@jkopriva I tested the usecase of https://issues.jboss.org/browse/JBIDE-26694 and the new war is deployed nicely using the new war-name. Would be great if you could also confirm this.

jkopriva:
@adietish I have tried it with eap and it works like a charm! With tomcat I was not able to get app working on new url, but it is correctly rsynced. Please can you look also on sonar issues?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/server/behavior/eap/OpenShiftEapPublishController.java
adietish: @robstryker maybe detecting a .failed marker and not wait in this case would be a valuable . Other than this I dont see how we'd detect such a failed deployment.

adietish: @robstryker looking at the code I see the ```waitForUndeployed``` looking for a .failed marker and stopping the wait if it is detected. 
```
	private void waitForUndeployed(File moduleDeployment, IProgressMonitor monitor) throws CoreException, InterruptedException {
		long start = System.currentTimeMillis();
		do {
				Thread.sleep(WAIT_FOR_UNDEPLOYED_TIMEOUT/UNDEPLOYED_CHECKS);
				syncPodsToDirectory(monitor);
		}
		while(!isUndeployed(moduleDeployment)
				&& System.currentTimeMillis() < start + WAIT_FOR_UNDEPLOYED_TIMEOUT
				&& !monitor.isCanceled());
	}

	private boolean isUndeployed(File moduleDeployment) {
		return fileExists(new File(moduleDeployment + DeploymentMarkerUtils.UNDEPLOYED),
				new File(moduleDeployment + DeploymentMarkerUtils.FAILED_DEPLOY));
	}
```
In that sense I believe that the most we can do is implemented. Right?

adietish:
@jkopriva sonarcloud does not detect any code smells or other complaints in the code. It complains about the lack of coverage. I'd think that the deployment test in ITs is good enough, agree?

PR #1867, opened 2019-09-06
[JBIDE-26813] use openshift-restclient-java 8.0.1

adietish:
this depends on https://github.com/openshift/openshift-restclient-java/issues/395 being merged and the client library 8.0.1 being published to maven repo.

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
+++ b/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
@@ -269,7 +269,7 @@ public class ConnectionWrapper extends AbstractOpenshiftUIElement<IOpenShiftConn
 		updateWithResources(loadProjects());
 		state.set(LoadingState.LOADED);
 		fireChanged();
-		for (ProjectWrapper project : projects.values()) {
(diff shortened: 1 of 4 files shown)


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/ConnectionWrapper.java
```diff
@@ -269,7 +269,7 @@ public void refresh() {
 		updateWithResources(loadProjects());
 		state.set(LoadingState.LOADED);
 		fireChanged();
-		for (ProjectWrapper project : projects.values()) {
+		for (ProjectWrapper project : new ArrayList<>(projects.values())) {
```
jeffmaury: What's the rationale for this ?

adietish: This is where a ConcurrentModificationException was triggered when you refresh a server twice in a row.
I included this small fix even though it's not related to the aim of this PR. I could separate it in terms of issue and PR if we agree on this being preferrable.

adietish:
testPR

PR #1880, opened 2019-09-30
JBIDE-26880 - fix default naming for crc servers

robstryker:

We typically don't add the (Tech Preview) to the server name itself, but rather to the server type. The server names should not have rogue (tech Preview) hanging out there forever because then when the user updates their workspace, those old servers will still be named tech preview, which is not desirable. 

This PR fixes the default naming for the server type to remove the Tech Preview and make sure both methods of creating runtimes use the same default name logic. 

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


adietish:
testPR

PR #1882, opened 2019-10-03
[JBIDE-26876] moved console url retrieval to client + support for OS4

adietish:
depends on the changes in the openshift-restclient-java: https://github.com/openshift/openshift-restclient-java/pull/401
This wont compile before the change for the client library is merged and the maven artifact published.

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/core/listeners/ConfigureCRCFrameworksListener.java
```diff
@@ -35,7 +35,7 @@
 
 public class ConfigureCRCFrameworksListener extends UnitedServerListener {
 
-	private static final String CRC_HOST = "https://api.crc.testing";
+	private static final String CRC_HOST_URL = "https://console-openshift-console.apps-crc.testing";
```
jeffmaury: I don't understand this change ?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
```diff
@@ -360,6 +361,16 @@ public boolean isDefaultHost() {
 		return false;
 	}
 
+	@Override
+	public String getConsoleURL(IProgressMonitor monitor) {
+		return client.adapt(IConsole.class).getConsoleUrl();
+	}
+
+	@Override
+	public <R extends IResource> String getConsoleURL(R resource, IProgressMonitor monitor) {
```
jeffmaury: What's the added value of using generic type instead of IResource ?

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/core/connection/ConnectionTest.java
```diff
@@ -564,6 +564,11 @@ public void should_clear_secure_data_when_connection_deleted() throws MalformedU
 		assertFalse(SecurePreferencesFactory.getDefault().nodeExists(key.getKey()));
 	}
 
+	@Test
+	public void shouldReturnConsoleURL() {
+
```
jeffmaury: Why is this test empty ?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/core/listeners/ConfigureCRCFrameworksListener.java
adietish: this url is what you get when you retrieve it via the crc cmd line:
```
$ crc console
```

tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/core/connection/ConnectionTest.java
adietish: left over, removed it.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
adietish: with generics you get the exact type, what's the drawback?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/core/listeners/ConfigureCRCFrameworksListener.java
adietish: reverted it, given that the REST endpoint is not identical to the console in OS4

adietish:
testPR

adietish:
openshift-restclient-java 8.1.0-SNAPSHOT was deployed to repository.jboss.org. This PR should now compile.
@jeffmaury please test again.

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/core/connection/Connection.java
jeffmaury: I don't see the added value as having the parameter typed as IResource. Generics are usefull in the case you want to return the same type for example but here you are returning a String so no need

jeffmaury:
Tested on OCP 3.11, works for me now.

odockal:
I tried CRC beta 5 and CDK 3.10 (ocp 3.11) and it worked. But I did not make it works with OCP 4.2 so far.

adietish:
testPR

adietish:
@jeffmaury dare to +1 this please?

PR #1883, opened 2019-10-21
[JBIDE-26697] dont watch unnecessary events for Event, PVC and Template

adietish:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
+++ b/plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015-2017 Red Hat, Inc.
+ * Copyright (c) 2015-2019 Red Hat, Inc.
  * Distributed under license by Red Hat, Inc. All rights reserved.
  * This program is made available under the terms of the
  * Eclipse Public License v1.0 which accompanies this distribution,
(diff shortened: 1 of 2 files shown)


adietish:
@odockal can QE please test this change? We're stopping to watch Builds, Events, PVCs, Templates. In terms of code (and quick manual testing) this should be fine. Builds and Events are not editable (Editors watch for changes since you cannot save a resource that you edit if it was changed on the server). Can you please verify throughout?

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
```diff
@@ -39,13 +39,23 @@
 
 public class WatchManager {
 
+	public static final String[] KINDS = new String[] { 
+			//ResourceKind.BUILD,
```
jeffmaury: Why BUILD are not watched ?

adietish: Builds are not editable (no editor that needs to watch them), Properties are not watching any resources, there's no occurrence where I see builds being watched. So why watch them?

adietish:
testPR

plugins/org.jboss.tools.openshift.core/src/org/jboss/tools/openshift/internal/core/WatchManager.java
adietish: @jeffmaury looking into it in more details I see you being right: added builds in again.

PR #1898, opened 2020-01-27
JBIDE-26993: Provide download infrastructure for cli

sbouchet:
added test

Signed-off-by: Stephane Bouchet <sbouchet@redhat.com>


tests/org.jboss.tools.openshift.test/src/org/jboss/tools/openshift/test/common/ui/utils/OdoDownloadHelperTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Test;
+
+public class OdoDownloadHelperTest {
+
+	public static final String ODO_CACHE_DIR = System.getProperty("user.home") + File.separatorChar + ".odo-tests"; //$NON-NLS-1$ //$NON-NLS-2$
+
+	@BeforeClass
+	public static void setup() throws IOException {
+		FileUtils.deleteDirectory(new File(ODO_CACHE_DIR));
+		UsageReportPreferences.setEnabled(false);
```
jeffmaury: Not needed, there is an env variable wich is set by our jobs.

sbouchet: oh thanks i did not find it for local tests.

PR #1906, opened 2020-02-12
[JBIDE-27047] Read <20k of pull secret, validate in background

adietish:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/CRCLocationSection.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/CRCLocationSection.java
@@ -1,5 +1,5 @@
 /******************************************************************************* 
- * Copyright (c) 2019 Red Hat, Inc. 
+ * Copyright (c) 2019-2020 Red Hat, Inc. 
  * Distributed under license by Red Hat, Inc. All rights reserved. 
  * This program is made available under the terms of the 
  * Eclipse Public License v1.0 which accompanies this distribution, 
@@ -32,15 +32,15 @@ import org.jboss.tools.openshift.internal.crc.server.core.adapter.CRC100Server;
 
 public class CRCLocationSection extends AbstractLocationSection {
 
-	private static String SECTION_TITLE = "CRC Details";
-	private static String LABEL_STRING = "crc binary file: ";
-	private static String COMMAND_NAME = "Modify crc binary Location";
-	private static String LOC_ATTR = CRC100Server.PROPERTY_BINARY_FILE;
+	private static final String SECTION_TITLE = "CRC Details";
+	private static final String LABEL_STRING = "CRC Binary: ";
+	private static final String COMMAND_NAME = "Modify crc binary Location";
+	private static final String LOC_ATTR = CRC100Server.PROPERTY_BINARY_FILE;
 	private Text pullSecretText;
 	private Button pullSecretBrowse;
 	private ControlDecoration pullSecretDecor;
-	private SelectionAdapter pullSecretSelListener;
 	private ModifyListener pullSecretModListener;
+	private final PullSecretValidation pullSecretValidation = new PullSecretValidation();
 
 	public CRCLocationSection() {
 		super(SECTION_TITLE, LABEL_STRING, COMMAND_NAME, LOC_ATTR);
@@ -84,30 +84,23 @@ public class CRCLocationSection extends AbstractLocationSection {
 	@Override
 	protected void addListeners() {
 		super.addListeners();
-		pullSecretSelListener = new SelectionAdapter() {
+		pullSecretBrowse.addSelectionListener(new SelectionAdapter() {
 			@Override
 			public void widgetSelected(SelectionEvent e) {
 				browseClicked(pullSecretText, FILE);
 				validate();
 			}
-		};
-		pullSecretBrowse.addSelectionListener(pullSecretSelListener);
-
-		pullSecretModListener = new ModifyListener() {
-			@Override
-			public void modifyText(ModifyEvent e) {
-				execute(new SetPullSecretPropertyCommand(server));
-			}
-		};
-		pullSecretText.addModifyListener(pullSecretModListener);
-		pullSecretText.addModifyListener(new ModifyListener() {
-			public void modifyText(ModifyEvent e) {
-				validate();
-			}
 		});
+
+		pullSecretModListener = (ModifyEvent e) -> 
+			execute(new SetPullSecretPropertyCommand(server));
+		pullSecretText.addModifyListener(pullSecretModListener);
+		pullSecretText.addModifyListener((ModifyEvent e) -> 
+			validate());
 
 	}
 
+	@Override
 	protected void validate() {
 		validateBinary();
 		validatePullSecret();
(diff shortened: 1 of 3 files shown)


adietish:
@sbouchet @odockal could you please review/test this?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/PullSecretValidation.java
```diff
@@ -0,0 +1,126 @@
...
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class PullSecretValidation {
+
+	private static final int MAX_PULLSECRET_FILE_SIZE = 20 * 1024; // 20k
+
+	private final AtomicReference<String> error = new AtomicReference<>();
+	private final AtomicReference<String> file = new AtomicReference<>();
```
sbouchet: for better maintenance, maybe rename it as it collide in several places ( constructor, ... )

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/PullSecretValidation.java
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		private String validateJsob(String file) {
+			String msg = null;
+			try {
+				if (!isValidJson(readPullSecretFile(file))) {
+					msg = NLS.bind("Pull Secret file {0} is invalid. Make sure it is valid JSON.", file);
+				}
+			} catch (IOException | UncheckedIOException e) {
+				msg = NLS.bind("Pull Secret file {0} is invalid. Make sure it is valid JSON: {1}", file, e.getLocalizedMessage());
```
sbouchet: Maybe add something like  : "validation error : " before setting the error message.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/PullSecretValidation.java
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		private boolean isValidJson(String json) {
+			try {
+				new JSONObject(json);
+			} catch (JSONException ex) {
+				try {
+					new JSONArray(json);
+				} catch (JSONException ex1) {
+					return false;
```
sbouchet: i know it's coming from the old file, but looks weird to me to actually perform a double try catch ...  maybe refactor to extract the JSONArray validation in another method can help :)

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/PullSecretValidation.java
```diff
@@ -0,0 +1,126 @@
...
+				}
+			} catch (IOException | UncheckedIOException e) {
+				msg = NLS.bind("Pull Secret file {0} is invalid. Make sure it is valid JSON: {1}", file, e.getLocalizedMessage());
+			}
+			return msg;
+		}
+
+		private String readPullSecretFile(String pullSecretFile) throws IOException {
+			String content = null;
+			try (Stream<String> lines = Files.lines(Paths.get(pullSecretFile))) {
```
odockal: I am getting proper validation I have to say. When I try to pass big binary file I get right message (Pull secret file /b/b/b is invalid. Make sure it is valud JSON: ) and then I get error message that was caught in try catch. java.nio.charset.MalformedInputException: Input length = 1.

This would mean that the file read failed due to charset exception (aa binary is not standard UTF-8 coded file, which is ok) but I think that this is not proper validation. Like maybe it is intentional, not sure here. I quite do not understand the limit then since the real validtion is based on if it is text file or binary (thus throwing charset exception).

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/PullSecretValidation.java
adietish: @sbouchet I dont think that this is a good idea. The message is displayed with an error icon which basically already says that this is an error. The very same is true for the error decorator that is displayed next to the input field with the path to the file. In other words, I'd argue that it's already obvious visually that this is an error. Why repeat yourself in the text? Good UX design would also emphasize to have rather short error messages.
Furthermore we have a problem with the available space: we have very limited space in the title are where this error is displayed. Shorter is imho thus better.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/CRCLocationSection.java
```diff
@@ -32,15 +32,15 @@
 
 public class CRCLocationSection extends AbstractLocationSection {
 
-	private static String SECTION_TITLE = "CRC Details";
-	private static String LABEL_STRING = "crc binary file: ";
-	private static String COMMAND_NAME = "Modify crc binary Location";
-	private static String LOC_ATTR = CRC100Server.PROPERTY_BINARY_FILE;
+	private static final String SECTION_TITLE = "CRC Details";
+	private static final String LABEL_STRING = "crc binary file: ";
```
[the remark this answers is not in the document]
adietish: +1, but to be consistent across the board it would be even better to use all capital letters (or turn all occurrences to capital letter in the first place, all others in small letters)

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/PullSecretValidation.java
adietish: the problem here was that the code tried to load the whole file and then caused an out-of-error. I am not validating this, I'm just preventing this by assuming that a pull secret file would not be larger than 20k. We could test the file size beforehand and refuse to load anything that's larger than 20k and would be then explicit about this limitation.
The validation error that you see is caused by reading a binary file where a text file is expected. I could catch this error and use a more userfriendly error "not a text file" or something.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/PullSecretValidation.java
adietish: Can you please explain where you see a collision?

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/PullSecretValidation.java
odockal: I tried latest change -> It filters files < 20 K bytes, which is fine now. I would still add a catch for malformedInputException which, as you said, is exactly reading a binary (not text file), and show more user friendly message. Thanks.

adietish: @odockal : I'm now catching this error and displaying a message that says that the file isn't a text file. Please review

PR #1910, opened 2020-02-19
Pull secret validation did not enable Finish button - JBIDE-27064

odockal:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/CRC100ServerWizardFragment.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/CRC100ServerWizardFragment.java
@@ -154,13 +154,10 @@ public class CRC100ServerWizardFragment extends CDKServerWizardFragment {
 		String errorMessage = pullSecretValidation.validate(pullSecretFile, 
 				(String error) -> {
 					togglePullSecretDecorator(error);
-					setComplete(error != null);
 					if (error != null) {
 						handle.setMessage(error, IMessageProvider.ERROR);
 					}
 				});
-		// dont complete until job validated pull secret
-		setComplete(false);
 		togglePullSecretDecorator(errorMessage);
 		return errorMessage;
 	}


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/ui/CRC100ServerWizardFragment.java
```diff
@@ -154,13 +154,10 @@ private String validatePullSecret() {
 		String errorMessage = pullSecretValidation.validate(pullSecretFile, 
 				(String error) -> {
 					togglePullSecretDecorator(error);
-					setComplete(error != null);
 					if (error != null) {
 						handle.setMessage(error, IMessageProvider.ERROR);
 					}
 				});
-		// dont complete until job validated pull secret
-		setComplete(false);
```
jeffmaury: TBH I don't understand the use case as it is working for me. So you removed the setComplete flag management but it looks ok for me. Can you describe the scenario ?

odockal: You can create CRC adapter without this change?
I can unfortunately reply with the same: TBH I have no idea why it does not work without this change...

odockal: Problem is that one cannot get finish button enabled even if he uses proper crc binary and pull secret file.

jeffmaury: Working good with 14.12.0.AM1.B5884

odockal: That's before the error occured. Can you please try newer one? I was reproducing this on B5949.

odockal: It came to me (I did not thoroughly check all the logical paths) that before there was no use of this setComplete method and logic whether enabled the button is here: https://github.com/jbosstools/jbosstools-openshift/blob/master/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/cdk/server/ui/CDKServerWizardFragment.java#L63

odockal: I am taking back my previous statement. You seems to be right that how it is implemented should be ok and finish should get enabled after validation. BUT, it does not work. I will try to dig a bit more.

odockal: It seems that there is something wrong in the sequence of calling some of CRC100ServerWizardFragment.PullSecretModifyListener#modifyText() -> CDKServerWizardFragment#validate() and CRC100ServerWizardFragment#findError(boolean) and CDKServerWizardFragment#isComplete().

Thing is, if CRC100ServerWizardFragment#findError(boolean) sets every time property complete to false (here), it is ok when findError is called from [validate], since the validate then sets complete property to correct value but it is not alright then findError is called from [isComplete] from the parent because the statement combines also super.isComplete which findError before sets to false, it never evaluates to true and finish button is never enabled.

[here]: https://github.com/jbosstools/jbosstools-openshift/pull/1910/files/6a2dbd8e2cd2b89f162c565f6b25cfe51f9fcae6#diff-4dfb8912b2e6122d5b826a3e1bf8dd09L163
[validate]: https://github.com/jbosstools/jbosstools-openshift/pull/1910/files/6a2dbd8e2cd2b89f162c565f6b25cfe51f9fcae6#diff-4dfb8912b2e6122d5b826a3e1bf8dd09L140
[isComplete]: https://github.com/jbosstools/jbosstools-openshift/blob/6a2dbd8e2cd2b89f162c565f6b25cfe51f9fcae6/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/cdk/server/ui/CDKServerWizardFragment.java#L65

odockal: @jeffmaury ^^

odockal: In short I think it is pointless to toggle completeness in findError at all because if findError method would return anything else than null it would not evaluate isComplete to true anyway.

jeffmaury:
I managed to reproduce from master. The problem with your PR is that Finish button is enabled as soon as you select the CRC binary. I figured out that the problem was that:
1) complete was set to true if there was an error from the validating job (should be the opposite)
2) validating job was not spawned the first first (ie pullSecretFile is null)
3) upon spawning validating job setComplete was always set to false where it should be based on the returned error message.

See my PR #1912

odockal:
Closed in favor of #1912.

PR #1927, opened 2020-06-05
JBIDE-27287: Add reporting to odo experience

jeffmaury:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/itests/org.jboss.tools.cdk.ui.bot.test/META-INF/MANIFEST.MF
+++ b/itests/org.jboss.tools.cdk.ui.bot.test/META-INF/MANIFEST.MF
@@ -5,7 +5,7 @@ Bundle-SymbolicName: org.jboss.tools.cdk.ui.bot.test
 Bundle-Version: 3.11.0.qualifier
 Bundle-Vendor: Red Hat
 Require-Bundle: org.junit,
- org.eclipse.reddeer.go;bundle-version="[2.1.0,3.0.0)",
+ org.eclipse.reddeer.go,
  org.eclipse.wst.server.core;bundle-version="1.5.0",
  org.eclipse.wst.server.ui;bundle-version="1.5.200",
  org.eclipse.jdt.launching;bundle-version="3.7.0",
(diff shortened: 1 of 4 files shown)


test-framework/org.jboss.tools.cdk.reddeer/META-INF/MANIFEST.MF
```diff
@@ -8,7 +8,7 @@ Bundle-ActivationPolicy: lazy
 Bundle-Vendor: Red Hat
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Require-Bundle: org.junit,
- org.eclipse.reddeer.go;bundle-version="[2.1.0,3.0.0)",
```
adietish, line 11: see comment above about the version range for reddeer

itests/org.jboss.tools.cdk.ui.bot.test/META-INF/MANIFEST.MF
```diff
@@ -5,7 +5,7 @@ Bundle-SymbolicName: org.jboss.tools.cdk.ui.bot.test
 Bundle-Version: 3.11.0.qualifier
 Bundle-Vendor: Red Hat
 Require-Bundle: org.junit,
- org.eclipse.reddeer.go;bundle-version="[2.1.0,3.0.0)",
+ org.eclipse.reddeer.go,
```
adietish, line 8: is the reason for eliminating the upper bound that we'll be using or are already using red deer 3.0? In general I'd advocate for an upper bound that excludes versions where we'd expect an API breakage. If we're using 3.0 already we maybe should now use 4.0 as an upper bound?

test-framework/org.jboss.tools.cdk.reddeer/META-INF/MANIFEST.MF
jeffmaury: The reason why I removed the upper bounds is that RedDeer switch to 3.0 when we changed the TP thus lots of builds failed but there was no API breakage AT ALL so I prefer discover them when they really EXIST

odockal: That's right. There is no API breakage but removed unused feature in IU. So it is new version but without a real breakage. I would maybe prefer to leave lower bound (2.1.0) in place, because between version 1 a 2 there is quite a lot breakages... But we control what reddeer version is used (via TP), so... it is at least nice to have...

PR #1965, opened 2020-12-15
JBIDE-27563: Add support for starters

jeffmaury:
- Also switched to odo 2.0.3

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/itests/org.jboss.tools.openshift.ui.bot.test/src/org/jboss/tools/openshift/ui/bot/test/AbstractODOTest.java
+++ b/itests/org.jboss.tools.openshift.ui.bot.test/src/org/jboss/tools/openshift/ui/bot/test/AbstractODOTest.java
@@ -17,6 +17,7 @@ import java.util.List;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.reddeer.common.wait.TimePeriod;
 import org.eclipse.reddeer.common.wait.WaitWhile;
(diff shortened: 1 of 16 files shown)


sbouchet (approved):
LGTM. I'm a little worried about the developer experience because the starter only works when the selected eclipse project is empty. there is no indication on the wizard about that.
Maybe adding a way to use an empty path in the file system and then import it as general project in eclipse should be an option ?

jeffmaury:
> 
> 
> LGTM. I'm a little worried about the developer experience because the starter only works when the selected eclipse project is empty. there is no indication on the wizard about that.
> Maybe adding a way to use an empty path in the file system and then import it as general project in eclipse should be an option ?

The starters are enabled only for empty projects, this is how it's implemented on the VSCode side. But there is a label which is displayed that explain that you can initialize your project from one of the available starters, did you get it ?

sbouchet:
> > LGTM. I'm a little worried about the developer experience because the starter only works when the selected eclipse project is empty. there is no indication on the wizard about that.
> > Maybe adding a way to use an empty path in the file system and then import it as general project in eclipse should be an option ?
> 
> The starters are enabled only for empty projects, this is how it's implemented on the VSCode side. But there is a label which is displayed that explain that you can initialize your project from one of the available starters, did you get it ?


retesting with latest changes and now i can see the label.

PR #1966, opened 2020-12-21
JBIDE-27588 Omit oc from crc initialization check

odockal:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/core/adapter/CRC100Server.java
+++ b/plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/core/adapter/CRC100Server.java
@@ -77,10 +77,6 @@ public class CRC100Server extends ServerDelegate {
 		if (!bin.exists() || !json.exists() || !log.exists()) {
 			return false;
 		}
-		File oc = new File(bin, "oc");
-		File ocExe = new File(bin, "oc.exe");
-		if( !oc.exists() && !ocExe.exists()) 
-			return false;
 		return true;
 	}
 


plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/core/adapter/CRC100Server.java
```diff
@@ -77,10 +77,6 @@ public boolean isInitialized() {
 		if (!bin.exists() || !json.exists() || !log.exists()) {
 			return false;
 		}
-		File oc = new File(bin, "oc");
```
jeffmaury, line 80: This check looks good to me but it is Windows only. It should be adapted rather than deleted

jeffmaury (changes requested):

odockal:
@jeffmaury I was working with corrupted crc, so this check removal is probably useless. With proper crc there are these files generated and check works.

odockal:
I will keep this open for a while, just in case.

plugins/org.jboss.tools.openshift.cdk.server/src/org/jboss/tools/openshift/internal/crc/server/core/adapter/CRC100Server.java
odockal: This is really quite a check I have to say, it works for both windows and linux but it is because in reality it checks existence of folder `oc` in directory `bin` which is there for linux and windows. `oc.exe` file exists under this `oc` directory. so I would say check for file `oc.exe` is useless in this case.

PR #1980, opened 2021-02-24
JBIDE-27611: Provides devfile code assist and validation

jeffmaury:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?

### DIFF
--- a/plugins/org.jboss.tools.openshift.core/plugin.xml
+++ b/plugins/org.jboss.tools.openshift.core/plugin.xml
@@ -365,5 +365,22 @@
           class="org.jboss.tools.openshift.internal.core.stack.NodeRemoteStackDebugger">
     </provider>
  </extension>
+ <extension
+       point="org.eclipse.core.contenttype.contentTypes">
+    <content-type
+          base-type="org.eclipse.wildwebdeveloper.yaml"
+          file-names="devfile.yaml,devfile.yml"
+          id="org.jboss.tools.openshift.core.devfile"
+          name="Devfile"
+          priority="normal">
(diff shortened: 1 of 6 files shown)


jeffmaury:
testPR

jeffmaury:
testPR

plugins/org.jboss.tools.openshift.core/plugin.xml
```diff
@@ -365,5 +365,22 @@
...
+          id="org.jboss.tools.openshift.core.devfile"
+          name="Devfile"
+          priority="normal">
+    </content-type>
+ </extension>
+ <extension
+       point="org.eclipse.wildwebdeveloper.json.schema">
+    <schema
+          pattern="devfile.y*ml"
+          url="platform:/plugin/org.jboss.tools.openshift.core/devfile.json">
```
sbouchet, line 382: shouldn't be 
```suggestion
          url="platform:/plugin/org.jboss.tools.openshift.core/resources/devfile.json">
```
i get error message on devfile.yaml saying that json is not reachable

sbouchet (changes requested):

plugins/org.jboss.tools.openshift.core/plugin.xml
jeffmaury: No location is correct as the file is at the root of the generated JAR. There may be something missing for Eclipse to find the file in dev mode

sbouchet (approved):
LGTM after review discussion

PR #1992, opened 2021-03-30
JBIDE-27739 - prompt for telemetry permission during crc setup

robstryker:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


jeffmaury (approved):
Works for me on Windows. However, I have a remark. We now have a succession of 3 different dialogs, the last one for telemetry which has a different style (white background).
Can't we at least reunify the first and last one (with the checkbox for telemetry). Also I don't think the second one is relevant as I don't think crc setup cleans the $HOME/.crc area

jeffmaury (approved):
LGTM

PR #1999, opened 2021-04-28
JBIDE-27829: Enhance navigation inside the Application Explorer

jeffmaury:

# Test
- Have a cluster with no projects, you should see a node with a link under the cluster node that allows to create a new project
- Have a project without deployments, you should see a node with a link under the project node that allows you to create a component
- Have an application without deployments (hard to reproduce go to OpenShift console and create a devfile based sample), you should see a node with a link under the application node that allows you to create a component


plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/models/applicationexplorer/CreateComponentMessageElement.java
```diff
@@ -0,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
```
sbouchet, line 2: just a typo but seems the template need a refresh :)

sbouchet (approved):
LGTM.

PR #2004, opened 2021-06-08
JBIDE-27676: Create component from component types and starters

jeffmaury:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


sbouchet:
LGTM but i think the component wizard dialog should be improved. 
it's the same dialog used when I create a component with the Create > Component on a project, and it's should not.
I I want to create a component using this action on the component tree, I have already picked the project and maybe the application . 
Using the create component from the devfileRegistry, I should be able to pick the project and the application I want to create the component too. right now, the project is the default one provided by the context, and i must type an application ( and not choose an already created one ) 
this lead to

plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/handler/applicationexplorer/CreateComponentHandler.java
```diff
@@ -44,31 +48,46 @@ public Object execute(final ExecutionEvent event) throws ExecutionException {
...
+			  }
+			} else {
+	      project = application.getParent();
 			}
-			project = application.getParent();
 		}
     final Shell parent = HandlerUtil.getActiveShell(event);
 		try {
-			openDialog(application, project, parent);
+			openDialog(componentType, application, project, parent);
```
sbouchet, line 70: I think there is a missing arg, because when I want to create a component from starter, the dialog does not show it on the project starter droplist, only the componentType is selected.

PR #2030, opened 2021-09-13
JBIDE-27985: Add JSON Schema based UI to operator based service creation

jeffmaury:

# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


sbouchet (commented):
LGTM but some remarks:
1/ i'm using dark mode and UI does not renders the textfields borders. ( see screenshot )
2/ there is no validation in the Ui and I'm getting no feedback when I'm using wrong input. using the example, I did not get errors because Name should be lowercase


Capture d’écran de 2021-09-17 15-21-12

jeffmaury:
> 
> 
> LGTM but some remarks:
> 1/ i'm using dark mode and UI does not renders the textfields borders. ( see screenshot )
> 2/ there is no validation in the Ui and I'm getting no feedback when I'm using wrong input. using the example, I did not get errors because Name should be lowercase
> 
> Capture d’écran de 2021-09-17 15-21-12

Hard to fix as I have a different view on Windows:
image

sbouchet (approved):
LGTM after the latest commit to handle the validation errors.
For the dark theme, it looks like there are still too many differences between how the OSs handle it.

PR #2097, opened 2022-12-06
JBIDE-28679 : migrate to odo V3.X

sbouchet:
# Pull Request Checklist
## General

## Code

## Testing
* Are there unit-tests?

## Function
* Does it work?


jeffmaury (changes requested):
- When a component is started in dev mode, forwarded ports are displayed in the explorer but the state of the component is not updated.
- when a component is started in dev mode and stop is required, seems the current odo session is cancelled but another is started immediately

sbouchet:
> * When a component is started in dev mode, forwarded ports are displayed in the explorer but the state of the component is not updated.

Fixed by refreshing tree in the call back and not right after start

> * when a component is started in dev mode and stop is required, seems the current odo session is cancelled but another is started immediately

Fixed by executing the stop callback when the process is terminated

> * on Windows as there are no way to properly stop the odo process, when a component was running in dev mode if will be reloaded in dev mode which it should not that dev state is only valid for a current Eclipse session

not sure to understand but it's likely the stop odo process on Windows that is missing. is it the "stop dev mode" that fails or the CTRL+C on terminal ? 

> * also I think debug is not working on Windows because the previous odo dev is not stopped correctly then I suppose when we run odo dev --debug we are waiting for a string that is not coming because the pod was not stopped and user is never asked to choose the port

Should be fixed with the above changes. if debug is started without any other process running, a dialog shows up on my case.

jeffmaury:
Tried:

- create comp
- start dev
- stop dev

But this got another odo process started:

jeffmaury (approved):
LGTM

